PR #40, opened 2017-10-19
Initialized Volume dataframe with np.int64 type to avoid integer overflow on Windows

cfillak:
I believe the Volume dataframes should be initialized explicitly as numpy.int64, because otherwise numpy will inherit the int type from the system's C compiler. This is fine for Unix and Mac systems, but even in 64 bit Windows, C Long is 32 bits. When pulling data for symbols with very large Volume, like ^GSPC (S&P500), negative values end up in the dataframes due to the 32-bit integer overflowing. Initializing as numpy.int64 solves this problem.


ranaroussi:
added it to version 0.1.0

PR #101, opened 2019-09-19
Sustainability info as dictionary

GregoryMorse:
Change sustainability information to a more easily usable dictionary


ranaroussi:
I've added it as a flag in all fundamentals-related method (`get_financials(..., as_dict=True)`)

Issue #102, opened 2019-09-19
Excessive HTML requests used, additional easily retrievable data not gotten


GregoryMorse:
The app may want to stop using and relying on Pandas tables for the financials and info as all of these can be fetched with a single call with a serious amount of additional information.

I provide the very straight-forward and simple parsing code to achieve this effect:
```
scrape_url = 'https://finance.yahoo.com/quote'
url = '%s/%s/%s' % (scrape_url, StockName, 'financials')
req = requests.get(url=url)
idx = req.text.find('root.App.main = ')
j = json.loads(re.sub('root.App.main = ', '', req.text[idx:].split('\n')[0])[:-1])
j['context']
j['context']['dispatcher']['stores']['StreamDataStore']['quoteData'][StockName] #contains Ticker.info
j['context']['dispatcher']['stores']['QuoteSummaryStore'].keys()
dict_keys(['cashflowStatementHistory', 'balanceSheetHistoryQuarterly', 'earnings', 'price', 'incomeStatementHistoryQuarterly', 'incomeStatementHistory', 'balanceSheetHistory', 'cashflowStatementHistoryQuarterly', 'quoteType', 'summaryDetail', 'symbol', 'pageViews'])
```

The rest of the puzzle to parse it into the table data (so you get the correct ordering and correct string captions) can be found in this javascript file:
https://s.yimg.com/uc/finance/dd-site/js/Quote.financials.938c6b86ad7b69ba7927.min.js

nono-london:
Thanks for the post, very useful.
One thing i have noted is that when using proxy, the json, is sometimes truncated and does not contain the information part: j['context'] etc.
I guess maybe y! notices robot activity, or maybe it is due to poor quality of the pool of proxies i used...
Best

GregoryMorse:
This already fully implemented in #104 

Your problem is probably not the proxy, Yahoo is browser header sensitive since it wants to make a compatible site.  So when using requests.get try:
my_headers = { 'User-agent' : 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.102 Safari/537.36 Edge/18.18362' }

nono-london:
Great help, thanks, look like it solved the problem.
Best

nono-london:
Hi,
I saw your code proposal for this app, but until it is accepted, I am not sure how I can use it.
So far i use the above code with loops (which doesn't look optimal...).
[...]

Issue #90, opened 2019-08-18
Support for quarterly financial data


martinbel:
It would be great to be able to get the quarterly financial data together with the yearly data simply by adding more columns to the DataFrame.

ps12345678:
How does one get the quarterly data, even if its separate data frame

ps12345678:
> It would be great to be able to get the quarterly financial data together with the yearly data simply by adding more columns to the DataFrame.

How does one get Quarterly Financial. Even as a separate dataframe

GregoryMorse:
Right now it uses pandas to get the only table there.  The quarterly simply uses JavaScript to pull out the data from a massive object on the web page.  Its pretty easy - need to find the <script> tag with this big object and make it JSON compatible which it probably already is then simply use JSON libs to parse it, and finally pull out the quarterly data.  Its not a large task, and a lot of other information is in the big JavaScript object as well including the normal info plus some more.

nono-london:
HI,
which inspector do you use to isolate/find this <script> tag or the
relevant url access to the JSON libs?
At the moment i use Chrome, and can only find the related button but not
the associated JavaScript...
Thanks!

GregoryMorse:
I simply as shown parsed the whole HTML blob as text for 'root.App.main = ' - then take the whole line and trim off that text and the semicolon.  It is a valid JSON string fortunately even though its technically JavaScript which sometimes has some subtle parsing differences with things like encoded characters e.g. single quotes.

Finding the script may be possible.  However you would probably need a very good HTML parser - and I found mostly in Python it was not possible using standard ones.  Mainly because modern websites design HTML for backwards compatibility and have non-perfect non-compliance issues throughout.  Hence the script tags are often not found as a starting point.  So just parsing the raw text blob is sufficient.  If you want to find it via script tags then certainly find a worthy HTML parser.

nono-london:
Hi Gregory,
I have now located the blob on the source page. Nevertheless, although i
manage to access the source code with requests as text using python(which
does show the full information as you mentioned), when i try to parse it
with Beautifulsoup, the root.App.main blob get altered and relevant
information doesn't show anymore.
I used 'lxml' and 'html.parser' without any success.
How do you point your program to select the right bit of html?
Thanks a lot for the help

GregoryMorse:
See #104 where the PR for this has the full code.

[...]

PR #104, opened 2019-09-22
Quarterly/yearly financials/cashflow/bsheet in 1 fetch

GregoryMorse:
The idea was already present in _get_fundamentals and here it is properly done out.  However, probably the _get_financial_translator routine should just be cached and hard-coded and updated as needed since those 2 JavaScript fetching calls are likely rarely ever changed and should at least be globally cached per day, etc.


ranaroussi:
What are you trying to accomplish here?

GregoryMorse:
> What are you trying to accomplish here?

1) Now the quarterly (not only yearly) financial data is retrieved and put in a Pandas Dataframe.

2) This also fetches all 3 financial data: cash flow, balance sheet and overall financials both yearly (which are done with 3 separate HTML retrievals) in one single page request.  However the method is a bit complicated since another JavaScript file contains a translation table to dig out the language strings.

3) It limits the amount of HTTP requests to Yahoo (though the JavaScript request for lookup table could just be hardcoded more than likely to avoid that whole _get_financial_translator function but it should be left as a code comment so it can be updated easier if it changes in the future).

Is it not working currently?

GregoryMorse:
Solves #90 and #102

GregoryMorse:
It is now timezone-aware, converts to float, uses efficient and correct, equivalent to other code datetime conversion with Pandas.

PR #156, opened 2019-12-31
renamed stock to msft in readme

guscht:
"stock" is not defined in the readme. the variable is called "msft".


mfrancis95:
LGTM.

Issue #66, opened 2019-05-25
Proxy use


nono-london:
Hi,
great work around the api and it is working well.
One think i was thinking to add would be use of proxies via request/session.
Best

ranaroussi:
done! just specify your https proxy using:

``` python
msft = yf.Ticker("MSFT")
msft.history(period="max", proxy="https://10.10.1.10:1080")
```

or via:
``` python
df = ys.download("MSFT", period="max", proxy="https://10.10.1.10:1080")
```

nono-london:
Hi,
Thanks for the update, but it seems that there is an issue due to the renaming..
I can' install it anymore, nor update...

Collecting fix-yahoo-finance
  Using cached https://files.pythonhosted.org/packages/a5/f3/aba3b563452faaea71b4161754991e3143751ec9141d83c4bce05992332a/fix-yahoo-finance-0.1.37.tar.gz

    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Users\%%\AppData\Local\Temp\pycharm-packaging\fix-yahoo-finance\setup.py", line 19, in <module>
        from fix_yahoo_finance import __version__
      File "C:\Users\%%\AppData\Local\Temp\pycharm-packaging\fix-yahoo-finance\fix_yahoo_finance\__init__.py", line 22, in <module>
        import yfinance
    ModuleNotFoundError: No module named 'yfinance'
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in C:\Users\%%%\AppData\Local\Temp\pycharm-packaging\fix-yahoo-finance\

ranaroussi:
Hi

Just use:

```
pip install yfinance
```

And then use:

```
import yfinance as yf

# do stuff here
...
```

On 26 May 2019, at 19:13, nono-london <notifications@github.com> wrote:

Hi,
Thanks for the update, but it seems that there is an issue due to the
renaming..
I can' install it anymore, nor update...

Collecting fix-yahoo-finance
Using cached
https://files.pythonhosted.org/packages/a5/f3/aba3b563452faaea71b4161754991e3143751ec9141d83c4bce05992332a/fix-yahoo-finance-0.1.37.tar.gz

ERROR: Complete output from command python setup.py egg_info:
ERROR: Traceback (most recent call last):
  File "<string>", line 1, in <module>
[...]

PR #157, opened 2020-01-01
Added missing proxy parameter

zhenwenc:
Downloading ticker history with `proxy=True` while setting `thread=False` won't work due to the `proxy` parameter doesn't get propagated to the subroutine function. See [this issue comment](https://github.com/ranaroussi/yfinance/issues/66#issuecomment-565692359).


zhenwenc:
Sorry I am not a Python developer, not sure what the CodeFactor issue means... would be great if anyone could give me a hand. Thanks!

PR #176, opened 2020-01-17
Hour interval should return date and time

rdiere:
#Fixes 125


kwrobert-work:
Ah, just ran into this bug now and was about to make my own pull request until I found this. +1,  I think this is an important fix and the modifications are so small it should be an easy merge.

silvavn:
This is deprecated since yahoo no longer offers data in intervals smaller than 1d.

PR #187, opened 2020-01-28
Hotfix/holders logic

jxyong50:
add condition to check that data for major holders and institutional holders are being returned by api
also add condition to check that there is data for institutional holders before inspecting further data


bradmetz:
solved in PR #459

PR #210, opened 2020-03-02
Add lxml package to requirements.txt

samarth261:
When I  installed this via pip, it complained that lxml wasn't found.


samarth261:
I saw the error message being thrown when I created a 'Ticker' object and printed the ticker.info or another field.

kubbing:
I can confirm, same thing happened to me.
Fresh macos Catalina and fresh python3.8 from from homebrew.

bradmetz:
fixed with merge of #373

PR #220, opened 2020-03-13
fix index error when calling ticker's info

haoshuji:
if the holders returned is a list of data frame, we should extract it before using it, otherwise there will be a index error


bradmetz:
solved in PR #459

Issue #234, opened 2020-03-30
Financial statements not working


mchandlernz:
Hello,

It seems the financial statements (income statement, bal sheet, cashflow) are not working as intended.

It's not showing an error, but is not returning any data.

jabarny:
> Hello,
> 
> It seems the financial statements (income statement, bal sheet, cashflow) are not working as intended.
> 
> It's not showing an error, but is not returning any data.

https://github.com/ranaroussi/yfinance/issues/191

check here.

Issue #191, opened 2020-02-01
Income Statement, Balance Sheet and Cashflows are coming up as an empty dataframe


karthikbhamidipati:
Income Statement, Balance Sheet and Cashflows are coming up as an empty data frame. 
This is because variable url in base.py is being updated to <address>/quote/<Ticker>/holders in line 283
and when fetching the financials in line 354 of base.py the url being used is <address>/quote/<Ticker>/holders/financials and it should be <address>/quote/<Ticker>/financials

masueror:
I also have the same problem, when i ask for financials I recibe:
Empty DataFrame
Columns: [Open, High, Low, Close, Adj Close, Volume]
Index: []
How can I fix it?

TimBstn:
Hey, I just had the same problem.
The problem is as vamshikarthik described it: Just change the following in the base.py:

 `url = "{}/{}".format(self._scrape_url, self.ticker)`
`holders = _pd.read_html(url+'\holders')`

The problem is in line 283.

Thanks vamshikarthik   for the tip.

Tim

masueror:
Thank you very much, problem just solve

kalinkirev:
It did not work for me I'm sorry. This applies for balance_sheet, cashflow, earnings, financials etc.

```
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
<ipython-input-8-3cf019396317> in <module>
----> 1 amzn.balance_sheet

~/anaconda3/envs/stok/lib/python3.7/site-packages/yfinance/ticker.py in balance_sheet(self)
    164     @property
    165     def balance_sheet(self):
--> 166         return self.get_balancesheet()
    167 
    168     @property

~/anaconda3/envs/stok/lib/python3.7/site-packages/yfinance/base.py in get_balancesheet(self, proxy, as_dict, freq)
    441 
    442     def get_balancesheet(self, proxy=None, as_dict=False, freq="yearly"):
--> 443         self._get_fundamentals(proxy)
    444         data = self._balancesheet[freq]
    445         if as_dict:

~/anaconda3/envs/stok/lib/python3.7/site-packages/yfinance/base.py in _get_fundamentals(self, kind, proxy)
    282         # holders
    283         url = "{}/{}/holders".format(self._scrape_url, self.ticker)
--> 284         holders = _pd.read_html(url+'\holders')
    285         self._major_holders = holders[0]
    286         self._institutional_holders = holders[1]

[...]

PR #246, opened 2020-04-07
Fixes the scrape URL for financial data

motasay:
Hi,

Thanks for the helpful library!

The `url` variable holds the value `https://finance.yahoo.com/quote/TICKER/holders` as set in line [283](https://github.com/ranaroussi/yfinance/blob/master/yfinance/base.py#L283). So the result of line [354](https://github.com/ranaroussi/yfinance/blob/master/yfinance/base.py#L354) is `https://finance.yahoo.com/quote/TICKER/holders/financials` which is incorrect. The code before the changes was not retrieving all the following properties as a result of the incorrect URL:

* `self._cashflow`
* `self._balancesheet`
* `self._financials`
* `self._earnings`

All of the above are retrieved fine after the small change to the URL. Let me know what you think please.

Cheers,
Mo

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -351,7 +351,8 @@ class TickerBase():
             pass
 
         # get fundamentals
-        data = utils.get_json(url+'/financials', proxy)
+        financials_url = "{}/{}/financials".format(self._scrape_url, self.ticker)
+        data = utils.get_json(financials_url, proxy)
 
         # generic patterns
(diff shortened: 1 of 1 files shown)


motasay:
Apparently there are many issues related to this - sorry just saw them!
#234 #191

robalar:
Can we get this merged as it is a super essential feature?

PR #251, opened 2020-04-11
Time zone conversion error in ticker option

oeclint:
Previous conversion of datetime column to other time zone did not work. Also maybe consider defaulting to America/New_York. As that is what is used by default for stocks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ branches:
     - master
 
 notifications:
+  slack: tradologics:HcnS6XusfcuS02waQPCG18oc
   webhooks:
     on_success: change  # options: [always|never|change] default: always
     on_failure: always  # options: [always|never|change] default: always


oliverlewis:
Agree to your suggestion of defaulting to "America/New_York"

yfinance/ticker.py
```diff
@@ -81,7 +81,7 @@ def _options2df(self, opt, tz=None):
         data['lastTradeDate'] = _pd.to_datetime(
             data['lastTradeDate'], unit='s')
         if tz is not None:
-            data['lastTradeDate'] = data['lastTradeDate'].tz_localize(tz)
+            data['lastTradeDate'] = data['lastTradeDate'].dt.tz_localize(tz)
```
oliverlewis, line 84: Hi, This is incorrect. Because the lastTradeDate is localized to utc only.

```
data['lastTradeDate1'] = pd.to_datetime(
    data['lastTradeDate'], unit='s', utc=True)
```

After localizing it to UTC. You then need to convert it to the correct timezone that you require
```
if tz is not None:
    data['lastTradeDate1'] = data['lastTradeDate1'].dt.tz_convert(tz)
```

Issue #259, opened 2020-04-16
.info fails to retrieve any info as it produces IndexError


Qrishna:
Hi,

I have noticed that the module fails to retrieve info for certain stocks(symbols) even though they are clearly available on the yahoo finance page. They all return an Index error as shown below:

```$ python
Python 3.7.6 (default, Dec 30 2019, 19:38:28) 
[Clang 11.0.0 (clang-1100.0.33.16)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import yfinance as yf
>>> my_stock = yf.Ticker("MMM")
>>> my_stock.info
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/yfinance/ticker.py", line 138, in info
    return self.get_info()
  File "/usr/local/lib/python3.7/site-packages/yfinance/base.py", line 415, in get_info
    self._get_fundamentals(proxy)
  File "/usr/local/lib/python3.7/site-packages/yfinance/base.py", line 286, in _get_fundamentals
    self._institutional_holders = holders[1]
IndexError: list index out of range
```

Here are a few stocks that I have tried that currently fail: MMM, CVX, XOM, JPM, MRK, TRV, UNH, VZ 

Does anybody have any idea why this happens only for certain stocks? Does the yahoo finance page load differently for different stocks? If not, is there a bug in the .info functionality? It seems like perhaps an assumption was made to get the second index `self._institutional_holders = holders[1]`, does this need to be handled properly? 

Thanks!

rodrigobercini:
Hey there,

I've added a "try" error handler for these cases in my fork, I've just tested and it seems to be working. I highly recommend uninstalling yfinance before doing so.
pip uninstall yfinance
pip install git+https://github.com/rodrigobercini/yfinance.git

ekinig:
Same here. Any chance to fix it?

asafravid:
**This happens for me as well - I avoid it using try + exception
However, this is just so the script can continue to the next ticker
What occurs effectively is that information is Lost!
Any chance this can be fixed? Or any solution known?**

asafravid:
I fixed this:

In lines 286..294 simply replace as follows:
`        if len(holders) > 1:
            self._institutional_holders = holders[1]
            if 'Date Reported' in self._institutional_holders:
                self._institutional_holders['Date Reported'] = _pd.to_datetime(
                    self._institutional_holders['Date Reported'])
[...]

PR #262, opened 2020-04-17
issues 259

Qrishna:
https://github.com/ranaroussi/yfinance/issues/259


bradmetz (changes requested):
Hi there. 

These changes look good but can you please remove the .DS_Store from the PR?  I will merge once that change is made.  Thanks!

Brad

PR #272, opened 2020-04-25
Fix empy balance_sheet

hristo-mavrodiev:
Fix the empty balance sheet and add a test to check for empty dataframes.


tranthuc-org:
Good

bradmetz:
holders issue is fixed in PR #459

Issue #277, opened 2020-04-29
RegularMarketPrice si the same as RegularMarketOpen


eto333:
Hi
I have noticed, that the "Regular Market Price" is equal to the "Regular Market Open".
Presumably it is because of the code in base.py, line 318:
self._info['regularMarketPrice'] = self._info['regularMarketOpen']

Is there a possibility to correct this and to have the "Regular Market Price"?

Thank you in advance!

romzie:
+1 only looked at this project because of this value but it is unavailable...

hex:
+1 Any fix to this?

alex4108:
PR 299 fixes this: https://github.com/ranaroussi/yfinance/pull/299

mediocre-code:
How long do PRs usually take for this repo? Should we fork it and add the fix ourself like others have suggested?

alex4108:
Given that the PR hasn't been acknowledged, I would @dat-mediocre-data

dpguthrie:
You can take a look at a similar package I wrote, [yahooquery](https://github.com/dpguthrie/yahooquery).  That data can be obtained pretty simply:

```
from yahooquery import Ticker

aapl = Ticker('aapl')
aapl.price
```

```
{
  "aapl": {
    "maxAge": 1,
    "preMarketChangePercent": 0.00975431,
    "preMarketChange": 3.72,
    "preMarketTime": "2020-07-09 13:29:59",
    "preMarketPrice": 385.09,
    "preMarketSource": "FREE_REALTIME",
    "postMarketChangePercent": 0.0019856798,
    "postMarketChange": 0.75997925,
    "postMarketTime": 1594339198,
    "postMarketPrice": 383.49,
    "postMarketSource": "FREE_REALTIME",
    "regularMarketChangePercent": 0.0035661324,
    "regularMarketChange": 1.3600159,
    "regularMarketTime": "2020-07-09 20:00:01",
    "priceHint": 2,
    "regularMarketPrice": 382.73,
    "regularMarketDayHigh": 385.27,
    "regularMarketDayLow": 378.7,
    "regularMarketVolume": 30439246,
    "regularMarketPreviousClose": 381.37,
    "regularMarketSource": "FREE_REALTIME",
    "regularMarketOpen": 385.05,
    "exchange": "NMS",
    "exchangeName": "NasdaqGS",
    "exchangeDataDelayedBy": 0,
    "marketState": "POSTPOST",
    "quoteType": "EQUITY",
[...]

PR #299, opened 2020-05-09
Map regular market price to current price

alex4108:
Fixes Issue #277: regularMarketPrice is reported as open price instead of current price.


7AC:
Looks like the CodeFactor issues pre-date this change, can we get this merged?

bradmetz:
solved in PR #459

PR #305, opened 2020-05-14
fixed _get_fundamentals in base

simoncai519:
fixed bug when no institutional holder info
fixed get fundamental method, previously contaminated by previous code redefining url variable


bradmetz:
fixed with merge of PR #459

PR #327, opened 2020-06-03
Fix "/" in tickers causing .download() to be stuck

### LINKED ISSUE #326: Download is stuck at X/Y completed
With my ticker list download does not always work properly. The problem seems to be caused by some of them containing a "/". Using yf.Ticker() and .history() on such tickers gives JSONDecodeError: Expecting Value. Having such a ticker in download list causes the progress to be stuck at some point.  What I found is that in both cases the problem arises at line 155 of base.py where the response is expected to be in JSON format but having a "/" in a ticker gives 404 error page and it is in XML format. For history() it just raises an error and stops executing but in download it gets stuck in a while loop at line 94 of multi.py.
Code example:

```
import yfinance as yf

stocks = ['XOM', 'IBM', 'GE', '3891289Q', 'F', 'CVX', 'BLS', 'AIG', '2196Q', 
               '653707Q', '2500Q', 'MO', '1715651D', '571300Q', '210967Q', '605555Q', 
               'BRK/A', 'WMT', '677004Q', '2200Q', 'BA']
data = yf.download(stocks, start="1993-01-01", 
                                end="2019-12-31")["Adj Close"]
print(data)
```

kdergachev:
Resolves #326 
If the respoonse to a request is 404 the format is XML. It starts with "<" and block of code added catches responses starting with "<" and adds them to error list. This seems to solve .download() being stuck. In the final result tickers with "/" are added to the dataframe with NaNs for the whole period and errors are displayed at the end saying that "/" tickers could have caused problems.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -152,14 +152,21 @@ class TickerBase():
             raise RuntimeError("*** YAHOO! FINANCE IS CURRENTLY DOWN! ***\n"
                                "Our engineers are working quickly to resolve "
                                "the issue. Thank you for your patience.")
-        data = data.json()
-
+        
+        # 404 error on request give XML formatted response. If the first 
+        # element is "<" XML assumed and error message is added
+        if data.text[0] == "<":
+            err_msg = "404 returned. Likely a ticker including '/'"
+        else:
+            data = data.json()
+            err_msg = "No data found for this date"
+            "range, symbol may be delisted"
+            
         # Work with errors
         debug_mode = True
         if "debug" in kwargs and isinstance(kwargs["debug"], bool):
             debug_mode = kwargs["debug"]
-
-        err_msg = "No data found for this date range, symbol may be delisted"
+            
         if "chart" in data and data["chart"]["error"]:
             err_msg = data["chart"]["error"]["description"]
             shared._DFS[self.ticker] = utils.empty_df()
@@ -504,7 +511,7 @@ class TickerBase():
         search_str = '"{}|'.format(ticker)
         if search_str not in data:
             if q.lower() in data.lower():
-                search_str = '"|'
+                search_str = '"|'.format(ticker)
                 if search_str not in data:
                     self._isin = '-'
                     return self._isin


bradmetz:
Do you have an example of a ticker that contains "/"?  I am trying to figure out if this is an issue of handling a 404 or if there are actually ticker symbols with "/" characters in them.  Thanks

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

kdergachev:
I don’t really know any tickers with “/“, guess it was more of compatibility problem as I tried to use Yahoo finance API to get prices for tickers that I got from Bloomberg terminal and supposedly where on yahoo they use “-“ on Bloomberg they use “/“.

bradmetz:
So I just looked at the issue (should have done that in the first place, sorry).  You had BRK/A which I assume is for BRK.A.  So I can confirm that the convention for yahoo is to use a "-" to replace "." in the ticker.  This is not the case when appending the exchange to a ticker where the convention is to use a '.' (example: BB.TO for Blackberry on the TSX).  Normalizing tickers across multiple sources is a problem I have run into in my other projects and I thought about writing up a small utility to help out (google notation is different as are some of the paid for APIs in how they write their tickers with exchanges).  But I think its outside the scope of a library for accessing yahoo finance to account for all these ticker notations.  What I did was add a small curr_sym = curr_sym.replace('.', '-') line to fix the symbol before the call.  If that makes sense I will close this PR.

kdergachev:
The 404 error as I understand comes only/mainly from "/" symbol so I added it to err_msg to point at that problem for whoever uses the package, but I guess your change would also help with the issue so closing both PR and the issue is fine. 

That is if you meant curr_sym = curr_sym.replace('/', '-')

bradmetz:
For your case yes you would replace the /.  Closing now.

Issue #314, opened 2020-05-20
Get info return - IndexError: list index out of range


yankiShechter:
For example:
yahoo_finance = yf.Ticker('VZ')
yahoo_finance.info

Return:
Traceback (most recent call last):
  File "C:\...\Testing\yahooFinance.py", line 98, in <module>
    yahoo_finance.info
  File "C:\...\Testing\venv\lib\site-packages\yfinance\ticker.py", line 138, in info
    return self.get_info()
  File "C:\...\Testing\venv\lib\site-packages\yfinance\base.py", line 415, in get_info
    self._get_fundamentals(proxy)
  File "C:\...\Testing\venv\lib\site-packages\yfinance\base.py", line 286, in _get_fundamentals
    self._institutional_holders = holders[1]
IndexError: list index out of range

Process finished with exit code 1

Best regards

rodrigobercini:
The base code needs an error handler for cases where there are no institutional holders. There are some PRs that could solve this issue, but while they are not merged, you can install my fork (don't forget to uninstall yfinance before and restart the runtime):
pip uninstall yfinance
pip install git+https://github.com/rodrigobercini/yfinance.git

I've just tested for VZ and it's working fine. Cheers!

![image](https://user-images.githubusercontent.com/18664907/82485559-02cbfa80-9ab2-11ea-8255-cd7419fe7cfb.png)

yankiShechter:
Thank you Rodrigo!
I'll try it.

PR #337, opened 2020-06-11
holders length not always 2 (i.e. Ford stock)

LazaroHurtado:
Yahoo Finance doesn't always provide "Top Institutional Holders" data, an example of this would be Ford stock ($F). This results in a "list index out of range" error when calling ".info" on a stock like Ford.
To fix this issue a small if statement was added to check if the stock does have the "Top Institutional Holders" data.


haansn08:
This would fix https://github.com/ranaroussi/yfinance/issues/314

bradmetz:
fixed with merge of PR #459

PR #345, opened 2020-06-18
avoid exception when no institutional holders

datiti:
for small companies, there are often no institutional holders


bradmetz:
solved in PR #459

Issue #189, opened 2020-01-30
IndexError for Tickers w/ no institutional holders in .info


jsultan2323:
Running the following code:
`deo = yf.Ticker('deo')`
`deo.info`
Produces the following error:
![image](https://user-images.githubusercontent.com/47575677/73489778-d2304b80-4370-11ea-81f4-e6cce268adaf.png)
This is only an issue when the Ticker does not have institutional holders on the "holders" page due to holders[1] having no value and causing _the list index out of range_ error.

Inevitably, it flows through to the info function and I haven't been able to dig deep enough to see how to correct it. 

I assumed just changing the index value to [0] would fix it or using a try/except statement, but to no avail.

Also, the condition of len(holders)>1 as in the pull request does not seem to work for me.

PeterChea:
This appears to occur when there is no "Top institutional holders" section under the "holders" and "major holders" page of a given ticker.

It's a very big issue causing around a third of my tickers to error.

cstampar:
I am having this exact same problem and haven't been able to find a solution yet.

PR #365, opened 2020-06-29
added conditional in case the stock doesn't list institutional holders fixes #159 and #189

### LINKED ISSUE #159: info IndexError for some Tickers
Hello,

In version 0.1.54, some symbols/tickers error out when their info is downloaded as seen below (one of these tickers is TWER).

F34AA2C1-E7C9-4FEB-953F-D7F1C613D2F3

keegan2149:


bradmetz:
fixed with merge of PR #459

PR #389, opened 2020-07-19
Timestamp fix when downloading option chains

ed353:
If using this code in a timezone other than UTC, timestamp strings created when querying option chains in ticker.py may be off by a day.

Fixed by using `datetime.datetime.utcfromtimestamp` instead of `datetime.datetime.fromtimestamp`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ yfinance/__pycache__/*
 dist
 yfinance.egg-info
 *.pyc
+*.swp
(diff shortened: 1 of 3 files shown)


xbno:
Thanks for fixing this! I ran into the exact same thing locally on mine, and came here to submit the same MR. I don't know anything about the checks that were unsuccessful but I'd like to bump this to be merged - its essential for pulling options data.

PR #396, opened 2020-07-25
Indian markets added

ratnarajsingh:
There were some issues and the package wasn't working. This was due to change in URL and some structural changes in the content of financials. Fixed them

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ yfinance/__pycache__/*
 dist
 yfinance.egg-info
 *.pyc
+*.ipynb*
+build/*
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -46,7 +46,8 @@ class TickerBase():
         self.ticker = ticker.upper()
         self._history = None
         self._base_url = 'https://query1.finance.yahoo.com'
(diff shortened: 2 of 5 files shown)


aswinme:
Hi ratnarajsingh, I'm new to github and i was waiting for yfinance to add indian stocks. Now i saw your pull request waiting for approval, my query is can i use your modified code ? I couldnt find any resources googling.please guide me

Rambatino:
https://stackoverflow.com/questions/20101834/pip-install-from-git-repo-branch

bradmetz:
Hi there,

So I looked into this since I found it odd that the URL would be the issue.  yahoo adds a geo prefix to their url depending on where you are but that should have no effect on the functionality of the library. Example, when I access yahoo finance it appends ca to the beginning since im in Canada but the library still functions.  I did, however, find a flawed assumption in the recent push to fix the institutional investors issue.  Example: ticker symbol: ZEEL.NS shows that you can have major holders and mutual fund holders with not institutional investors.  I will write up a fix this afternoon to account for this case.  In general, if you append the .NS suffix to the ticker on the NSE it should work fine. 

So I am not going to incorporate this change but there should be a new release out shortly that deals with the institutional investors issue.  
Thanks

PR #397, opened 2020-07-25
Fixed a bug that ocurrs when a ticker doesn't have institutional holders.

tavito:
I was getting an error for a ticker that in Yahoo Finance does not have institutional holders in this line: 

`self._institutional_holders = holders[1]`


bradmetz:
fixed with merge of PR #459

PR #406, opened 2020-08-02
Cofirming length of holders list prior to accessing it

RedSpiderMkV:
When I was using this to retrieve data for the symbol, MCRO.L (Micro Focus International), I was getting an exception.  This was due to the holders list having fewer elements than expected so an index out of range exception was being thrown.

I've modified it so that it checks the length of the list prior and only setting it if the index exists.

Additionally, it might be worth adding MCRO.L to the list of symbols the tests run on.

Thanks!

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -379,8 +379,8 @@ class TickerBase():
 
     def _set_institutional_holders(self, holders):
         if len(holders) <= 1:
-	    return
-			
+            return
+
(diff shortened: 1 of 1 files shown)


RedSpiderMkV:
Sorry! New installation of editor didn't have tabs to spaces conversion set...

bradmetz:
fixed with merge of PR #459

PR #428, opened 2020-08-20
Financials fix

alienbrett:
Fixes a few issues with the balance sheet and financials. I don't take credit for the first commit, that would be the solutions posted in issue threads.


bradmetz:
solved in PR #459

Issue #416, opened 2020-08-11
list index out of range


jonathanli2012:
`tick1 = yf.Ticker("ABC")
print(tick1.info)`

`    print(tick.info)
  File "/home/jon/.local/lib/python3.8/site-packages/yfinance/ticker.py", line 138, in info
    return self.get_info()
  File "/home/jon/.local/lib/python3.8/site-packages/yfinance/base.py", line 415, in get_info
    self._get_fundamentals(proxy)
  File "/home/jon/.local/lib/python3.8/site-packages/yfinance/base.py", line 286, in _get_fundamentals
    self._institutional_holders = holders[1]
IndexError: list index out of range`

pieterjandotpy:
I had the same issue and was able to fix this by inserting an if statement like this at line 286:

        if len(holders) > 1:
            self._institutional_holders = holders[1]
            if 'Date Reported' in self._institutional_holders:
                self._institutional_holders['Date Reported'] = _pd.to_datetime(
                    self._institutional_holders['Date Reported'])
            if '% Out' in self._institutional_holders:
                self._institutional_holders['% Out'] = self._institutional_holders[
                    '% Out'].str.replace('%', '').astype(float)/100

davidchisnall:
I can confirm the above fix works. Without this, some high profile things can’t be accessed (e.g. BRK-B). Is it possible to have a new release with this fix soon?

Issue #412, opened 2020-08-06
Moscow market


sepy97:
Hi! thanks for the lib, its great! Can't get data from Moscow market (e.g. MOEX.ME, ROSN.ME) though :( 
installed the library through pip, those tickers are presented on yahoo finance website...

> line 286, in _get_fundamentals
>     self._institutional_holders = holders[1]
> IndexError: list index out of range

Thank you in advance

pieterjandotpy:
It has nothing to do with the markets in Moscow, see my fix here: #416

Issue #405, opened 2020-08-01
LSE institutional holders


as595:
Hi, I was having an issue with some TIDMs from the LSE (example: OPM.L) because the institutional holders are not listed in the returned data, which made the base Ticker class crash on this line:

self._institutional_holders = holders[1]

because len(holders)=1.

I put in a quick fix locally, but it might be a problem other people are seeing too so I just wanted to flag it up.

as595:
I should have refreshed the issues list before posting this - I see others have already raised it - sorry!

Issue #401, opened 2020-07-29
stock.info errors out on certain tickers


gyansinha:
# MWE

```python
import pandas as pd, yfinance as yf
stock = yf.Ticker("JPM")
stock.info["marketCap"]

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-64-4f51d5f66618> in <module>
      1 import pandas as pd, yfinance as yf
      2 stock = yf.Ticker("JPM")
----> 3 stock.info["marketCap"]

~/admin/db/dev/Python/projects/venv/lib/python3.6/site-packages/yfinance/ticker.py in info(self)
    136     @property
    137     def info(self):
--> 138         return self.get_info()
    139 
    140     @property

~/admin/db/dev/Python/projects/venv/lib/python3.6/site-packages/yfinance/base.py in get_info(self, proxy, as_dict, *args, **kwargs)
    413 
    414     def get_info(self, proxy=None, as_dict=False, *args, **kwargs):
--> 415         self._get_fundamentals(proxy)
    416         data = self._info
    417         if as_dict:

~/admin/db/dev/Python/projects/venv/lib/python3.6/site-packages/yfinance/base.py in _get_fundamentals(self, kind, proxy)
    284         holders = _pd.read_html(url)
    285         self._major_holders = holders[0]
--> 286         self._institutional_holders = holders[1]
    287         if 'Date Reported' in self._institutional_holders:
    288             self._institutional_holders['Date Reported'] = _pd.to_datetime(

IndexError: list index out of range
```

ahf90:
Somebody already has an open pull request to fix the issue: https://github.com/ranaroussi/yfinance/pull/397
The problem is that some stocks don't have "Institutional Holders".  For example, [TSLA does](https://finance.yahoo.com/quote/TSLA/holders?p=TSLA), but [MMM does not](https://finance.yahoo.com/quote/MMM/holders?p=MMM).

PR #439, opened 2020-09-01
Prevents accessing an empty list of holders

### LINKED ISSUE #438: Retrieve data from ETR instead of NASDAQ
This more a questions: if I retrieve MSFT shares data its ok. If I try MSF, which is not on NASDAQ but on ETR (xetra, German) it finds nothing.

jc0b:
This fixes #438, #416, #412, #405, and #401 (among others).

This issue has been widely reported, and the proposed PR would allow querying of tickers without institutional holders.


jc0b:
It looks like #430 is a better fix for these issues. Hopefully it can be merged soon!

PR #442, opened 2020-09-03
Bugfix/ get fundamentals

andvikt:
fix: IndexError when there is less that 2 holder (yf.Ticker('MMM').info)
fix: lxml must be included in requirement


hulaud (commented):
FYI:  A try / catch might be an overkill for this. 
You could just replace line 286 with
self._institutional_holders = holders[1] if len(holders) > 1 else {}

bradmetz:
solved in PR #459

PR #449, opened 2020-09-11
Switched to selenium and google chrome for scrapping

beglitis:


dpuljic01:
Hey, what's your reason for switching to selenium from requests? I don't think we should do that, isn't it going to be slower than pure `requests`?

beglitis:
@dpuljic01 maybe yes but since a couple of days ago requests() had stopped working for me and others. The hypothesis was that Yahoo would pick up we are scrapping their site and block us. Selenium and Chrome headless present themselves as real browsers and are more difficult to detect. Indeed switching to selenium and Chrome headless made the problem go away.

dpuljic01:
@beglitis I see the issue now, nice work!

schlunsen:
Have you considered just changing the USER-AGENT on the request?

beglitis:
> Have you considered just changing the USER-AGENT on the request?

Yes and it didn't work for me for some reason. I guess Yahoo might be looking at multiple data points for deciding whether I am scrapping their site including source IP, geo-location etc. For me switching to selenium and chrome worked.

PR #488, opened 2020-10-26
Fix incorrect financials url

JakeRuth:
Hi,

First off, this is a great package thanks! I am using it for some novice/fun data mining for stock investing stuff.

I noticed this bug playing around with the lib, I was getting no data returned when calling: `yf.Ticker('MSFT').balance_sheet`

Upon further investigation is seems the url that was being called was still set to the values needed for holders resulting in a 404, the url being called ended up being: https://finance.yahoo.com/quote/APHA/holders/financials (before fix in this PR)

Fix here seems to work/data comes back now after changing the url with this one liner.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -283,7 +283,7 @@ class TickerBase():
         # holders
         url = "{}/{}/holders".format(self._scrape_url, self.ticker)
         holders = _pd.read_html(url)
-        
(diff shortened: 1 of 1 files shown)


JakeRuth:
The checks here don't seem to be related to my diff.  Is this okay?

Issue #490, opened 2020-10-27
1h & 60m time interval


odufb:
hi all,

I get data from 1h & 60m time intervals in  0830-0930 format. Is it possible to get data  in 0800-0900 format. 
I hope question is meaninful 
best regards,

KennSmithDS:
Hi All,

I'm not sure if this a related issue, but when I try to pull anything other than daily data, the dataframe index values returned are not unique and all still in the date format, not in datetime with the hour or minute granularity.  Also I am unable to readjust them to the appropriate datetime with hour format.

For example:

`df_test = yf.download(tickers='NFLX', period='100d', interval='1h')`
`df_test.tail(20)`

![nflx_date_not_hour](https://user-images.githubusercontent.com/30271891/97770700-d69db680-1af2-11eb-8044-51f162d9edbc.PNG)

KennSmithDS:
If I run the comparable syntax for the pandas_datareader, the '1h' still has issues, but the '1m' yields expected data:

`df_test = pdf.get_data_yahoo(tickers='NFLX', period='100d', interval='1m') `

![nflx_pdf_min](https://user-images.githubusercontent.com/30271891/97771020-14500e80-1af6-11eb-875e-ba22b9459193.PNG)

odufb:
@KennSmithDS  please  check issue #125.
my problem is different 
![Adsız](https://user-images.githubusercontent.com/32509250/97774376-4732ea80-1b68-11eb-90bc-870b9ee6bbd5.png)
as you see above picture time interval  like this
 2020-10-26 **09:30**:00-04:00  -  2020-10-26 **10:30**:00-04:00
is it possible get data like
 2020-10-26 **09:00**:00-04:00  -  2020-10-26 **10:00**:00-04:00

thanks for all replies.

mcsoini:
> as you see above picture time interval like this
> 2020-10-26 **09:30**:00-04:00 - 2020-10-26 **10:30**:00-04:00
> is it possible get data like
> 2020-10-26 **09:00**:00-04:00 - 2020-10-26 **10:00**:00-04:00

[...]

PR #499, opened 2020-11-13
Fix datetime index

mcsoini:
Fixes the problem that the time section of the datetime index is discarded when `interval="1h"`
This is mentioned e.g. in  #490 by @KennSmithDS


mcsoini:
Failing CodeFactor check refers to "Complex Method" issues.

thasmin (approved):
I just tested this code and can confirm this adds the time to the index

iSevenDays:
I can confirm this works for me as well

PR #516, opened 2020-11-26
Fix date and datetime interval, ranaroussi#125

highfestiva:
Pandas' index.date drops timezone information, so conversion into
exchange timezone has to happen in intra-day intervals only.


highfestiva:
@ranaroussi It seems the CodeFactor fails unrelated to my fix. Might want to clean that up in a separate commit, or 'lax the settings.

Issue #244, opened 2020-04-06
Expense ratio for ETFs .info returns None


aditya08:
'annualReportExpenseRatio' is working fine for mutual funds but not for ETFs.

```python
Python 3.7.3 [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import yfinance as yf
>>> vti = yf.Ticker('VTI')
>>> vti.info['annualReportExpenseRatio']
>>> print(vti.info['annualReportExpenseRatio']) 
None
>>> vtsax = yf.Ticker('VTSAX')
>>> vtsax.info['annualReportExpenseRatio']
0.0004
```

ycc1107:
The issue code did not go to the `fundProfile` to find `annualReportExpenseRatio`.

PR #557, opened 2021-01-17
(#244) Fix .info['annualReportExpenseRatio'] for ETFs

johnnymo87:
As reported in issue #244, there is a bug with querying for an ETF's
expense ratio. For mutual funds, this value is found in
`.info['annualReportExpenseRatio']`, but for ETFs it is not.

```py
>>> import yfinance as yf
>>> print(yf.Ticker('VTSAX').info['annualReportExpenseRatio'])
0.0004
>>> print(yf.Ticker('VTI').info['annualReportExpenseRatio'])
None
```

The github user @ycc1107 commented on the issue, identifying the root
cause, which is that for some reason this value is only available inside
the `'fundProfile'` key of the `data` dictionary, a key that is only
present when querying mutual funds and ETFs.
The `'fundProfile'` key leads to a dictionary with a
`'feesExpensesInvestment'` key, which leads to a dictionary with an
`'annualReportExpenseRatio'` key, which has the value we need.

The github user @ycc1107 opened [a PR](https://github.com/ranaroussi/yfinance/pull/248) to fix this issue, but I'm
concerned that it changes too much, as it recursively merges the
entirety of the `'fundProfile'` dictionary into the `.info` dictionary.
So I'm proposing an alternative fix here that attempts to write only
`.info['annualReportExpenseRatio']` and nothing else.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -346,6 +346,13 @@ class TickerBase():
         except Exception:
             pass
 
+        expenseRatio = (data
+                       .get('fundProfile', {})
+                       .get('feesExpensesInvestment', {})
+                       .get('annualReportExpenseRatio', None))
+        if expenseRatio is not None:
+            self._info['annualReportExpenseRatio'] = expenseRatio
+
         # events
         try:
             cal = _pd.DataFrame(


carlos-ventura:
MR closed without any changes. Dont think it is to hard to update it as it only requires 4/5 added lines of code and constitues an important info

ValueRaider:
I'm going through old issues/PRs, and see this was closed when looks to me like should have been merged. But codebase has changed a lot since - can someone create a fresh new PR?

PR #577, opened 2021-01-31
Add market state attribute to the info

naveenspace7:
It seemed like the info is missing the market state. I had added one line to obtain this information (holds "CLOSED" or "OPEN").


naveenspace7:
P.S. The above failures have no relation to my change.

naveenspace7:
Later when I used it, it was not working as expected. Closing this for now.

Issue #607, opened 2021-02-23
pdr_get_data doesn't stop after raised run time error


arielbell:
Hi team,

I'm downloading a list of tickets.
In cases that one of the tickers fails to download, the pdr_get_data doesn't stop running and gets stuck.

The error: raise RuntimeError("*** YAHOO! FINANCE IS CURRENTLY DOWN! ***\n"
RuntimeError: *** YAHOO! FINANCE IS CURRENTLY DOWN! ***
Our engineers are working quickly to resolve the issue. Thank you for your patience.

Where it gets stuck: [*********************100%***********************]  51 of 55 completed (there were 4 failed).

Any commendation? I would be happy to continue with whichever tickers that were downloaded.  
I found this [thread](https://github.com/ranaroussi/yfinance/commit/6a8c637af758ff3de53f2ca3dafb6bf10f1902b9) but i'm not sure its related.

Thanks,
Ariel

arielbell:
I believe it gets stuck here: 
while len(shared._DFS) < len(tickers):
            _time.sleep(0.01)

As we don't assign any ticker to the shared._DFS.
Just a  thought from a quick overview

claudinho85:
I'm realizing the same error. 
How can we solve this problem?

Thanks
Claudio

arielbell:
I tried this locally and it worked:
![image](https://user-images.githubusercontent.com/23659488/109061895-e7041d80-76ef-11eb-8e85-6ac67c5db661.png)

I took the error handling from the next section and added it here. 
it doesn't solve the issue of retrieving the data, but it helps the code finish its task as expected.

`if "Will be right back" in data.text:
            #data = data.json()
            err_msg = "No data found for this date range, symbol may be delisted"
            shared._DFS[self.ticker] = utils.empty_df()
            shared._ERRORS[self.ticker] = err_msg
            return shared._DFS[self.ticker]`

claudinho85:
Thanks a lot for your response! :-)
However, what I've realized since yesterday is that I no longer retrieve historical data from just one request using yfinance.

Just to be more detailed, I used pdr.get_data_yahoo(tickers_list) with 5000 tickers in tickers_list and it worked till yesterday.
I dunno if Yahoo recognised this kind of behaviour and blocked it or it is weird that it doesn't work anymore.

How do you retrieve historical data for a huge group of tickers? I need to divide these requests with a time.sleep() between one and the other?

Thanks again

claudinho85:
[...]

Issue #598, opened 2021-02-16
yf.download hangs/stuck downloading last few syms


hansiniupadhyay:
The last few symbols does not get downloaded and the download is stuck with no error message.
Call:
price=yf.download(sym_list,startdate,enddate)

sym_list is a list of 8318 syms
![yfinance stuck](https://user-images.githubusercontent.com/79159854/108094661-0abec800-704d-11eb-97c5-3341c403b5d9.PNG)

hansiniupadhyay:
Looks like this is because the threads are not closing properly for sym data that yfinance wasn't able to download. How can I overcome this and close the threads that didn't close properly since the data wasn't downloaded ?

GF-Huang:
Same issue.

![image](https://user-images.githubusercontent.com/4510984/108501247-f815ee80-72eb-11eb-9b9c-64b21e36e622.png)

hansiniupadhyay:
For now updating the base.py is working for me: Replace below code in base.py

        # Getting data from json
        url = "{}/v8/finance/chart/{}".format(self._base_url, self.ticker)
        data = _requests.get(url=url, params=params, proxies=proxy)
        # if "Will be right back" in data.text:
        qqq=0
        while "Will be right back" in data.text and qqq<3:
            # raise RuntimeError("*** YAHOO! FINANCE IS CURRENTLY DOWN! ***\n"
            #                    "Our engineers are working quickly to resolve "
            #                    "the issue. Thank you for your patience.")
            _time.sleep(2)
            qqq=qqq+1
            data = _requests.get(url=url, params=params, proxies=proxy)
            if "Will be right back" not in data.text:
                break
        if "Will be right back" in data.text:
            raise RuntimeError("*** YAHOO! FINANCE IS CURRENTLY DOWN! ***\n"
                                "Our engineers are working quickly to resolve "
                                "the issue. Thank you for your patience.")
        data = data.json()

hansiniupadhyay:
I wonder what the owner of the library thinks of this. I hope he responds to the thread.

 I figured this was happening because few of the threads were not closing correctly and hence the process not finishing completed. This happens for random syms where the data is not downloaded successfully.

GF-Huang:
I was surprised at how much PRs there was and wondered if the project was still alive.

beglitis:
The project is not really but the community is. Since we are all coders you can imagine we have forked the code and we fix bits and pieces in our own code bases depending on the issues we come across and the findings by others.

[...]

PR #617, opened 2021-03-05
fix infinite wait loop in threaded mode in case of any requests error

ecourreges:
fix infinite wait loop in threaded mode in case of any requests error
not using threads when only one ticker even if threads=True

Fixes the hanging parts of #607 and #598

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -73,6 +73,19 @@ class TickerBase():
             "yearly": utils.empty_df(),
             "quarterly": utils.empty_df()}
 
+    def __handle_error(self, shared, err_msg, **kwargs):
+        debug_mode = True
+        if "debug" in kwargs and isinstance(kwargs["debug"], bool):
+            debug_mode = kwargs["debug"]
+
+        shared._DFS[self.ticker] = utils.empty_df()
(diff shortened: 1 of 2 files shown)


ecourreges:
Maybe the complexity limit should be raised until the methods are refactored, the CodeFactor of my PR and others are not failing because of our contribution

yfinance/base.py
```diff
@@ -73,6 +73,19 @@ def __init__(self, ticker):
...
+    def __handle_error(self, shared, err_msg, **kwargs):
+        debug_mode = True
+        if "debug" in kwargs and isinstance(kwargs["debug"], bool):
+            debug_mode = kwargs["debug"]
+
+        shared._DFS[self.ticker] = utils.empty_df()
+        shared._ERRORS[self.ticker] = err_msg
+        
+        if "many" not in kwargs and debug_mode:
+            print('- %s: %s' % (self.ticker, err_msg))
```
kucharskim, line 85: Shouldn't that go to stderr?

kmcentush:
Was there a reason this was closed rather than being merged? I'm still seeing this infinite while loop issue on version 0.1.54.

Modifying my install directly with these above changes solved the issue for me.

PR #620, opened 2021-03-10
Added an important financialCurrency property

asafravid:
Without it, companies whos Earnings and Revenues are in non-USD currency, cannot be used to manualy calcualate (when EV/R ratio is unavailable) EV/R


asafravid (commented):
Return 'USD' when 'financialCurrency' is not available

PR #643, opened 2021-03-24
2 important updates

asafravid:
1. Conversion rate help using 'financialCurrency' property in earnings
2. Add important try+catch statements
3. sync-up with ranaroussi/yfinance v0.1.56


asafravid:
Synced with ranaroussi/yfinance v0.1.56

+ resolved

PR #648, opened 2021-03-26
Fix issue 570

ppeloton:
Hi,
issue is that currently start parameter for request is set to 1.1.1900 (actually to the corresponding UNIX timestamp -2208994789). My fix is to use the timestamp for 1.1.1950 (-631159200), so the 100 year limit set by yahoo finance is not exceeded. Any other date not more than 100 years prior to current date would be enough (e.g. 1.1.1940 or 1.1.1930), but I think the chosen date is fine.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -115,7 +115,7 @@ class TickerBase():
 
         if start or period is None or period.lower() == "max":
             if start is None:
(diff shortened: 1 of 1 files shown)


silvavn:
Good catch. LGTM

PR #649, opened 2021-03-27
wrote some test cases for balance sheet

jusongb:


jusongb:
that was an accident, lmao (i was supposed to do it inside a forked repo)

PR #652, opened 2021-03-27
Financials positive

elake:
Requires dill: pip install dill
Test logic is in the docstring for the financials property.
The test is run with the negative test in test_financials.py, or it can be run separately: python -m yfinance.ticker -v


elake:
Whoops, accidentally made this PR for the wrong repo. Please ignore!

PR #664, opened 2021-03-29
Testing for analyst recommendation function and fix #660

### LINKED ISSUE #660: Analyst recommendations function crashes only when input does not contain required key
We tried to find the very specific part of the exception that's causing problems in the `analyst_recommendations` function by developing unit tests and doc tests. All pandas functions worked properly. It seems that the function will only throw an exception if the data passed to the function does not contain the required keys or the data type is not "dict". If this is the only issue, it might be possible to develop a better solution than using `except Exception: pass`, such as checking the type of data and testing if the data contains the required keys before executing the function.

The PR #309 tried to solve this problem by adding an if condition: ` if data ["upgradeDowngradeHistory"] ["history”] `

xyqshelly:
This PR isolates the analyst recommendation function and creates unit tests and doc tests for the function.
Including the following test cases:
1. Test the case where correctly formatted data is passed into analyst_recommendations(). Test if the titles of the output of analyst_recommendations(data) are correctly converted from camel case to title case.
2. Test the case where correctly formatted data is passed into analyst_recommendations(). Test if the returned value is in the DataFrame format.
3. Test the case where correctly formatted data is passed into analyst_recommendations(). When the var output is not None, the recommendation's index name should be updated to "Date".
4. Test the case where correctly formatted data is passed into analyst_recommendations(). Test the case the order of the data returned in ascending date order, oldest to newest date at the end.
5. Test the case where correctly formatted data is passed into analyst_recommendations(). Test the case if the index values for the data are in DateTime format.
6. Test if an exception will be raised if incorrect data is passed into analyst_recommendations(). This test was marked as an "expected failure" because if an exception is thrown, it means that the function will crash when an invalid variable is passed. The original code works with no issue because it uses except Exception: pass to hide the bugs.

Bug found during the testing has been posted in issue #660. This PR also implements a fix for the issue. (checking the type of data and testing if the data contains the required keys before executing the function).
Documents for the test cases could be found in this forked repo: https://github.com/mxiuwang/yfinance/tree/master/docs

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -359,18 +359,7 @@ class TickerBase():
             pass
 
         # analyst recommendations
-        try:
-            rec = _pd.DataFrame(
-                data['upgradeDowngradeHistory']['history'])
-            rec['earningsDate'] = _pd.to_datetime(
-                rec['epochGradeDate'], unit='s')
-            rec.set_index('earningsDate', inplace=True)
-            rec.index.name = 'Date'
-            rec.columns = utils.camel2title(rec.columns)
-            self._recommendations = rec[[
-                'Firm', 'To Grade', 'From Grade', 'Action']].sort_index()
-        except Exception:
-            pass
+        self.analyst_recommendations(data)
 
         # get fundamentals
         data = utils.get_json(ticker_url+'/financials', proxy)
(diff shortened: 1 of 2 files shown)


yfinance/base.py
```diff
@@ -405,6 +394,27 @@ def cleanup(data):
 
         self._fundamentals = True
 
+    def analyst_recommendations(self, data):
+        # test if the type of data is dict
+        if isinstance(data, dict):
```
silvavn, line 399: This should use `assert`

xyqshelly: Isn't assert for testing? We used assert in our testing. The change in base.py was our software fix.

PR #685, opened 2021-04-12
Fix threads implementation

jrdi:
I invested some time fixing the threads implementation and also adding other improvements such as:

* Retries with backoff
* Normalize output for single tickers
* Remove the shared dfs, prune to race conditions

I'd appreciate it if you can provide some feedback.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ setup(
     packages=find_packages(exclude=['contrib', 'docs', 'tests', 'examples']),
     install_requires=['pandas>=0.24', 'numpy>=1.15',
                       'requests>=2.20', 'multitasking>=0.0.7',
-                      'lxml>=4.5.1'],
+                      'lxml>=4.5.1', 'futures; python_version == "2.7"'],
     entry_points={
         'console_scripts': [
             'sample=sample:main',
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -112,6 +112,7 @@ class TickerBase():
                     Optional. If passed as False, will suppress
                     error message printing to console.
         """
+        df = utils.empty_df()
 
         if start or period is None or period.lower() == "max":
             if start is None:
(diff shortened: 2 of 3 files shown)


arnaudgelas:
@jrdi Although I agree with your new implementation, it would be great if you could provide (at least) 1 example which was failing without your fix and now works with it.

ValueRaider:
@fredrik-corneliusson I believe you are a heavy user of `download()` - is this PR worth resolving & merging in?

fredrik-corneliusson:
Yes I think the threads implementation would be in need of a improvement, any exceptions raised in threads results in the whole download hanging.

starboi-63:
> @jrdi Although I agree with your new implementation, it would be great if you could provide (at least) 1 example which was failing without your fix and now works with it.

I'm not sure if this issue is local to the machine I was working with (M1 Macbook Pro, macOS 13.5 Ventura, Python 3.11.4), but for large downloads (7000+ tickers over 2 years), I get `RuntimeError: can't start new thread`. This doesn't stop the download execution immediately, but it consistently hangs mid way through. The specific list of tickers I used to cause the issue is from [NASDAQ's API](https://api.nasdaq.com/api/screener/stocks?tableonly=true&limit=25&offset=0&download=true).

I fixed the issue by implementing a [yf_download_batches](https://github.com/starboi-63/growth-stock-screener/blob/main/growth_stock_screener/screen/iterations/utils/scraping.py) function that splits large `download()` calls into smaller ones and combines the data using `pandas.concat()`. My function only saves adjusted close data, but I think the changes made by @jrdi would also fix the issue I was having by handling the RuntimeError.

ValueRaider:
Anyone can submit a pull request, just saying.

> but for large downloads (7000+ tickers over 2 years) @starboi-63 

Have you considered [yfinance-cache](https://github.com/ValueRaider/yfinance-cache) ?

Issue #678, opened 2021-04-06
Missing info using yFinance library


claudinho85:
Hi guys, just a question for you

I was using yfinance library for python and I realized that some information is missing compared to Yahoo Web Site.

For example, I was wondering why "Statistics" and "Analysis" sections under https://finance.yahoo.com/ (see attached screen) are missing inside yFinance python lib.

Please, help me! :-)

YFinance is the best one, I hope these sections will be integrated in the next future.

Otherwise, I can contribute to do that 

Thanks a lot,
Claudio

![Cattura2](https://user-images.githubusercontent.com/6480148/113710341-d928bb80-96e3-11eb-8eee-3912ab3e8812.PNG)
![Cattura1](https://user-images.githubusercontent.com/6480148/113710342-da59e880-96e3-11eb-9883-a324bd1c4bd8.PNG)

jemshit:
#701

PR #701, opened 2021-05-04
Add "financialData" to "info"

jemshit:
- Added `data.financialData` dictionary into `self._info` dictionary. It contains important data such as `totalDebt`, `quickRatio`, `debtToEquity`, `currentPrice`, `currentRatio` from Statistics page of stock
- Fixed `data.summaryDetail` being added twice into `self._info`
- Update `.gitignore` using [gitignore.io](gitignore.io). Contains rules for Python, PyCharm, VirtualEnv, MacOs, Windows, Linux. Otherwise all files of virtualenv is not ignored

Order of `defaultKeyStatistics`, `summaryDetail` matters, added comment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,322 @@ yfinance.egg-info
 .coverage
 .vscode/
 build/
+.idea/
+
+
+# Created by https://www.toptal.com/developers/gitignore/api/python,pycharm+iml,virtualenv,macos,windows,linux
+# Edit at https://www.toptal.com/developers/gitignore?templates=python,pycharm+iml,virtualenv,macos,windows,linux
+
+### Linux ###
+*~
+
+# temporary files which can be created if a process still has a handle open of a deleted file
+.fuse_hidden*
+
+# KDE directory preferences
+.directory
+
+# Linux trash folder which might appear on any partition or disk
+.Trash-*
+
+# .nfs files are created when an open file is removed but is still being accessed
+.nfs*
+
+### macOS ###
+# General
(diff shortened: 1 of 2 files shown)


claudinho85:
Hi jemshit, is it in roadmap a kind of integration of all data about Statistics and Analisys as asked before? 
Is it feasable to retrieve that information from web scraping or not? 

I've noticed a weird behaviour on that page ...

Thanks a lot

jemshit:
@claudinho85 Statistics are inside "info". I am not sure about Analysis

jemshit:
Web scarping is done on (example) https://finance.yahoo.com/quote/AAPL and https://finance.yahoo.com/quote/AAPL/financials. When you right-click on browser page (firefox) and click "view page source", you see the whole data. If the data you want is there, then it is feasible to retreive

claudinho85:
Unfortunately, info found on those pages are not present as written here https://github.com/ranaroussi/yfinance/issues/678 
but I could be wrong ... Actually I hope to be wrong :-) 

What about it?

chinmayandhare:
Hi jemshit, is it possible to add shares outstanding data to income statement when we pull it so that we get it for currennt as well as previous years?

jemshit:
> Hi jemshit, is it possible to add shares outstanding data to income statement when we pull it so that we get it for currennt as well as previous years?

I don't think so

ranaroussi:
I've added the `financialData` item without merging the pull requests since I didn't want to pull in your `.gitignore` file :)

Issue #147, opened 2019-12-13
feature request : add 'silent' option


ksjk1998:
Lets say I download 'AAAA'. this would output:

1 Failed download:
- AAAA: No data found, symbol may be delisted

Is it possible to ensure this print statement doesn't output anything?

ksjk1998:
yeah, instead of being the guy to suggest the change i decided to BE the change lol, here's my repo link if you want a free feature added.

https://github.com/ksjk1998/yfinance.git

pranjal-joshi:
> Lets say I download 'AAAA'. this would output:
> 
> 1 Failed download:
> 
> * AAAA: No data found, symbol may be delisted
> 
> Is it possible to ensure this print statement doesn't output anything?

If this has been successfully implemented, Can you make a Pull Request to this repo so we all can get the same feature that we want.

imLightSpeed:
You can already do this when passing `debug=False` to methods ie history.

>import yfinance as yf
> #Delisted ACR-C
> t = yf.Ticker("ACR-C")
> print(t.history(debug=False))

This will not print that the ticker is delisted

pranjal-joshi:
> You can already do this when passing `debug=False` to methods ie history.
> 
> >import yfinance as yf
> > #Delisted ACR-C
> > t = yf.Ticker("ACR-C")
> > print(t.history(debug=False))
> 
> This will not print that the ticker is delisted
> 

`download` method internally prints error.
Provided `Ticker` method example isn't helpful in this case

pranjal-joshi:
This can be closed as the feature implemented in #717

PR #717, opened 2021-05-21
[New Feature] Argument added for silencing error printing

pranjal-joshi:
`show_errors` argument added to `download` method with default value `True`.
`if show_errors == False` then errors **will not be printed.**

This is a fix for **Symbol Delisted** error printed if the data for that symbol is unavailable on Yahoo! Finance and discussed in #147


pranjal-joshi:
Thanks for merging! 😄

shaunadam:
I don't seem to be able to use this in 0.2.51. Was it later removed?

ValueRaider:
Read the docs

PR #743, opened 2021-06-28
self._history() graceful handling in case it fails

asafravid:
get_dividends(), get_splits() and get_actions() can still fail otherwise as they did not check for `if self._history is not None` and `"<action>" in self._history`


asafravid:
Add headers to session

PR #761, opened 2021-07-09
return 403 if missing headers

elminson:


ranaroussi:
Already implemented in 0.1.61

Issue #768, opened 2021-07-10
Better request in base.py


gilsapir86:
Hi,
I can't open a PR, so this is the change I did to make it much more stable localy:
in file:
yfinance/base.py
line 165
r = urllib.request.urlopen(data.url)
data = json.load(r)

much better than :
data = data.json()

ranaroussi:
There's no `urllib.request.urlopen(data.url)` anywhere in the code...

Issue #767, opened 2021-07-10
Unable to get any stock price data - JSONDecodeError: Expecting value: line 1 column 1 (char 0)


Aditya-Bahl:
import yfinance as yf

apple = yf.Ticker("AMZN")

hist = apple.history(period="max", start="2021-05-20", end="2021-06-04")

Aditya-Bahl:
It was working fine, but now just gives the JSON Decode Error.

315jordan09:
Looks like I am getting the same error.
Seems like a server may be down? Not really sure.

This is the line I am running:
return yf.download(tickers = ticker, period = "5d", group_by = 'ticker', threads = False)

This is the error I am getting:
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

315jordan09:
Solution:
pip3 install yfinance --upgrade --no-cache-dir
or use pip if you are not using pip3

Aditya-Bahl:
It worked for me in PyCharm, but is still not working in my Conda environment even after running those pip commands.

jim-schwoebel:
Yup same issue!

jim-schwoebel:
Thank you also for building this scraper. It is awesome.

jim-schwoebel:
pip3 install yfinance --upgrade --no-cache-dir solved it for me too!

stephenhky:
yeah, just works for me by simply upgrading it.

SimonHuddart:
I'm having the same issue and upgrading the package did not appear to fix it.

juanpacajigas:
The upgrade did it for me. Thanks!

PR #771, opened 2021-07-10
Better request in base.py

jla524:
Per https://github.com/ranaroussi/yfinance/issues/768 (thanks to @gilsapir86!). This should fix the error in https://github.com/ranaroussi/yfinance/issues/767.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -26,6 +26,8 @@ import datetime as _datetime
 import requests as _requests
 import pandas as _pd
 import numpy as _np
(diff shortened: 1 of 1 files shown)


jla524:
I found another way to do this using requests, but it requires me to specify a User-Agent
```
r = requests.get(data.url, headers={'User-Agent': None})  
data = r.json()
```

ranaroussi:
This doesn't fix anything. yfinance is already using the requests library

PR #784, opened 2021-07-16
Be safe in reading json

rtadros125:
When bad requests happen in using `yf.download`, either because an element in the list is wrongly formatted, or because of shaky internet connection, one bad ticker causes this line to blow up, and the `download` gets stuck. This small fix works around any type of bad requests leading to mis-formatted returns.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -159,12 +159,21 @@ class TickerBase():
             raise RuntimeError("*** YAHOO! FINANCE IS CURRENTLY DOWN! ***\n"
                                "Our engineers are working quickly to resolve "
                                "the issue. Thank you for your patience.")
-        data = data.json()
 
         # Work with errors
         debug_mode = True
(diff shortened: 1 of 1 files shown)


ranaroussi:
Hi @rtadros125 - can you explain the use of `debug_mode` and `kwargs["many"]` please?

rtadros125:
> Hi @rtadros125 - can you explain the use of `debug_mode` and `kwargs["many"]` please?

Hi @ranaroussi, I just copied the same structure as used in the following parts of the code for consistency. I did not investigate what they are. Look [here](https://github.com/ranaroussi/yfinance/blob/main/yfinance/base.py#L170) and [here](https://github.com/ranaroussi/yfinance/blob/main/yfinance/base.py#L192).

ValueRaider:
PR #907 fixes this problem with same solution.

PR #786, opened 2021-07-19
add pre-market price to info

neon-paradise:
I suppose if I needed this field, then others may also need it


patrickxchong:
Yes please! I would love to have this, been trying to figure out how I can hack in premarket data.

PR #857, opened 2021-10-21
try + except still required - topHoldings, summaryDetail

asafravid:
Reason is that a failure would be harmful without a graceful try+except

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -386,13 +386,19 @@ class TickerBase():
         except Exception:
             pass
 
-        # For ETFs, provide this valuable data: the top holdings of the ETF
-        if data.get('topHoldings'):
-            self._info.update(data['topHoldings'])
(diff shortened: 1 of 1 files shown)


asafravid:
Fix must be merged into master as-is (it fixes as required in my repositories which use yfinance extensively).
This is how `yfinance` works: 
The developer using `yfinance` is required and responsible to do the checks of whether a certain parameter exists in `info` as required after he got the info data and this is EXACTLY what I do in both of my repositories which use yfinance and require this graceful handling: https://github.com/asafravid/sss as well as https://github.com/asafravid/EtfCorrelationScanner [the latter was the origin for the important Fix]
The fix is already submitted in my clone of yfinance: https://github.com/asafravid/yfinance

asafravid:
@ranaroussi can you merge this pull request?
It's important for ETF scanning support and in general better-copes with missing data, etc
Very similar to previous similar pull requestes I made which were merged, and just as important

PR #907, opened 2021-12-15
JSONDecodeError When Data is None 

tsherknus:
Resolves an issue in base.py during historic retrieval of data.

In base.py history(), session.get can return None. Add an additional condition on line 190 of base.py has been added, which will throw a RuntimeError before the json parse exception can be thrown.

https://imgur.com/a/IgTqKah


tsherknus:
Anything I can do to help move this issue along?

ranaroussi (approved):

PR #920, opened 2021-12-31
Resolves UnboundLocalError in base.py history()

tsherknus:
In base.py history(), session.get can return None or 404. Clarify exception catching within the history() method of base.py.

Ensure none and 404 values are caught prior to quote processing.

This prevents json parse exception, connection exception, or unbound exception from being thrown during data collection.

https://imgur.com/a/iXfZsvb

Includes updated unit testing.

PR #921, opened 2021-12-31
data = None following exception w/ checks for this

isosphere:
Prior to these modifications, data could be unset due to an exception. 

Following that exception, we'd try to find "chart" in a variable that has never been declared. These modifications prevent that while hopefully keeping the intended functionality.


isosphere:
Closing in favour of the earlier submitted #920.

PR #941, opened 2022-01-25
Explicitly set datetime index on actions table

### LINKED ISSUE #937: Ticker.history: history's index is not of type DateTimeIndex when dividend and split tables are empty
**Issue**:

Ticker.history (line 295):
>       df.index = df.index.tz_localize("UTC").tz_convert(data["chart"]["result"][0]["meta"]["exchangeTimezoneName"])
E       AttributeError: 'Index' object has no attribute 'tz_localize'

**Reason**: 

During the concatenation of a proper price table with empty dividend and split tables, the price table's index has changed from _pandas.DatetimeIndex_ to _pandas.Index_ because the empty tables had this index type.

**Proposed solution**: 

Modify _parse_actions_ (in utils.py) to initialise the dividend and split tables with : 
`dividends = _pd.DataFrame(columns=["Dividends"], index=_pd.DatetimeIndex([]))`
 `splits = _pd.DataFrame(columns=["Stock Splits"], index=_pd.DatetimeIndex([]))`

Local modules' versions:

finance : 0.1.69
pandas : 1.4.0
python : 3.10

moshfrid:
Explicitly set datetime index on actions df so that if the actions df is empty, it doesn't reset the index when combined with the rest of the dataframe. Closes #937. Credit to @ProgrammingDao

### DIFF
--- a/yfinance/utils.py
+++ b/yfinance/utils.py
@@ -197,8 +197,8 @@ def parse_quotes(data, tz=None):
 
 
 def parse_actions(data, tz=None):
-    dividends = _pd.DataFrame(columns=["Dividends"])
-    splits = _pd.DataFrame(columns=["Stock Splits"])
+    dividends = _pd.DataFrame(columns=["Dividends"], index=_pd.DatetimeIndex([]))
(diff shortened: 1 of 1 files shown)


ProgrammingDao:
Thank you @moshfrid, changes look good to me.

Issue #949, opened 2022-01-31
Compatibility with windows 10


angelo-daumas:
In Windows 10, time.mktime does not work for negative values. Therefore, lines 146 through 159 of [this file](https://github.com/ranaroussi/yfinance/blob/main/yfinance/base.py) will fail when using that OS if the start date is earlier than 1970.

Would it be possible to make this library compatible with windows 10 by changing the calls to time.mktime() to calls to the timestamp() method of the datetime class instead? This will work as long as the created datetime object is timezone-aware.

If a timezone aware datetime object is not ideal, this solution could be used instead: `(dt - datetime(1970, 1, 1)).total_seconds()`.

More info: https://stackoverflow.com/questions/59199985/why-is-datetimes-timestamp-method-returning-oserror-errno-22-invalid-a

If this change is okay, I could make a PR.

PR #950, opened 2022-01-31
Add compatibility with dates before 1970 on Windows

angelo-daumas:
See issue: https://github.com/ranaroussi/yfinance/issues/949

In order to avoid any issues, I simply created an alternative implementation of [`time.mktime`](https://docs.python.org/3/library/time.html#time.mktime). This implementation is then used where the original had been.

Of course, you could also make it so it's only used when running on Windows, or only when time.mktime raises an error. If this is deemed to be a better alternative, let me know so I can change the code.


angelo-daumas:
Closing as this has been fixed by #1048

Issue #1013, opened 2022-06-09
Get earnings date historics


nono-london:
Hi, 
I couldn't find a way to access earnings date history for a given ticker.
It would be great if could be added to the library.
Data seem to be easyly accessible:

`
web_request = requests.get(url='https://finance.yahoo.com/calendar/earnings?symbol=TD.TO',headers=headers)

web_tables = pd.read_html( web_request.text)

print(web_tables[0])
`
Thanks!

rickturner2001:
Hey @nono-london  there are many ways to get access to upcoming and previous earning dates. Check this out.

```
import yfinance as yf

ticker = yf.Ticker("MSFT")

# Upcoming earnings 
ticker.calendar.loc['Earnings Date']

# Previous earnings (quarterly)
ticker.get_earnings(freq="quarterly")

# Previous earnings (yearly)
ticker.get_earnings(freq="yearly")
```

I hope this helps!

nono-london:
Rick Tx for your message. When I used quaterly, i seem to be receiving only the last year, the Yahoo dataset is much bigger in term of quarters history:
yfinance:
![image](https://user-images.githubusercontent.com/7720585/173204455-7c310767-c2dc-4395-be38-4b09c36aac66.png)
yahoo using requests and url above:
![image](https://user-images.githubusercontent.com/7720585/173204516-d6449e9f-b818-4e78-b2e3-3ab714cfe3ad.png)

I have created my own function, but seems like it could be easily added to the package.

rickturner2001:
Yes, I totally agree with you. I sent a Pull request yesterday, let's see if they merge it. Nice catch!

https://github.com/ranaroussi/yfinance/pull/1014

nono-london:
tx a lot

ValueRaider:
You should add a class variable `TickerBase._earnings_history` for caching the table, as done with most other properties.

ValueRaider:
Also I think you should make the EPS columns float type (currently they are str), and the 'Earnings Date' column should be date type (I can help with timezones if you get stuck).

nono-london:
on my function i created a new function for date, as I wasn't sure if 'Earnings Date' was always coming in the same timezone -EST in my case but didn't test for german stocks for instance, but yeah agreed, would be great to have all in datetime UTC for instance. and relevant types for other columns

nono-london:
[...]

PR #1014, opened 2022-06-11
Added get_earnings_history to fetch earnings data

rickturner2001:
I've been using yfinance for a while and I noticed that it's missing methods to fetch substantial earnings data. The method sends a request to https://finance.yahoo.com/calendar/earnings?symbol={ticker} where {ticker} is self.ticker in the parent class. I used pandas to parse the HTML to pandas.DataFrame since the content in the HTML is a table.

### DIFF
--- a/test_yfinance.py
+++ b/test_yfinance.py
@@ -52,6 +52,8 @@ class TestTicker(unittest.TestCase):
             ticker.options
             ticker.news
             ticker.shares
+            ticker.earnings_history
+
 
     def test_holders(self):
         for ticker in tickers:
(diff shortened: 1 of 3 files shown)


ValueRaider:
I think this needs some polish before merging. See my comments here: https://github.com/ranaroussi/yfinance/issues/1013#issuecomment-1153007573

ValueRaider:
Hi Ran, are you sure you meant to merge this implementation rather than mine? This only fetches one page (so misses pre-1998), doesn't convert types nor set date as index.

For more details see discussion here: https://github.com/ranaroussi/yfinance/issues/1013#issuecomment-1153007573

Issue #1030, opened 2022-07-13
TypeError: an integer is required (got type str) for endDt


matrak:
For the `Ticker#history` function in the fix introduced by commit: https://github.com/ranaroussi/yfinance/commit/9d5366d707d3f45ab84d3e92dc4edcb9498dcb7f
the `end` parameter has to be an integer, but the conversion to integer happens only under certain conditions:
`start or period is None or period.lower() == "max"`

In the current version (0.1.74) the call: `ticker.history(period='2y', interval='1d', end='2022-07-13')` would cause the following exception:
```
endDt = _pd.to_datetime(_datetime.datetime.fromtimestamp(end))
typeError: an integer is required (got type str)
```

ValueRaider:
Setting aside the exception for now - how do you want the `end` argument to be used in this scenario? Currently it is ignored completely in Yahoo request, but you setting it implies you want it to be used. I ask because the correct fix might be to force user to leave start/end blank when setting period.

matrak:
Sorry if this is an edge case, but from what I see (in the returned data) the `end` parameter is processed in the function call.

For the fix I would suggest to handle the `end` parameter independently from other input variables.
I would gladly provide a merge request with a fix by the end of the week.

ValueRaider:
Correct that the `yfinance` function call uses `end` now, but this is a recent addition that post-processes the data after the Yahoo web request. For a long time before this commit, when `period` was set `yfinance` would completely ignore `end`.

This is a valid issue because `end` is ambiguous when `period` also set. I think it is nonsense to set both `period` and `start/end`.

ValueRaider:
That PR is what I think is best solution. What do you think?

Issue #115, opened 2019-10-18
Unable to obtain daily stock prices using yfinance 


RSB-Balaji:
`import pandas_datareader.data as pdr`
`import yfinance as yf`
`yf.pdr_override()`
`data = pdr.get_data_yahoo("AAPL",start ='2018-01-01',end='2018-12-31')`

The exception says:
Exception in thread Thread-1:
ConnectionResetError: [Errno 54] Connection reset by peer

During handling of the above exception, another exception occurred:
requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(54, 'Connection reset by peer'))

ranaroussi:
This seems like a connection issue. Closing for now. Please reopen issue if the problem persists.

PR #1031, opened 2022-07-13
Address mixing of period and start/end dates

ValueRaider:
This is a **proposed** fix for #1030 - if `period` is set, and user also set `start` or `end`, then ignore `start` and `end` and print a message to user.

~But maybe better fix is to use `start`/`end` to cap the received table. I honestly don't know.~ **Discussion welcome**


ValueRaider:
Issue #115 wants mixing of period and dates, so alternative solution in #1030 better.

Issue #1028, opened 2022-07-11
KeyError: Timestamp


isetthestandard:
Traceback (most recent call last):
  File "pandas\_libs\index.pyx", line 545, in pandas._libs.index.DatetimeEngine.get_loc
  File "pandas\_libs\hashtable_class_helper.pxi", line 2131, in pandas._libs.hashtable.Int64HashTable.get_item
  File "pandas\_libs\hashtable_class_helper.pxi", line 2140, in pandas._libs.hashtable.Int64HashTable.get_item
KeyError: 1657006800000000000

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:\anaconda3\envs\trading_bot\lib\site-packages\pandas\core\indexes\base.py", line 3621, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas\_libs\index.pyx", line 513, in pandas._libs.index.DatetimeEngine.get_loc
  File "pandas\_libs\index.pyx", line 547, in pandas._libs.index.DatetimeEngine.get_loc
KeyError: Timestamp('2022-07-05 09:40:00+0200', tz='Europe/Berlin')

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "D:\anaconda3\envs\trading_bot\lib\site-packages\pandas\core\indexes\datetimes.py", line 679, in get_loc
    return Index.get_loc(self, key, method, tolerance)
  File "D:\anaconda3\envs\trading_bot\lib\site-packages\pandas\core\indexes\base.py", line 3623, in get_loc
    raise KeyError(key) from err
KeyError: Timestamp('2022-07-05 09:40:00+0200', tz='Europe/Berlin')

The above exception was the direct cause of the following exception:

KeyError: Timestamp('2022-07-05 09:40:00+0200', tz='Europe/Berlin')

i Got this error message by some German stock with an interval of 1 minute.
Are there any possibilities to avoid this and get the correct data ?

best regards

ValueRaider:
Post code that reproduces error.

isetthestandard:
this is a short excerpt of the code. the first passes of the loop work, then suddenly this error message appears.

`import yfinance as yf
import pandas as pd
from datetime import date
import datetime

ticker = yf.Ticker('LHA.DE')
start = date.today() - pd.Timedelta(days=7)
end = date.today()
df = ticker.history(interval='1m', start=start, end=end)
start_time = datetime.time(9,00,00)
for i in range(len(df)):
[...]

PR #1032, opened 2022-07-14
Add 'keepna' argument

ValueRaider:
Keeping NaNs rows returned by Yahoo may be useful because the timepoints are still valid so provides list of trading intervals.

New argument `keepna=False` controls whether to keep/drop these NaN rows.

#1028 relevant

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -106,7 +106,7 @@ class TickerBase():
 
     def history(self, period="1mo", interval="1d",
                 start=None, end=None, prepost=False, actions=True,
-                auto_adjust=True, back_adjust=False,
+                auto_adjust=True, back_adjust=False, keepna=False,
                 proxy=None, rounding=False, tz=None, timeout=None, **kwargs):
         """
         :Parameters:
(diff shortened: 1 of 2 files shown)


ValueRaider:
This is breaking 15m, let me fix ...

ValueRaider:
False alarm. 15m and 30m prices looking good.

ProgrammingDao:
Might be good to include the following line in the impacted drop calls: https://github.com/ranaroussi/yfinance/blob/ec6279736b570815ec017691b428c90d910b1739/yfinance/base.py#L320

ValueRaider:
> Might be good to include the following line in the impacted drop calls:
> 
> https://github.com/ranaroussi/yfinance/blob/ec6279736b570815ec017691b428c90d910b1739/yfinance/base.py#L320

It should stay. Because `how='all'` is specified, then only if all columns are NaN are rows dropped. But ["Volume","Dividends","Stocks Splits"] should have their NaNs replaced with 0, so only with a yfinance bug would rows be dropped. We want to keep Yahoo NaNs, not buggy yfinance NaNs.

Issue #973, opened 2022-03-17
the latest date appears NaN


anarchy89:
so I ran this command, 

yf.Ticker('ALB').history(period='1d')

its supposed to look like this 

Mar 17, 2022 | 187.84 | 196.09 | 187.84 | 195.21 | 195.21 | 403,080
-- | -- | -- | -- | -- | -- | --
Mar 17, 2022 | 0.395 Dividend
Mar 16, 2022 | 185.87 | 193.98 | 185.76 | 191.25 | 190.85 | 890,700
Mar 15, 2022 | 175.36 | 182.70 | 172.78 | 181.95 | 181.57 | 834,400

but what I got was this,

	Open	High	Low	Close	Volume	Dividends	Stock Splits
Date							
2022-03-17	NaN	NaN	NaN	NaN	NaN	0.395	0

it only showed me the latest empty dividend data.

how can I fix this?

anarchy89:
even if I run this

yf.Ticker('ALB').history(),

the latest date shows me only the blank dividend data when it should be showing me 



Date | Open | High | Low | Close* | Adj Close** | Volume
-- | -- | -- | -- | -- | -- | --
Mar 17, 2022 | 187.84 | 196.09 | 187.84 | 195.21 | 195.21 | 403,080


according to the yahoo finance data

anarchy89:
why does it not show the actual OHLC data? 

this is as far as it goes



  | Open | High | Low | Close | Volume | Dividends | Stock Splits
-- | -- | -- | -- | -- | -- | -- | --

218.038740 | 220.543551 | 194.427604 | 196.613083 | 6180600.0 | 0.000 | 0
199.606897 | 200.994026 | 186.623758 | 190.366013 | 3060900.0 | 0.000 | 0
188.709446 | 199.217696 | 188.210478 | 191.453766 | 1635500.0 | 0.000 | 0
192.601380 | 195.714932 | 180.446544 | 181.624100 | 1644000.0 | 0.000 | 0
171.904222 | 184.578005 | 169.579033 | 183.819565 | 1979800.0 | 0.000 | 0
184.019148 | 191.603458 | 179.807863 | 191.553558 | 1135300.0 | 0.000 | 0
192.202214 | 200.465112 | 190.136482 | 195.485413 | 1583600.0 | 0.000 | 0
194.078319 | 196.533245 | 183.859479 | 186.054932 | 992000.0 | 0.000 | 0
[...]

Issue #134, opened 2019-11-20
History does have NaN values


slapslash:
when downloading the history of a ticker it does have rows with nan values in it.
as there is a quotes.dropna(inplace=True) in base.py I do assume, this is not intended.
I narrowed it to at least being dates, where splits or dividends happened.
By the way, those rows do not always show up in yahoo's data :/
```python
stock = yf.Ticker('RAW.DE')
history = stock.history(period='max', auto_adjust=True, back_adjust=True, rounding=True, actions=False)
```
```text
             Open   High    Low  Close   Volume
Date                                           
2012-06-27    NaN    NaN    NaN    NaN      NaN
2013-07-03    NaN    NaN    NaN    NaN      NaN
2014-06-11    NaN    NaN    NaN    NaN      NaN
2018-04-03  29.36  29.36  28.62  28.90   9553.0
2018-04-04  28.74  28.78  28.01  28.27  11568.0
...           ...    ...    ...    ...      ...
2019-11-14  22.75  22.75  22.09  22.09   6361.0
2019-11-15  21.76  21.80  21.65  21.80   1670.0
2019-11-18  21.80  21.80  21.25  21.43   3765.0
2019-11-19  21.80  21.85  21.68  21.85    107.0
2019-11-20  21.58  21.68  21.58  21.68    100.0
```

silvavn:
This is no longer the case.

abishekh:
Can confirm this is still happening.

silvavn:
@abishekh Please provide reproduction code

masayang:
`stock = yf.Ticker('SPY')`
`history = stock.history(period='max', auto_adjust=True, back_adjust=True, rounding=True, actions=False, interval="1wk")`
`print(history.to_string())`

`2022-03-14  417.78  441.57  412.72  444.52  5.559255e+08`
`2022-03-18     NaN     NaN     NaN     NaN           NaN`
`2022-03-21  442.43  451.03  438.78  452.69  3.842645e+08`
`2022-03-28  450.12  460.08  447.21  452.92  4.455269e+08`
`2022-04-04  451.18  455.86  441.56  447.57  3.980834e+08`
`2022-04-11  442.20  443.83  434.77  437.79  3.460740e+08`
`2022-04-18  434.93  448.07  423.61  426.04  4.269370e+08`
`2022-04-25  421.85  427.79  409.44  412.00  5.966142e+08`
[...]

PR #1049, opened 2022-08-15
Fix dividends & splits

ValueRaider:
Dividends and splits should be _merged_ with price data not appended to bottom, produces a cleaner table (less NaNs). For weekly/monthly intervals, backdate events to start of week/month for a clean merge.

Merging verifies that no events were lost in process.

Fixes [#973](https://github.com/ranaroussi/yfinance/issues/973) [#134](https://github.com/ranaroussi/yfinance/issues/134)


ValueRaider:
Merge is messed up, will create new branch and PR

Issue #1062, opened 2022-09-17
yfinance info logo_url value problem


Hcevahir34:
Hi,

I try to get logos from yfinance but while ticker is 'ARCLK.IS',  logo_url value is empty.

I hope someone get update 'logo_url' value with 'https://logo.clearbit.com/arcelik.com.tr'

Thanks for helps,

code:

ticker = yf.Ticker('ARCLK.IS')
inf = ticker.info
print(inf)

output:

logo_url': ''

Jossan84:
Hi. Today I made a pull request, that I think fix this problem

ValueRaider:
Thanks Jossan, this will go out in next release.

PR #1064, opened 2022-09-19
Bugfix: Get logo url when no website exists

Jossan84:
Hi.
This is a very helpful project. I noticed that when no website exists on the info, the logo url is not generated. I add a little change that in this case uses the small name that allways is present.
I think this could help with this Issue:
 
yfinance info logo_url value problem #1062

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -460,9 +460,12 @@ class TickerBase():
 
         self._info['logo_url'] = ""
         try:
-            domain = self._info['website'].split(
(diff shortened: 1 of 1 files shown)


ValueRaider:
Hi Jossan. Thanks for contributing a PR, but as you may have noticed the PRs are not being merged in. Only one person can merge and they have effectively abandoned this.

So I'm trying to set up a replacement fork with a team of maintainers, are you interested in volunteering?

Issue #1044, opened 2022-07-31
Wrong values for Ebit and missing values for Extraordinary items


NicsunXnus:
1) I have noticed that the values for Ebit are the same as Operating income for some stocks. I have tried this with 2 stocks "GOOG" and "TSM" and they both showed Ebit values that are the same as Operating income. However this is not the case for "D05.SI" and instead the Operating Income is the same as the Income before Tax. I was unable to verify for this stock since much of the data is not found directly on yahoo finance so I assume that they must have been calculated within the code.

2) Extraordinary items are showed as none for some stocks, even when they are displayed on the yahoo finance page. Example stocks used are "D05.SI","GOOG" & "TSM". I am thinking this is because of naming issues, as on the yahoo finance page, it is listed as unusual items.

Thank you

ValueRaider:
Did you search through existing issues before submitting? Your issue is familiar.

NicsunXnus:
Maybe I was a bit too hasty when I posted :P. I just searched through the issues and under #1004, found that there may be a possible fix from git-shogg's code (for the ebit and operating income identical values issue)? I notice that there is a pending merge request of his code into this codebase so hopefully that gets settled soon!

Regarding the missing extraordinary items issue, I am certain that it is a new issue as I was not able to find any similar issues under the open and closed issues sections.

ValueRaider:
They're all the same issue really - yfinance returning different financials than shown on website. The pending PR should fix all of these, it just needs more work for quarterlys.

Issue #1079, opened 2022-10-08
EBIT wrongly returns Opreating Income for (quaterly) financials


aliencaocao:
```
import yfinance as yf
msft = yf.Ticker('msft')
msft.financials.iloc[:, 0]
```
Under 'Ebit', the value is wrong. In fact, it seems to be returning Opreating Income instead. Same issue for both quarterly and yearly.

Running latest 0.1.77, same issue exist in 0.1.75 which means is a long-term bug. Tested with INTC as well.

https://finance.yahoo.com/quote/MSFT/financials?p=MSFT

ValueRaider:
Have you searched through existing issues? It likely already exists

aliencaocao:
Yes, found 7 open issues, none had a proper solution, its all workarounds. I already have a workaround, but I would still like to report this because this is clearly a bug with the library.

ValueRaider:
We haven't forgotten about those 7 issues, so please don't duplicate. Just comment on those threads.

Issue #1004, opened 2022-05-19
Financial values (Income Statement) are differnt between website and yfinance


prischon:
Values for example EBIT or Gross Profit are always diferent. I cheked differend companies (IBM, Google, Deutsche Wohnen and so on). The values are always different between webside and yfinance

a part of the code:
IBM = yf.Ticker("IBM")
print(IBM.financials)

then compare the values between the site (see [link](https://finance.yahoo.com/quote/IBM/financials?p=IBM)) and output values from python instruction.

ValueRaider:
I did a little digging and the fault lies with Yahoo not yfinance. 

Check for yourself - the data returned by API is present in HTML source. Right click -> view page source, and search for e.g. "grossProfit", you'll see it's different to what is being displayed in table.

prischon:
Meanwhile I checked beautifulsoup over „[yahoofinancials](https://pypi.org/project/yahoofinancials/)“. The library “[yahoofinancials](https://pypi.org/project/yahoofinancials/)” is old and does not work, but the call “beautifulsoup” works. Over the call beautifulsoup I got the correct data for EBIT (the name of value was “anualEBIT”). Also other values were identical with values on the site https://finance.yahoo.com/. The beautifulsoup scraped the data from the site https://finance.yahoo.com/.

From my point of view the API “yfinance” works not correctly.

prischon:
I checked now also the HTML code. Here are present the volumenas and names (see name "annualEBIT")

![grafik](https://user-images.githubusercontent.com/103684905/172477700-d473785a-fea3-4c1f-b5d9-9c7b959b9bfe.png)

ValueRaider:
I am confused by your statements. For me, yahoofinancials is returning the same value as yfinance:

```
from yahoofinancials import YahooFinancials as YF
tick = YF("IBM")
print(tick.get_ebit())
```

> 5772000000

Please post your BeautifulSoup code.

prischon:
I think, I wasn't precise enough. 

What I done:
I put some break points during the execution and cheek the conten of variables. I did no change any code line inside yahoofinancials. 

Now I done the same with the code of yfinance and I saw the same problem. Please put a break point inside the file "utils.py"

![grafik](https://user-images.githubusercontent.com/103684905/176539166-55b17ac3-3d68-4b9e-88b1-dccf0be18202.png)

Explained flatly:
[...]

Issue #844, opened 2021-09-29
Wrong EBIT numbers with MSFT example


byagiz:
Hi guys,

I was playing around with the examples in the code description page and realised that:

> import yfinance as yf
> MSFT = yf.Ticker("MSFT")
> print(MSFT.financials)

shows wrong EBIT numbers for MSFT. I read from my terminal following output:
2021-06-30         2020-06-30         2019-06-30        2018-06-30
69916000000.0   53145000000.0   42959000000.0   35058000000.0

cross checking with yahoo's website shows me different numbers:
        TTM                 6/29/2021        6/29/2020       6/29/2019       6/29/2018
EBIT	73,448,000	73,448,000	55,627,000	46,374,000	39,207,000

Is there a reason why this is happening? If you have any questions please contact me!

asafravid:
Seems you’re trying to compare yearly values to quarterly values… please recheck

Asaf - אסף

walterlsb:
I'm getting a problem with ebit figures not matching as well.
This is my code and result:


```
import yfinance as yf
MSFT = yf.Ticker("MSFT")
fin = MSFT.get_financials(as_dict=True, freq='quarterly')
ebitTTM = 0
for key in fin.keys():
    print(key,fin[key]['Ebit'])
    ebitTTM = ebitTTM + fin[key]['Ebit']

print('TTM',ebitTTM)

```
Result:
```
2021-09-30 00:00:00 20238000000.0
2021-06-30 00:00:00 19095000000.0
2021-03-31 00:00:00 17048000000.0
2020-12-31 00:00:00 17897000000.0
TTM 74278000000.0
```

The website shows:
EBIT (x1000)

TTM: 77,798,000
9/30/2021: 21,063,000
6/30/2021: 19,958,000
3/31/2021: 17,869,000
12/31/2020: 18,908,000

I can see the numbers are similar - but shouldn't they be the same?

teeejer:
When I check for EBIT values, it looks like I'm actually getting the Operating Income values.
```
Ebit                                     69916000000.0   53145000000.0   42959000000.0   35058000000.0
Operating Income                         69916000000.0   53145000000.0   42959000000.0   35058000000.0
```
This is my workaround for what it's worth.
```
MSFT = yf.Ticker("MSFT")
fin = MSFT.get_financials(as_dict=True, freq='yearly')
ebit = 0
for key in fin.keys():
[...]

PR #1092, opened 2022-10-18
Fix wrong data for yearly and quarterly EBIT

aliencaocao:
This is a quick & dirty fix. It currently ignores custom headers and proxies that the user may set.

It uses the API directly to get data instead of the current HTML parsing way. This way can get much more data, but it is too much work to migrate to API for all the data, so I just fixed the EBIT field as many have reported the issue.

If you would like me to work on migrating to the new API, I am happy to help, but can only start working on it in December. This API way works for income statements, balance sheet and cashflow, both quarterly and yearly and TTM.

Hope it will get merged soon to help out numerous people facing the issue. I know there are similar fixes but they are all larger scaled so takes longer to test and merge.

Fixes https://github.com/ranaroussi/yfinance/issues/1044 https://github.com/ranaroussi/yfinance/issues/1004 (EBIT part of it but the rest can also be fixed by similar method) https://github.com/ranaroussi/yfinance/issues/844 https://github.com/ranaroussi/yfinance/issues/1079

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -43,6 +43,7 @@ import json as _json
 from . import shared
 
 _BASE_URL_ = 'https://query2.finance.yahoo.com'
+_FUNDAMENTALS_BASE_URL_ = 'https://query1.finance.yahoo.com/ws/fundamentals-timeseries/v1/finance/timeseries/'
 _SCRAPE_URL_ = 'https://finance.yahoo.com/quote'
 _ROOT_URL_ = 'https://finance.yahoo.com'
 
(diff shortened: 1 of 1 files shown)


ValueRaider:
We're currently working on a broader fix, although not using that url for everything. Currently in `dev` branch undergoing testing. If you try it and find issues put them here: https://github.com/ranaroussi/yfinance/discussions/1081

ValueRaider:
That's a question for @git-shogg. I think their reason is via-html enables reproduction of the nested structure, but via API just provides string-value.

aliencaocao:
Ok, that makes sense to me, although I personally do not think keeping the nested structure will be helpful to anyone, since, from a quantitive analysis usage standpoint, the user would just need the numerical data of the particular field, but does not really care where it belongs under, that is more for financial knowledge not for quantitative analysis.
Still, it introduces unnecessary complexity and slowdowns in code since regex and json parsing are both slow.

ValueRaider:
That was my thought too (complexity), but @git-shogg feels strongly that nested structure is useful for DCF: https://github.com/ranaroussi/yfinance/pull/776#issuecomment-1279914639

I'll create a thread in the discussion on this.

aliencaocao:
Alright. I shall close this PR then, since it is already implemented in dev branch. Thanks!

Issue #1107, opened 2022-10-23
Whatever you did to timezone / cache recently broke my AWS Lambda function


ryanlampkin:
I never had this problem before this week. Seems like you've made changes to utils [image below]. With respect to AWS lambda, I am not able to "make directory / store".

![Untitled](https://user-images.githubusercontent.com/39038038/197408883-87ba8851-a931-42ab-a933-09f20488999a.png)

ryanlampkin:
Last working version for me is 0.1.75

ValueRaider:
Sorry about that. I've sent out a new release - 0.1.81

This was because I recently added a local caching store for timezone data, because accessing info[] first time is relatively expensive. It assumed access to default local cache folder.

I've also added function `set_tz_cache_location()`, in case there is somewhere you can write.

ValueRaider:
This appears to be solved so will close.

PR #1108, opened 2022-10-23
Fix cache error on read only system

ValueRaider:
Silently handle errors in the timezone-cache without stopping execution.

Add `set_tz_cache_location()` so user can provide alternative location, in case default not accessible.

Resolves issue #1107

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,6 @@ Yahoo! finance API is intended for personal use only.**
 
 The `Ticker` module, which allows you to access ticker data in a more Pythonic way:
 
-Note: yahoo finance datetimes are received as UTC.
-
 ```python
 import yfinance as yf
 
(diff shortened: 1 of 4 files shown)


ValueRaider:
@fredrik-corneliusson - would you mind porting these changes in `base` and `utils` into _TzCache?

fredrik-corneliusson:
Sure, but will probably not have time to work on it today.

paulmcq (commented):
_yfinance/utils.py in get_cache_dirpath:_
``` 
    if _cache_dp is None:
        dp = _os.path.join(_ad.user_cache_dir(), "py-yfinance")
    else:
        dp = _os.path.join(_cache_dp, "py-yfinance")
```
_may be more readable as:_
```
    udc = _cache_dp or _ad.user_cache_dir()
    dp = _os.path.join(udc, "py-yfinance")
    return dp
```

ValueRaider:
@paulmcq Thanks for reviewing. But we are planning to merge in big changes from `dev` branch soon, so these particular review comments will be redundant.

Issue #1076, opened 2022-10-05
KeyError: 'exchangeTimezoneName' when accessing ticker history


manoelmarques:
Current yfinance 0.1.75 raises an exception when running  the following:
```
import yfinance as yf


def _main():
    stock_data = yf.download(
        "AFN",
        start="2018-01-01",
        end="2018-12-31",
        group_by="ticker",
        threads=False
    )
    print(stock_data)


if __name__ == "__main__":
    _main()
```
```
  stock_data = yf.download(
  File "/Users/manoel/opt/anaconda3/envs/Qiskitenv/lib/python3.8/site-packages/yfinance/multi.py", line 120, in download
    data = _download_one(ticker, period=period, interval=interval,
  File "/Users/manoel/opt/anaconda3/envs/Qiskitenv/lib/python3.8/site-packages/yfinance/multi.py", line 205, in _download_one
    return Ticker(ticker).history(period=period, interval=interval,
  File "/Users/manoel/opt/anaconda3/envs/Qiskitenv/lib/python3.8/site-packages/yfinance/base.py", line 148, in history
    end = utils._parse_user_dt(end, self._get_ticker_tz())
  File "/Users/manoel/opt/anaconda3/envs/Qiskitenv/lib/python3.8/site-packages/yfinance/base.py", line 328, in _get_ticker_tz
    tkr_tz = self.info["exchangeTimezoneName"]
KeyError: 'exchangeTimezoneName'
```

Running the same sample against the previous release returns:
```
[*********************100%***********************]  1 of 1 completed
                 Open       High        Low      Close  Adj Close  Volume
Date                                                                     
2018-01-02  29.799999  29.799999  29.799999  29.799999  29.799999       0
2018-01-03  30.000000  30.000000  30.000000  30.000000  30.000000       0
2018-01-04  30.000000  30.000000  30.000000  30.000000  30.000000       0
2018-01-05  30.200001  30.200001  30.200001  30.200001  30.200001       0
2018-01-08  30.000000  30.000000  30.000000  30.000000  30.000000       0
2018-01-09  30.200001  30.200001  30.200001  30.200001  30.200001       0
2018-01-10  32.799999  32.799999  32.799999  32.799999  32.799999       0
2018-01-11  31.799999  31.799999  31.799999  31.799999  31.799999       0
2018-01-12  30.799999  30.799999  30.799999  30.799999  30.799999       0
[...]

PR #1109, opened 2022-10-23
Fix for #1076

fredrik-corneliusson:
Make sure to only do 1 call to self.info and simplified the code but retaining the logic.


ValueRaider:
Good. Just formally mentioning issue - #1076

PR #1112, opened 2022-10-24
Improve performance of fetching Ticker timezone

ValueRaider:
Instead of fetching and building `info`, just fetch basic price data which contains timezone. ~1.7x speedup. Hopefully also reduces failures to get timezone (issue #1076)

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -558,7 +558,7 @@ class TickerBase():
 
     def _fetch_ticker_tz(self, proxy=None, timeout=None):
         # Query Yahoo for basic price data just to get returned timezone
-        
+
         params = {"range":"1wk", "interval":"1d"}
 
         # setup proxy in requests format
@@ -576,20 +576,6 @@ class TickerBase():
             if "Will be right back" in data.text or data is None:
                 return None
             data = data.json()
-        except Exception:
-            data = None
-
-        fail = False
-        if data is None or not type(data) is dict:
-            fail = True
-        elif type(data) is dict and 'status_code' in data.keys():
-            fail = True
-        elif not "chart" in data or data["chart"]["result"] is None or not data["chart"]["result"]:
-            fail = True
-        if fail:
-            return None
-
-        try:
             return data["chart"]["result"][0]["meta"]["exchangeTimezoneName"]
         except:
             return None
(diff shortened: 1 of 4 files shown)


ValueRaider:
@fredrik-corneliusson

yfinance/base.py
```diff
@@ -559,6 +556,28 @@ def _get_ticker_tz(self):
         self._tz = tz
         return tz
 
+    def _fetch_ticker_tz(self, proxy=None, timeout=None):
+        # Query Yahoo for basic price data just to get returned timezone
+
+        params = {"range":"1wk", "interval":"1d"}
```
fredrik-corneliusson, line 562: I think range can be "1d" as well to get even less data.

yfinance/base.py
```diff
@@ -559,6 +556,28 @@ def _get_ticker_tz(self):
...
+
+        # Getting data from json
+        url = "{}/v8/finance/chart/{}".format(self._base_url, self.ticker)
+
+        session = self.session or _requests
+        try:
+            data = session.get(url=url, params=params, proxies=proxy, headers=utils.user_agent_headers, timeout=timeout)
+            data = data.json()
+            return data["chart"]["result"][0]["meta"]["exchangeTimezoneName"]
+        except:
```
fredrik-corneliusson, line 578: Maybe if debug is True print the exception 
example:
```
except Exception as e:
    if debug_mode:
        print("Failed to get ticker '{}' reason: {}".format(self.ticker, e))
```

yfinance/base.py
```diff
@@ -533,7 +533,7 @@ def _fix_unit_mixups(self, df, interval, tz_exchange):
         return df
 
 
-    def _get_ticker_tz(self):
+    def _get_ticker_tz(self, proxy=None, timeout=None):
```
fredrik-corneliusson, line 536: Is _get_ticker_tz called without proxy and timeout in any other place? 
As proxy and timeout is passed to the history method and not Ticker constructor, any calls to this from another methods might not have the correct proxy. (However I do not fully understand the use of proxies in yfinance so this might not be an issue).
Also if timeout is None the request will hang indefinitely, so maybe default to something reasonable.
https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts
Maybe to make sure any new code do not call it without setting these do not make them optional and fore timeout to something else than None.

fredrik-corneliusson (changes requested):
Looks good, just had a few suggestions.

PR #1118, opened 2022-10-25
Better handling of error from yahoo API, added missing pytz dependency and fixed if statement syntax warnings

fredrik-corneliusson:


ValueRaider:
I'm curious about the new else block in _fetch_ticker_tz(). What scenario required it?

Issue #1111, opened 2022-10-24
Yfinance Intermittently working


locfinessemonster:
I have used yfinance for a while using a very old mac and haven't had any problems. I recently upgraded my computer to Windows 11 and am trying to run basic yfinance functions like yf.download in jupyter notebook on a chrome browser and I keep gettting the error message  "No data found for this date range, symbol may be delisted" as well as error 10060. I have tried downgrading to older library versions and using the "pip install yfinance --upgrade --no-cache-dir" recommended fix but I am still getting this problem. The workaround I came up with is to keep retrying the yf.download and sleeping for a couple seconds if it fails. It seems to eventually download after around 10 tries or so but this doesn't seem to be the most elegant solution. I have tried changing my firewall settings and it doesn't seem like my computer is blocking anything. I am pretty stumped, if you have any suggestions I would really appreciate it!

locfinessemonster:
Code with errors that I am seeing:
![image](https://user-images.githubusercontent.com/74224128/197453374-9ee2d61b-85ba-4c37-9de0-90523456f041.png)
![image](https://user-images.githubusercontent.com/74224128/197453437-420b7490-ce1b-431d-80be-2648960429c6.png)

locfinessemonster:
Work around code:
![image](https://user-images.githubusercontent.com/74224128/197453869-f2452de6-97e3-45bd-93e6-1d6485a0ffcd.png)

locfinessemonster:
Longer sleep interval between requests seems to help me get it consistently on the second try, still a mystery as to why I consistently can't get it on the first try. 
![image](https://user-images.githubusercontent.com/74224128/197455377-df044570-0707-4b92-b000-339060b373f0.png)

ValueRaider:
Did you try 0.1.74 or older?

jordangrodecki:
Also had this bug after building a new venv - spent ages before dropping back to 0.1.74 which works fine.

ValueRaider:
We think we know why and plan a release very soon.

locfinessemonster:
> Also had this bug after building a new venv - spent ages before dropping back to 0.1.74 which works fine.

I tried pip install yfinance==0.1.7.4 and still got the same issue.
Tried yfinance==0.1.67 and got the same error

locfinessemonster:
> We think we know why (related to #1076) and plan a release very soon.

Awesome, thank you for the update.

ValueRaider:
[...]

PR #1122, opened 2022-10-26
Fixed some of the formatting warnings reported by PyCharm IDE

fredrik-corneliusson:
Fixed formatting warning reported by the code inspection in PyCharm.
This will make it easier to spot the real errors when you use a code inspector tool.
Project should probably use some sort of automatic code formatter and checker in the future but this is a little step in adhering to PEP8.

### DIFF
--- a/yfinance/multi.py
+++ b/yfinance/multi.py
@@ -29,10 +29,10 @@ from . import Ticker, utils
 from . import shared
 
 
-def download(tickers, start=None, end=None, actions=False, threads=True, ignore_tz=True, 
+def download(tickers, start=None, end=None, actions=False, threads=True, ignore_tz=True,
              group_by='column', auto_adjust=False, back_adjust=False, repair=False, keepna=False,
              progress=True, period="max", show_errors=True, interval="1d", prepost=False,
-             proxy=None, rounding=False, timeout=10, **kwargs):
+             proxy=None, rounding=False, timeout=10):
     """Download yahoo tickers
     :Parameters:
         tickers : str, list
@@ -126,8 +126,8 @@ def download(tickers, start=None, end=None, actions=False, threads=True, ignore_
             data = _download_one(ticker, period=period, interval=interval,
                                  start=start, end=end, prepost=prepost,
                                  actions=actions, auto_adjust=auto_adjust,
-                                 back_adjust=back_adjust, repair=repair, keepna=keepna, 
-                                 proxy=proxy, 
+                                 back_adjust=back_adjust, repair=repair, keepna=keepna,
+                                 proxy=proxy,
                                  rounding=rounding, timeout=timeout)
             shared._DFS[ticker.upper()] = data
             if progress:
@@ -145,7 +145,7 @@ def download(tickers, start=None, end=None, actions=False, threads=True, ignore_
 
     if ignore_tz:
         for tkr in shared._DFS.keys():
-            if (shared._DFS[tkr] is not None) and (shared._DFS[tkr].shape[0]>0):
+            if (shared._DFS[tkr] is not None) and (shared._DFS[tkr].shape[0] > 0):
                 shared._DFS[tkr].index = shared._DFS[tkr].index.tz_localize(None)
 
     if len(tickers) == 1:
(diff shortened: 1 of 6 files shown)


yfinance/multi.py
```diff
@@ -29,10 +29,10 @@
 from . import shared
 
 
-def download(tickers, start=None, end=None, actions=False, threads=True, ignore_tz=True, 
+def download(tickers, start=None, end=None, actions=False, threads=True, ignore_tz=True,
              group_by='column', auto_adjust=False, back_adjust=False, repair=False, keepna=False,
              progress=True, period="max", show_errors=True, interval="1d", prepost=False,
-             proxy=None, rounding=False, timeout=10, **kwargs):
+             proxy=None, rounding=False, timeout=10):
```
ValueRaider, line 35: Instead of dropping kwargs, can you add it to `_download_*()` calls below. I think real problem is it's not being passed through.

ValueRaider (commented):
Just that one comment.

yfinance/multi.py
fredrik-corneliusson: Not sure I understand the *args and **kwargs that was added to functions but never used.
Is it to support deprecated arguments or is there plans to use variable args in the future? Then it can be added when needed without breaking the API.

ValueRaider: https://github.com/ranaroussi/yfinance/blob/fb5c67b3bddb103f4b53f2e2274355a3c166b3d7/yfinance/base.py#L143

fredrik-corneliusson: I've never seen this in other code (unless they are used) and they were reported as errors by the code inspection.
Keeping them just confuses users of the API and will make extra arguments typos in keyword arguments to pass silently.
I would suggest to keep them out. Do you still want them added in _download_*()?

fredrik-corneliusson: Ah now I think I get what you mean.
The history function supports kwargs to pass debug options.
```
            **kwargs: dict
                debug: bool
                    Optional. If passed as False, will suppress
                    error message printing to console.
                raise_errors: bool
                    Optional. If True, then raise errors as
                    exceptions instead of printing to console.
```

I think it would make more sense to just add them as normal keyword arguments. That way coding editors would be able to do auto complete and detect typos.
Converting them to normal keyword arguments would not break the API.

There are plans for many debug related options then I would add it as a normal keyword argument that expects a dict or a special debug options object.

ValueRaider: They should stay in some form - `debug_mode` useful for debugging, and I use `raise_errors` in my loops. In fact I just asked a user with issue to set `debug_mode` (#1111), so I need to patch `multi.py` anyway.

It shouldn't confuse users because most don't need to know.

fredrik-corneliusson: saw that there is a undocumented internal "many"-option passed to the history function. Looks like a way to supers raising exceptions and printing errors. So it is to not litter the API with internal options.

I would suggest to have a public method without it and a protected method with the implementation that supports it.

ValueRaider: Ok that's fair

fredrik-corneliusson:
I think there was a bug in the pull request, 
```
Exception in thread Thread-35:
  File "\Python\Python38\lib\threading.py", line 932, in _bootstrap_inner
    self.run()
  File "\Python\Python38\lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "\venv\lib\site-packages\multitasking\__init__.py", line 104, in _run_via_pool
    return callee(*args, **kwargs)
  File "\yfinance-tz-cache\yfinance\multi.py", line 202, in _download_one_threaded
    data = _download_one(ticker, start, end, auto_adjust, back_adjust, repair,
  File "\yfinance-tz-cache\yfinance\multi.py", line 215, in _download_one
    return Ticker(ticker).history(
  File "\yfinance-tz-cache\yfinance\base.py", line 339, in history
    df = utils.safe_merge_dfs(df, dividends, interval)
  File "\yfinance-tz-cache\yfinance\utils.py", line 455, in safe_merge_dfs
    raise Exception("Lost data during merge despite all attempts to align data (see above)")
Exception: Lost data during merge despite all attempts to align data (see above)
```
Consider reverting

ValueRaider:
That revert got messy so I nuked it all.

ValueRaider:
Revert got messy so I nuked it all.

PR #1048, opened 2022-08-12
Fix timezone & datetime handling

ValueRaider:
This big update fixes all problems related to timezones:
1. Convert Yahoo datetimes from UTC into timezone of exchange [#978](https://github.com/ranaroussi/yfinance/issues/978) [#1010](https://github.com/ranaroussi/yfinance/issues/1010)
2. If user provided timezone-naive datetimes then assume they mean timezone of exchange [#1010](https://github.com/ranaroussi/yfinance/issues/1010)
3. Use `datetime` module for epoch read/write instead of `time` [#954](https://github.com/ranaroussi/yfinance/issues/954) [#949](https://github.com/ranaroussi/yfinance/issues/949)
4. Yahoo can mess up DST so day start time is off by an hour - e.g. Brazil around Jan-2022 is 23:00 instead of 00:00. So detect and adjust to 00:00
5. Process user-provided `tz` argument correctly - use it to convert, not localize

This requires knowing exchange timezone, which means looking up `Ticker.info[]` but that's expensive! Splitting `info[]` creation out of `_get_fundamentals()` into `_get_info()` helped but still slow. So solution is to cache the timezone in users cache folder, because why would it change?

I've been using this personally for a month now, no problems.

Will fix these issues: [#971](https://github.com/ranaroussi/yfinance/issues/971)

These issues probably fixed too (difficult to test): [#890](https://github.com/ranaroussi/yfinance/issues/890) [#871](https://github.com/ranaroussi/yfinance/issues/871) [#805](https://github.com/ranaroussi/yfinance/issues/805) [#413](https://github.com/ranaroussi/yfinance/issues/413)

PR #1138, opened 2022-10-31
Move get 'trailingPegRatio' into _get_info(), simplify & optimise

### LINKED ISSUE #1114: Found code that will always fail in silence

ValueRaider:
Resolve #1114, which was caused by redundant code so just removed it.

Move the logic into `_get_info()` - I introduced this bug in my big timezone update (#1048), so 'trailingPegRatio' would have disappeared from `info[]`.

Simplify the code as loop is unnecessary. Also disable it, replacing with a direct GET - scraping a whole page for one variable is a waste. Keep the original code in case do want all key statistics in future.

@rickturner2001 I believe you first requested that 'trailingPegRatio' be fixed (#903).

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -712,6 +712,50 @@ class TickerBase():
         except Exception:
             pass
 
+        # Complementary key-statistics. For now just want 'trailing PEG ratio'
+        session = self.session or _requests
+        keys = {"trailingPegRatio"}
+        if len(keys)>0:
+            # Simplified the original scrape code for key-statistics. Very expensive for fetching
+            # just one value, best if scraping most/all:
+            #
+            # p = _re.compile(r'root\.App\.main = (.*);')
+            # url = 'https://finance.yahoo.com/quote/{}/key-statistics?p={}'.format(self.ticker, self.ticker)
+            # try:
+            #     r = session.get(url, headers=utils.user_agent_headers)
+            #     data = _json.loads(p.findall(r.text)[0])
+            #     key_stats = data['context']['dispatcher']['stores']['QuoteTimeSeriesStore']["timeSeries"]
+            #     for k in keys:
+            #         if k not in key_stats or len(key_stats[k])==0:
+            #             # Yahoo website prints N/A, indicates Yahoo lacks necessary data to calculate
+            #             v = None
+            #         else:
+            #             # Select most recent (last) raw value in list:
+            #             v = key_stats[k][-1]["reportedValue"]["raw"]
+            #         self._info[k] = v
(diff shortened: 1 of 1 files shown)


rickturner2001:
I completely forgot about this! 

@ValueRaider  Why are we setting **keys** to a set with a string inside, just to check if  len(keys) > 0 one line below?

ValueRaider:
@rickturner2001 Just habit of defensive coding - I made an iterable so check length. No harm.

PR #1140, opened 2022-11-02
Improve bad ticker handling

ValueRaider:
Improve handling of bad ticker, and add a good test

Add session to tests/ticker.py

Remove redundant `get_earnings_history()`

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -154,7 +154,7 @@ class TickerBase:
             tz = self._get_ticker_tz(debug, proxy, timeout)
             if tz is None:
                 # Every valid ticker has a timezone. Missing = problem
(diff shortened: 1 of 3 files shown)


yfinance/base.py
```diff
@@ -616,6 +616,11 @@ def _get_info(self, proxy=None):
         ticker_url = "{}/{}".format(self._scrape_url, self.ticker)
 
         # get info and sustainability
+        json_data = utils.get_json_data_stores(ticker_url, proxy, self.session)
```
ValueRaider, line 619: This function call can fail, need to handle IndexError
`json_str = html.split('root.App.main =')[1].split(`
> IndexError: list index out of range

PR #1130, opened 2022-10-29
Formatting and code cleanups

fredrik-corneliusson:
Fixed some of the formatting errors reported by PyCharm as well as simplified some code constructs.

PR #1144, opened 2022-11-05
Fixed some of the formatting errors reported by PyCharm as well as simplified some code constructs.

fredrik-corneliusson:
This is same at previous PR just with fixed merge conflicts

### DIFF
--- a/yfinance/utils.py
+++ b/yfinance/utils.py
@@ -129,7 +129,7 @@ def get_json_data_stores(url, proxy=None, session=None):
     # return data
     new_data = _json.dumps(data).replace('{}', 'null')
     new_data = _re.sub(
-        r'\{[\'|\"]raw[\'|\"]:(.*?),(.*?)\}', r'\1', new_data)
+        r'{[\'|\"]raw[\'|\"]:(.*?),(.*?)}', r'\1', new_data)
 
     return _json.loads(new_data)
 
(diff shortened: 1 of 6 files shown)


ValueRaider:
See my review in that last PR about a bug you introduced, because it's still present: https://github.com/ranaroussi/yfinance/pull/1130

fredrik-corneliusson:
Sorry, cannot see the review, maybe it got lost when I merged

ValueRaider:
#1130

fredrik-corneliusson:
image
This is what I see. no review at all.

fredrik-corneliusson:
As that PR is gone for me could you just please add the comment on this PR?

yfinance/utils.py
```diff
@@ -470,45 +474,47 @@ def fix_Yahoo_returning_live_separate(quotes, interval, tz_exchange):
...
-                last_rows_same_interval = dt1.month==dt2.month
+                last_rows_same_interval = dt1.month == dt2.month
             elif interval == "3mo":
-                last_rows_same_interval = dt1.year==dt2.year and dt1.quarter==dt2.quarter
-            if last_rows_same_interval:
+                last_rows_same_interval = dt1.year == dt2.year and dt1.quarter == dt2.quarter
+            else:
+                last_rows_same_interval = None
+
+            if last_rows_same_interval is not None:
```
ValueRaider, line 500: This change has modified behaviour, now it's always dropping last row even when it shouldn't. `last_rows_same_interval` default must be False not None, and the `if` must check for True/False

If you aren't convinced by this then do the following:

- in base.py, duplicate the call to `utils.fix_Yahoo_returning_live_separate()` so it is called twice

Then fetch weekly data. The first call will work correctly, fixing a Yahoo error. But the second call, acting on data which no longer has the Yahoo error, will remove all of this weeks data - incorrect.

=> So if the data never had the Yahoo error, e.g. because day was Monday, then first call will remove this weeks data.

ValueRaider:
Spooky! I've copied over my comment.

yfinance/utils.py
fredrik-corneliusson: Good catch, have updated the PR

PR #1154, opened 2022-11-09
Major refactoring

fredrik-corneliusson:
Made fundamentals, quote, analysis and holders into lazy scraped modules in order to improve performance and code maintainability.

### DIFF
--- a/yfinance/scrapers/fundamentals.py
+++ b/yfinance/scrapers/fundamentals.py
@@ -1,0 +1,150 @@
+import pandas as pd
+
+from yfinance import utils
+from yfinance.data import TickerData
+
+
+class Fundamentals:
+    _SCRAPE_URL_ = 'https://finance.yahoo.com/quote'
+
+    def __init__(self, data: TickerData, proxy=None):
+        self._data = data
+        self.proxy = proxy
+
+        self.ticker_url = "{}/{}".format(self._SCRAPE_URL_, self._data.ticker)
+
+        self._earnings = None
+        self._financials = None
+        self._shares = None
+
+        self._financials_data = None
+        self._fin_data_quote = None
+        self._basics_already_scraped = False
+        self._already_scraped_financials = False
+
+    @property
+    def earnings(self):
+        if self._earnings is None:
+            self._scrape_earnings(self.proxy)
+        return self._earnings
+
+    @property
+    def financials(self):
+        if self._financials is None:
+            self._scrape_financials(self.proxy)
+        return self._financials
+
+    @property
+    def shares(self):
+        if self._shares is None:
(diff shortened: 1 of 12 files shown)


fredrik-corneliusson:
This is quite a patch, and will probably need to be discussed and improved.
It looks more daunting than it is it's mostly just code broken out into more manageable classes.
This should make it easier to get an overview of the code and reduce risk for introducing bugs. It also have a quite nice performance improvement as it decreases unnecessary scraping.
Before and after times:
image

ValueRaider:
I plan to delay review & merge until first I push out rc2 with all other merges so far.

fredrik-corneliusson:
ok, sounds good, when is rc2 release scheduled?

ValueRaider:
"Schedule" is a strong word, but aiming within a week. I need to push one bugfix for dividends/splits. Then fix bugs in my yfinance_cache and run it for a few trading days, it can catch edge-case issues.

yfinance/scrapers/fundamentals.py
```diff
@@ -0,0 +1,150 @@
...
+                                 'periodType', 'currencyCode'], inplace=True)
+            shares.rename(
+                columns={'reportedValue': "BasicShares"}, inplace=True)
+            self._shares = shares
+        except Exception:
+            pass
+
+    def _scrape_financials(self, proxy):
+        self._scrape_basics(proxy)
+        if self._already_scraped_financials:
```
ValueRaider, line 101: Why not just check `if self._financials is not None:` ? Then don't need `_already_scraped_financials`.

Comment also applies to `_basics_already_scraped`

fredrik-corneliusson: I have already rewritten this so financials is fetched in a better way and avoids unnecessary requests.  Can update this draft pull request

fredrik-corneliusson:
Some general notes:
Note about the added type annotations to the API: I have used this version quiet a bit in Jupyther notebook and having type annotations makes code completion work much better.

One thing that would be good to know more about is the use of proxy argument that is passed per "get"-function.
Is that by design or would it make sense to have a proxy set on the Ticker instance and use that for all reqquests? Then the code could be simplified a bit. But maybe there are good use cases for having different proxies for each request.
Anyway that would be for a new major version as it would break the current API.

ValueRaider:
I don't use proxies so no opinion. May be best to ask those few that do - https://github.com/ranaroussi/yfinance/issues?q=proxy+in%3Atitle

Anything else pending before I merge?

ValueRaider (approved):

fredrik-corneliusson:
Just included fix for #1172. I think it's now ready for merge.

PR #1157, opened 2022-11-10
Fix fundamentals regression bug

fredrik-corneliusson:
The wrong data was returned for balance_sheet and cashflow


fredrik-corneliusson:
@ValueRaider This should probably go in the next RC2 release.

fredrik-corneliusson:
Yes that should be fine

Issue #980, opened 2022-04-01
Yfinance period='Max' option bug


LucGuj:
The yfinance version 0.1.70 shows a bug when it comes to download historical data using the period='max' option. In fact if I run the code below I get data starting from 1950 while the data go back to 1928 (colab link: https://colab.research.google.com/drive/1eByqpQGTxKH7KmYUy-FDri6HyYALKzzw#scrollTo=spxN65Qi4EJ4). As further validation if I run the same code using yfinance 0.1.63 I get the proper historical data starting from 1928 (colab link: https://colab.research.google.com/drive/1XvUF6ngMskT-aKduMTi_jgsvsRvSYLz_#scrollTo=iR3GS3f9kU3D) 

```
import yfinance as yf

ticker = '^GSPC'
Stock = yf.Ticker(ticker)
Stock_Data = Stock.history(period="max",auto_adjust = False)

Stock_Data
```

![image](https://user-images.githubusercontent.com/14195165/161288556-99482bb5-fe2e-48d2-af79-ca5ee89dd74f.png)

![image](https://user-images.githubusercontent.com/14195165/161288758-de167003-313a-4b2e-863a-26c213c8bb3f.png)

Thanks in advance for your support and help
Luca

ppeloton:
Raised https://github.com/ranaroussi/yfinance/pull/1169 request to fix issue

PR #1169, opened 2022-11-13
Fixing issue 980 by changing default timestamp for start parameter in…

ppeloton:
default timestamp for "start" parameter was set to 01/01/1950, so no data before that date was fetched. Set default value to 01/01/1900. This should fix https://github.com/ranaroussi/yfinance/issues/980

… base.py


ValueRaider:
Can you create another PR to merge into `release/0.1`. Currently we have two versions on the go.

Issue #632, opened 2021-03-16
Ticker info, what are do all the dict fields mean?


caballerofelipe:
Is there somewhere where I can find what the yf.Ticker.info returned fields mean?

I am new to the library so some fields don't make sense to me and I would like to read their definition.

tomazbm:
you can try:

```
msft = yf.Ticker('msft')
last_dividend = msft.info
for key, value in last_dividend.items():
    print(key)
```

And then google the keys you are interested in.

silvavn:
We for sure need to document what each property means.

ValueRaider:
Can we revive this issue for a discussion?

Do users still think this requires `yfinance` to document, given it is simply forwarding what Yahoo has?

`info` has ~150 entries for a stock, does each need documenting? If not all, then which and why not possible to simply refer to 3rd party e.g. Investopedia?

ValueRaider:
@0silverback0

ValueRaider:
I think is Yahoo's job to document their data they return, not us.

PR #1178, opened 2022-11-17
ticker info docs

0silverback0:
Would this be  a good start for documentation of the Ticker.info key definitions. If so I can continue on


0silverback0:
I was addressing issue #632 about documentation for the keys returned from Ticker.info()

ValueRaider:
Ok. But before diving into documenting first need to think carefully about design. E.g.
- I see a lot of copy-pasting Investopedia - why not just link directly?
- Where to put documentation? In another Python data structure? On website?

0silverback0:
I was thinking the documentation for this particular information would be a page in root in order to keep the look as clean as possible.

I left the synopsis and link for the developer to be able to research in more depth if need be. 

I believe providing the developer with as much information possible as to how to interact with your API is the responsibility of the API owner.

fredrik-corneliusson:
Keep in mind that if it's documentation from other sources it has to have a open source license permitting copying. If not it would  probably result in a copyright violation.

0silverback0:
> Keep in mind that if it's documentation from other sources it has to have a open source license permitting copying. If not it would  probably result in a copyright violation.

I believe investopedia is ok but I will double check it and make necessary changes. I would like to know if this is something that is still wanted before I continue working?

fredrik-corneliusson:
Hi, I think with every new bit in information we add it will have to be supported and updated causing extra work for current maintainer. If there are good external sources for that information elsewhere maybe it would be better to document where to find it?
But what I think is really needed is updated documentation for the yfinance API (in the readme and or docstrings). Especially for the upcoming 0.2 changes in Dev branch that has changed quite a bit.

0silverback0:
Ok I'll take a look at it and if I can handle it I'll get right to work on that

CollieIsCute:
@ValueRaider  Aside from Ticker.info, I think it'll be awesome if we make a github wiki which record all api of yfinance, such as
- Definition and meaning of each parameters within yfinance.
- Type of return objects

With the docs, newbies like me could spend less time tracing source code. I spent much time reading source code to find out something like "what's difference between `download()` and  `history()`" and "Which is more efficient when iterating? List of Ticker objs or Tickers obj?"

ValueRaider:
@CollieIsCute Good point but is different to this PR. Can you raise in another Issue/PR, and maybe someone can volunteer to improve API doc.

ValueRaider:
Wiki is live so this should go into Wiki

PR #1183, opened 2022-11-20
Made fetching earnings_dates faster

fredrik-corneliusson:
Avoid unnecessary request when fetching earnings_date. Added support to limit argument to only fetch as many as needed.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -60,7 +60,6 @@ class TickerBase:
         self._shares = None
 
         self._earnings_dates = {}
-        self._earnings_history = None
 
         self._earnings = None
         self._financials = None
@@ -847,11 +846,13 @@ class TickerBase:
         self._news = data.get("news", [])
         return self._news
 
-    def get_earnings_dates(self, limit=1000, proxy=None) -> Optional[pd.DataFrame]:
+    def get_earnings_dates(self, limit=12, proxy=None) -> Optional[pd.DataFrame]:
         """
         Get earning dates (future and historic)
-        :param limit: max amount of upcoming and recent earnings dates, set to smaller value to reduce amount
-                      of requests needed if ticker has a long history that is not of interest.
+        :param limit: max amount of upcoming and recent earnings dates to return.
+                      Default value 12 should return next 4 quarters and last 8 quarters.
+                      Increase if more history is needed.
+
         :param proxy: requests proxy to use.
         :return: pandas dataframe
         """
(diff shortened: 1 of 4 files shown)


yfinance/base.py
```diff
@@ -916,7 +927,7 @@ def get_earnings_dates(self, proxy=None) -> Optional[pd.DataFrame]:
 
         dates = dates.set_index("Earnings Date")
 
-        self._earnings_dates = dates
+        self._earnings_dates[limit] = dates
```
ValueRaider, line 930: I don't understand purpose of this 'dict cache'. data.py is already caching the data.

yfinance/base.py
```diff
@@ -847,11 +847,18 @@ def get_news(self, proxy=None):
         self._news = data.get("news", [])
         return self._news
 
-    def get_earnings_dates(self, proxy=None) -> Optional[pd.DataFrame]:
-        if self._earnings_dates is not None:
-            return self._earnings_dates
+    def get_earnings_dates(self, limit=1000, proxy=None) -> Optional[pd.DataFrame]:
```
ValueRaider, line 850: Set `limit` default to 12, that will be enough data for most users - next 4 quarters and last 8 quarters.
Will need to think about documentation so users know how to get full history, i.e. `get_earnings_dates()` vs `earnings_dates` property.

yfinance/base.py
```diff
@@ -916,7 +927,7 @@ def get_earnings_dates(self, proxy=None) -> Optional[pd.DataFrame]:
 
         dates = dates.set_index("Earnings Date")
 
-        self._earnings_dates = dates
+        self._earnings_dates[limit] = dates
 
         return dates
```
ValueRaider, line 933: Do me favour and delete `get_earnings_history()` below and related API points. It's entirely redundant, an incomplete implementation of `get_earnings_dates()`.

ValueRaider (changes requested):
3 comments

yfinance/base.py
fredrik-corneliusson: Caching this here will avoid doing the reparsing of the html table and the data manipulation.
Also the data cache is limited to the 64 latest unique requests so if you have some ticker instances active it will result in cache misses.

yfinance/base.py
fredrik-corneliusson: Sounds good, removed it.

yfinance/base.py
fredrik-corneliusson: Fixed

ValueRaider: Adjust README too, currently says "# show all earnings dates"

ValueRaider (approved):

Issue #1180, opened 2022-11-19
Extend & improve caching


ValueRaider:
To support PR #926 will be useful to cache `info["quoteType"]`.

I suggest refactoring `_TzCache` to be generic by accepting a new third argument - `key`, e.g. `key="quoteType"`. Need to refactor too `_KVStore` migrate old databases.

A good time to also move cache code into separate file.

ValueRaider:
Copying @fredrik-corneliusson comment to from PR here, in case branch nuke deletes it.

I do not see use for expiration TZ cache as it is more or less static data. But for a prober cache implementation I thinks it makes sense depending on what is being cached and how long it will be valid.
If it ever of interest here is the code:
```
class TimeStampedKVCache:
    """
    Simple Sqlite backed key/value cache with timestamp, key and value are strings timestamp is datetime.
    Should be thread safe.
    """
    DSL = """
    create table if not exists "kv" 
    (key TEXT primary key, value TEXT, ts timestamp) 
    without rowid
    """

    def __init__(self, filename):
        self._cache_mutex = Lock()
        with self._cache_mutex:
            self.conn = sqlite3.connect(filename,
                                        timeout=10,
                                        check_same_thread=False,
                                        detect_types=sqlite3.PARSE_DECLTYPES)
            self.conn.execute('pragma journal_mode=wal')
            self.conn.execute(self.DSL)
            self.conn.commit()
        atexit.register(self.close)

    def close(self):
        if self.conn is not None:
            with self._cache_mutex:
                self.conn.close()
                self.conn = None

    def get(self, key: str) -> Optional[str]:
        """Get value for key if it exists else returns None"""
        item = self.conn.execute('select value from "kv" where key=?', (key,))
        if item:
            return next(item, (None,))[0]

    def get_ts(self, key: str) -> Optional[datetime]:
        """Get time stamp for key if it exists else returns None"""
        item = self.conn.execute('select ts from "kv" where key=?', (key,))
        if item:
            return next(item, (None,))[0]

    def set(self, key: str, value: str, ts: datetime = None):
        if ts is None:
            ts = datetime.utcnow()
        naive_ts = ts.replace(tzinfo=None)
        with self._cache_mutex:
            self.conn.execute('replace into "kv" (key, value, ts) values (?,?,?)', (key, value, naive_ts))
            self.conn.commit()

    def bulk_set(self, kvdata: Dict[str, str], ts: datetime = None):
        if ts is None:
[...]

PR #1166, opened 2022-11-11
Add support for capital_gains data

thirumalairajr:
Hi,

Adding support for capital_gains for mutual funds & etfs based on the issue https://github.com/ranaroussi/yfinance/issues/926

Result:
image

PR #1188, opened 2022-11-21
Extend cache to other info fields

ValueRaider:
Implement request #1180. Each key value has it's own table e.g. "Timezone", so can easily cache other `info` fields.

And moved cache to separate Python file.

Will allow fixing performance degradation from #1166

EDIT: Another way to fix performance degradation so this PR less urgent. #1203


ValueRaider:
@fredrik-corneliusson Any thoughts?

fredrik-corneliusson:
Looks like a good refactoring into separate module for cache.

However would it not be easier to have separate db:s (KVStores or caches) instead of adding separate tables?
That could be done by just having a dictionary pointing to caches:
```
caches = {'tz_cache':  Cache("tz_cache"), "another_cache":  Cache("another_cache")}
```
Or is there an upside to keep different caches in one sqlite-database I do not see?

My reasoning is that a more complex schema will introduce some common pitfalls that a developer has to know when working with SQL such as SQL injection issues (as the PR has), migration versioning etc. Also keeping them as separate files would make it easier for people without SQL knowledge to manage caches (and debug). for example clearing a specific cache would be as easy as removing a file instead of having to connect with sql-client and issue commands to drop a table.

When I implemented the SQLITE-based cache it was just as a quick way of fixing the immediate concurrency and performance issues with the previous csv-file-based cache but still treating the cache as a simple dictionary.

That said I think _KVStore could be improved to support support timestamps per key to allow for eviction. I have some code for that if it would be of interest.

What are your thoughts?

ValueRaider:
I didn't think of debugging aspect, that's a good point, justifies separate files. I'll nuke this branch.

I don't see a use case for expiration, but if you do then go ahead.

fredrik-corneliusson:
I do not see use for expiration TZ cache as it is more or less static data. But for a prober cache implementation I thinks it makes sense depending on what is being cached and how long it will be valid.
If it ever of interest here is the code:
```
class TimeStampedKVCache:
    """
    Simple Sqlite backed key/value cache with timestamp, key and value are strings timestamp is datetime.
    Should be thread safe.
    """
    DSL = """
    create table if not exists "kv" 
    (key TEXT primary key, value TEXT, ts timestamp) 
    without rowid
    """

    def __init__(self, filename):
        self._cache_mutex = Lock()
        with self._cache_mutex:
            self.conn = sqlite3.connect(filename,
                                        timeout=10,
                                        check_same_thread=False,
                                        detect_types=sqlite3.PARSE_DECLTYPES)
            self.conn.execute('pragma journal_mode=wal')
            self.conn.execute(self.DSL)
            self.conn.commit()
        atexit.register(self.close)

    def close(self):
        if self.conn is not None:
            with self._cache_mutex:
                self.conn.close()
                self.conn = None

    def get(self, key: str) -> Optional[str]:
        """Get value for key if it exists else returns None"""
        item = self.conn.execute('select value from "kv" where key=?', (key,))
        if item:
            return next(item, (None,))[0]

    def get_ts(self, key: str) -> Optional[datetime]:
        """Get time stamp for key if it exists else returns None"""
        item = self.conn.execute('select ts from "kv" where key=?', (key,))
        if item:
            return next(item, (None,))[0]

    def set(self, key: str, value: str, ts: datetime = None):
        if ts is None:
            ts = datetime.utcnow()
        naive_ts = ts.replace(tzinfo=None)
        with self._cache_mutex:
            self.conn.execute('replace into "kv" (key, value, ts) values (?,?,?)', (key, value, naive_ts))
            self.conn.commit()

    def bulk_set(self, kvdata: Dict[str, str], ts: datetime = None):
        if ts is None:
            ts = datetime.utcnow()
        naive_ts = ts.replace(tzinfo=None)
        records = tuple((k, v, naive_ts) for k,v in kvdata.items())
        with self._cache_mutex:
            self.conn.executemany('replace into "kv" (key, value, ts) values (?,?,?)', records)
            self.conn.commit()

    def delete(self, key: str):
        with self._cache_mutex:
            self.conn.execute('delete from "kv" where key=?', (key,))
            self.conn.commit()

```

Issue #1195, opened 2022-11-23
Question: Is there an efficient way to get the currency?


ymyke:
The `history()` function doesn't return the currency the prices are in.

In order to get the currency, one has to invoke the `get_info()` function, which is rather costly – on the order of a few seconds. 

Is there a more efficient way to get the currency?

ValueRaider:
There is but it's slightly hacky. Yahoo returns currency with price data but no official way to add it to Pandas table. Can add a `.currency` attribute to table but it'll disappear as soon as you modify table, e.g.:
```
df = yf.download(...)
print(df.currency)  # Good
print(df)  # Pandas won't show currency
df = df.copy()
print(df.currency)  # AttributeError exception
```
>>>

ymyke:
Thanks! I tried this, however, it doesn't work:

```python
df = yf.download("MSFT")
print(df.currency)  # produces AttributeError
```

Am I missing something?

ymyke:
I saw that Yahoo returns some meta data. Example below. The currency is in there. Along with all kinds of other information that could be useful here and there.

Feature request: Would be great if – upon retrieval – you could simply store the raw "meta" dictionary in the Ticker object so it would be accessible if necessary/useful.

(Maybe this is more or less what you were suggesting above and what you were describing there is not how it works today but how it could work in the future?)



```python
{'currency': 'USD',
 'symbol': 'COIN',
 'exchangeName': 'NMS',
 'instrumentType': 'EQUITY',
 'firstTradeDate': 1618407000,
 'regularMarketTime': 1669237204,
 'gmtoffset': -18000,
 'timezone': 'EST',
 'exchangeTimezoneName': 'America/New_York',
 'regularMarketPrice': 45.57,
 'chartPreviousClose': 43.39,
 'previousClose': 43.39,
 'scale': 3,
 'priceHint': 2,
 'currentTradingPeriod': {'pre': {'timezone': 'EST',
   'start': 1669194000,
   'end': 1669213800,
   'gmtoffset': -18000},
  'regular': {'timezone': 'EST',
   'start': 1669213800,
   'end': 1669237200,
   'gmtoffset': -18000},
  'post': {'timezone': 'EST',
   'start': 1669237200,
   'end': 1669251600,
   'gmtoffset': -18000}},
 'tradingPeriods': [[{'timezone': 'EST',
    'start': 1669213800,
    'end': 1669237200,
    'gmtoffset': -18000}]],
 'dataGranularity': '1m',
[...]

PR #1197, opened 2022-11-25
#1195 Implemented a way to get the raw metadata

fredrik-corneliusson:


ValueRaider:
I think fetching metadata separate from prices is wrong approach, should be returning what Yahoo returns with the prices. What if the second fetch is different? Yahoo uses a mix of data providers so possible.

I'm also not sure about hardcoding the interval, as metadata may be different with longer/shorter.

fredrik-corneliusson:
Agree, but should there be a public accessor for metadata?
If so, what should be the default behavior if history has not run and no metadata has been extracted yet?
Maybe raise an error with information that history needs to be called first?
Also should it be called history_metadata instead of just metadata?

ValueRaider:
> Also should it be called history_metadata instead of just metadata?

Definitely needs that context, "metadata" alone too vague.

All other questions apply to dividends and splits e.g. `get_dividends()`. Doesn't warn user if no prices fetched, but seems no one has complained so just copy that approach?

In case you hadn't noticed, @ymyke is implementing same feature  #1198, similar discussion.

fredrik-corneliusson:
Closing this PR as it just duplicates #1198, will continue discussion there.

PR #1198, opened 2022-11-25
Add meta info as new attribute `_meta` to TickerBase

ymyke:
Meta information is returned by Yahoo anyway and can for example be used
to check the currency of the data retuerned. Like so:

```python
from yfinance import ticker
msft = ticker.Ticker("MSFT")
hist = msft.history()
print(msft._meta["currency"])
```

See also https://github.com/ranaroussi/yfinance/issues/1195

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -46,6 +46,7 @@ class TickerBase:
         self.ticker = ticker.upper()
         self.session = session
         self._history = None
+        self._meta = {}
         self._base_url = _BASE_URL_
         self._scrape_url = _SCRAPE_URL_
         self._tz = None
@@ -233,6 +234,9 @@ class TickerBase:
                 else:
                     print('%s: %s' % (self.ticker, err_msg))
             return utils.empty_df()
+
+        # store the meta information to make it accessible from the outside
+        self._meta = data["chart"]["result"][0]["meta"]
 
         # parse quotes
         try:


ValueRaider:
Needs polish. Some quick initial thoughts:
- variable name needs to clearly indicate this is price metadata not e..g earnings
- why default to `{}` instead of `None`?
- needs documenting in README, and maybe a property accessor in ticker.py

ValueRaider:
Touch base with @fredrik-corneliusson , seems he's implemented as well. #1197

ymyke:
1. Variable name: How is this just price metadata? Looks like general metadata to me. So _meta seems a fair name. It's also called meta in the returned data dict. – If you don't agree, please suggest concrete name.
2. {} default: Because it's a dictionary. Similary, _expirations is defaulted to {} and _news to [].
3. README: Happy to add once we have alignment on 1, 2, 4.
4. Had a quick review of #1197, but that looks like a significant refactoring with uncertain deadline. Please chime in @fredrik-corneliusson if there is a clear timeline there.

ValueRaider:
1. I suppose it's more exchange metadata than prices - `currentTradingPeriod`, `exchangeTimezoneName`. Word "metadata" on it's own needs context - meta of what data? And to better distinguish against `info[]` which is also a form of general stock/company metadata.
2. Ok
4. Deadline irrelevant here, implementations take however long they need, get merged when good enough. Point is it's implementing same feature request.

fredrik-corneliusson:
@ymyke I made the PR before I saw you also made one, in retrospect and with @ValueRaider comments on that PR I think that mine can be scrapped.
I agree with @ValueRaider that it probably need a public accessor if it is supposed to be an official part of the API as accessing a private property ("_meta") is not something a consumer of the API is supposed to do.

I would suggest if the meta value is accessed and no history has been fetched a suitable  exception with message should be raised instead of just empty dict or None.

Regarding the naming, the API is already quite a handful, and careful naming of any new properties is needed. 
The method that sets the metadata is named history, maybe "history_metadata" would make it clear what produced it?

ymyke:
Alright, thanks for the discussion. I'll work some more on it. Accessor makes sense and I'll add one.

Open questions:

1. Naming: Either `exchange_metadata` or `history_metatadata`. I'd go with the former (`exchange_metadata`). → Agree?
2. Behavior if `history()` not called yet. 
a. Raise exception: I don't like that because none of the other accessors behave that way.
b. Just return `{}`. Works, in line with other accessors. Con: User needs to have implicit knowledge that an empty dict can mean either "no metadata available" or "no retrieval done yet".
c. Call `history()` to make sure the metadata gets called and stored. Also the approach by some of the other accessors, however, there it's usually with really history-related data. Here, we'd call `history()` to retrieve something not history-related. Pro: Somewhat clearer behavior from user's perspective. Con: Call to `history()` feels like a strange side effect; takes time; potential to override earlier history results that were retrieved with different parameters.
→ I'd go with b. WDYT?

Meta question, help please: My original PR was for main rather than dev. I simply changed that online to dev via Github GUI. But now I need to continue working on it. What is the best way to proceed to work on right files locally? Should I better start off with a new PR?

ValueRaider:
Meta Q - backup your code folder then try `git rebase dev` on your branch (remember to `pull dev` first). Should work here. Then verify by viewing diff of backed-up-branch vs main, and comparing against diff of rebased-branch vs dev.

fredrik-corneliusson:
1. Either is fine
2. I would probably want a way to differ missing metadata because of history not called or if the metadata was actually returned as empty. Making it return None instead of a empty dict would make it more obvious.  Either way I think it would be nice if accessing the property before it is set by a history call shows a warning message so that users will not be fooled that metadata is missing for ticker .

Regarding None vs {}:
> {} default: Because it's a dictionary. Similary, _expirations is defaulted to {} and _news to [].

In both these cases they are not returned directly but accessing them via the public API will trigger requests to get the actual data first. 
Note: In Python variables starting with underscore is considered private and not public:
https://docs.python.org/3/tutorial/classes.html#private-variables

ymyke:
Looking at the code some more, I'd reconsider the "no history behavior". Cf `get_dividends`, `get_splits`, `get_capital_gains`, etc. – they all get the history if it hasn't been retrieved yet. Example:

```python
    def get_dividends(self, proxy=None):
        if self._history is None:
            self.history(period="max", proxy=proxy)
        if self._history is not None and "Dividends" in self._history:
            dividends = self._history["Dividends"]
            return dividends[dividends != 0]
        return []
```

So, even though the exchange metadata is not part of the history key, it at least gets retrieved together with the history and I therefore suggest to go down that route, which would be option 2c above.

WDYT, @ValueRaider @fredrik-corneliusson ?

ValueRaider:
I wasn't around when `get_dividends` and `get_splits` were implemented, but I think they are terrible functions . Reasons: What if user called `history(interval="1m")` before? What if user wants different period?

So implement `meta` independently, however seems best. If you want to fetch history if None,  use `period=1d`.

ymyke:
Got it, thanks. 👍

ymyke:
Closing, see #1235.

Issue #1199, opened 2022-11-25
Bug: Timestamp conversion fails on Windows platform


ymyke:
On Windows (my Windows at least), this:

```python
import yfinance as yf
data = yf.download("AAPL")
```

produces an `OSError` because the `start` parameter gets set to a default value that cannot be converted to a time via `utcfromtimestamp`.

Here's the line where it gets set:
https://github.com/ranaroussi/yfinance/blob/744e70ffff2c4cf0a384e88e855af60862e71397/yfinance/base.py#L167

Here's the line where it later breaks:
https://github.com/ranaroussi/yfinance/blob/744e70ffff2c4cf0a384e88e855af60862e71397/yfinance/base.py#L319

Windows spec:
```
Edition	Windows 11 Pro
Version	22H2
Installed on	‎2022-‎10-‎15
OS build	22621.819
Serial number	PF39DFCV
Experience	Windows Feature Experience Pack 1000.22636.1000.0
```

System spec:
```
Processor	11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz   2.80 GHz
Installed RAM	16.0 GB (15.7 GB usable)
Device ID	3C2CE2C1-3557-41D1-BD95-D0997E6A4CA4
Product ID	00330-53928-86026-AAOEM
System type	64-bit operating system, x64-based processor
Pen and touch	No pen or touch input is available for this display
```

ymyke:
This seems also to be the cause to https://github.com/ranaroussi/yfinance/issues/1171

ymyke:
See PR for very straightforward fix. Feels rather crude but has the advantage that no other refactoring is necessary.

fredrik-corneliusson:
Sorry #1171 should have been closed as this is fixed in dev branch.
The fix is to use the pandas version that does not have the issue.
Is this exception present in the stable releases?

fredrik-corneliusson:
I think this is fixed in dev branch, @ymyke could you test and verify?

ymyke:
Yes, @fredrik-corneliusson, verified that this is fixed on dev. Closing this. Thanks!

Issue #1171, opened 2022-11-13
Dividends not working on windows (DEV)


fredrik-corneliusson:
Ticker dividends is affected by CPython windows bug: https://github.com/python/cpython/issues/81708

Converting negative timestamps on windows generates "OSError: [Errno 22] Invalid argument".

Code to reproduce on windows:
```
import yfinance as yf
t = yf.Ticker("IBM")
t.dividends
```

Traceback:
```
File ~\yfinance\base.py:304, in TickerBase.history(self, period, interval, start, end, prepost, actions, auto_adjust, back_adjust, repair, keepna, proxy, rounding, timeout, debug, raise_errors)
    302 dividends, splits = utils.parse_actions(data["chart"]["result"][0])
    303 if start is not None:
--> 304     startDt = _pd.to_datetime(_datetime.datetime.utcfromtimestamp(start))
    305     if dividends is not None:
    306         dividends = dividends[dividends.index >= startDt]

OSError: [Errno 22] Invalid argument
```

fredrik-corneliusson:
Fixed in dev branch

fredrik-corneliusson:
Doublechecked and it is fixed in dev branch (as part of refactoring PR).
The fix is to use pandas Timestamp instead of datetime .
```
# Does not work on windows if before 1970
pd.to_datetime(_datetime.datetime.utcfromtimestamp(start))
# Works
pd.Timestamp(start, unit='s')
```

PR #1200, opened 2022-11-25
Fix timestamp overflow issue

ymyke:
See https://github.com/ranaroussi/yfinance/issues/1199

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,14 +7,23 @@ assignees: ''
 
 ---
 
-*** READ BEFORE POSTING ***
+# READ BEFORE POSTING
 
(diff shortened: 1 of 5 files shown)


ValueRaider:
With this fix what is the oldest data returned for ticker "^GSPC"? I can go back to 1927.

ymyke:
1970-01-02

ValueRaider:
Need to go before 1970. Read this https://stackoverflow.com/a/41965216

fredrik-corneliusson:
I think this is already fixed as part of fix of #1171
https://github.com/ranaroussi/yfinance/issues/1171#issuecomment-1328076245

Running windows I cannot reproduce the error in dev.

ValueRaider:
@fredrik-corneliusson I agree

ymyke:
I verified and agree to. Branch deleted.

PR #1203, opened 2022-11-26
Get quote type from metadata instead info[] -> faster

ValueRaider:
Fix performance degradation from #1166 by replacing `info[]` access with metadata access. Requires small change to capital gains logic - always request it, but only return for ETFs & mutual funds.

@thirumalairajr Any obvious problems?


thirumalairajr:
@ValueRaider LGTM!! it should work

PR #1212, opened 2022-11-28
For ETF Tickers, use annual report expense ratio from fundProfile

bolstycjw:
This PR replaces the data['annualReportExpenseRatio'] with the one in data['fundProfile']['feeExpenseInvestment']['annualReportExpenseRatio'] so that it displays the correct value instead of None.


bolstycjw:
nvm, saw the other PR with fundProfile, will use that instead

PR #1214, opened 2022-11-29
Verify ticker history call

fredrik-corneliusson:
Make sure calling ticker history does not introduce any requests to yahoo that could trigger spam-block long before the chart API.


fredrik-corneliusson:
Wrong target

ValueRaider:
Just FYI, you can change the target up top

PR #1215, opened 2022-11-29
Test to verify ticker history request.

fredrik-corneliusson:
Make sure calling ticker history does not introduce any requests to yahoo that could trigger spam-block long before the chart API.

### DIFF
--- a/tests/ticker.py
+++ b/tests/ticker.py
@@ -135,6 +135,26 @@ class TestTickerHistory(unittest.TestCase):
     def tearDown(self):
         self.ticker = None
 
+    def test_history(self):
+        data = self.ticker.history("1y")
+        self.assertIsInstance(data, pd.DataFrame, "data has wrong type")
+        self.assertFalse(data.empty, "data is empty")
+
+    def test_no_expensive_calls_introduced(self):
+        """
+        Make sure calling history to get price data has not introduced more calls to yahoo than absolutely necessary.
+        As doing other type of scraping calls than "query2.finance.yahoo.com/v8/finance/chart" to yahoo website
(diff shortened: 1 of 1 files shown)


tests/ticker.py
```diff
@@ -135,6 +135,26 @@ def setUp(self):
...
+        Make sure calling history to get price data has not introduced more calls to yahoo than absolutely necessary.
+        As doing other type of scraping calls than "query2.finance.yahoo.com/v8/finance/chart" to yahoo website
+        will quickly trigger spam-block when doing bulk download of history data.
+        """
+        session = requests_cache.CachedSession(backend='memory')
+        ticker = yf.Ticker("GOOGL", session=session)
+        ticker.history("1y")
+        actual_urls_called = tuple(session.cache.urls)
+        expected_urls = (
+            'https://query2.finance.yahoo.com/v8/finance/chart/GOOGL?range=1y&interval=1d&includePrePost=False&events=div%2Csplits%2CcapitalGains',
```
ValueRaider, line 154: My concern is future changes could change/reorder request `params`, breaking this test. Why not just test only 1 url in cache?

fredrik-corneliusson: It was added so any change (subtitle or not) will get reviewed. Having the exact URL will give a nice detail assertion error of what has changed between old and new so it can be reviewed; if ok the test can then be updated with the new URL with by a copy/paste. 
Just comparing numbers will not give any details so the new URL could have the wrong parameters. 
Another reason is that the test will act as documentation of what request history method triggers. 
I see it as a safety net so no accidental extra request or change goes unnoticed.

PR #1216, opened 2022-11-29
Fix a couple of minor issues in README

ymyke:
- Typos in variable name
- `Ticker` doesn't support several tickers
- `Tickers` doesn't return named tuple
- "1m" in `download` would produce an error for longer timeframes, so changing the example to "5d"

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,9 +108,9 @@ msft.recommendations
 msft.recommendations_summary
 # show analysts other work
 msft.analyst_price_target
-mfst.revenue_forecasts
-mfst.earnings_forecasts
-mfst.earnings_trend
+msft.revenue_forecasts
+msft.earnings_forecasts
+msft.earnings_trend
 
 # show next event (earnings, etc)
 msft.calendar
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -160,7 +160,7 @@ the Ticker constructor.
 import requests_cache
 session = requests_cache.CachedSession('yfinance.cache')
 session.headers['User-agent'] = 'my-program/1.0'
-ticker = yf.Ticker('msft aapl goog', session=session)
+ticker = yf.Ticker('msft', session=session)
```
ValueRaider, line 163: Instead, replace with Tickers(), I think that was original intent

ymyke: `AttributeError: 'Tickers' object has no attribute 'actions'`

README.md
```diff
@@ -201,7 +200,7 @@ data = yf.download(  # or pdr.get_data_yahoo(...
         # fetch data by interval (including intraday if period < 60 days)
         # valid intervals: 1m,2m,5m,15m,30m,60m,90m,1h,1d,5d,1wk,1mo,3mo
         # (optional, default is '1d')
-        interval = "1m",
+        interval = "5d",
```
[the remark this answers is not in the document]
ymyke: The issue was that "1m" produces an error, not if it is a sensible param or not. "5d" does not produce an error and is a supported and documented error. So I'd leave it like this. Otherwise, I'd rather suggest to remove the 5d option altogether from the library.

README.md
ValueRaider: Good point, `Ticker()` is correct solution

PR #1219, opened 2022-11-30
Fix #1218

### LINKED ISSUE #1218: 0.2.0rc2 - error when _analyst_trend_details is None
yfinance: **0.2.0rc2** dev branch

```
import yfinance as yf

yf.Ticker("446070.KS").balance_sheet
```
Gives error:

```
[...]
if len(self._analyst_trend_details) != 0:
TypeError: object of type 'NoneType' has no len()
```

sooheon:
Don't check for len of NoneType

Cases where _analyst_trend_details is an empty dataframe are handled  in the try - except below.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -378,7 +378,7 @@ class TickerBase:
 
         data_cols = [c for c in ["Open","High","Low","Close","Adj Close"] if c in df_row.index]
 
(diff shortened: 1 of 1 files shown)


ValueRaider:
You need to change target to `dev`. See #1084

PR #1221, opened 2022-12-01
Default enable 'pretty' financials, explain in README

ValueRaider:
Change financials `pretty` format default to `True` to match historic behaviour. But I also added a note to README that other formats available, so hopefully both groups of users are happy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,17 +75,17 @@ msft.capital_gains
 # show share count
 msft.shares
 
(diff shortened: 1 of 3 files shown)


ValueRaider:
@git-shogg Any opinion on this? 'pretty' printing replicates historic behaviour, and can be disabled via the `get_*()` methods

PR #1235, opened 2022-12-12
Add `history_metadata` property

ymyke:
Here's the PR for `history_metadata`.

- See also https://github.com/ranaroussi/yfinance/issues/1195.
- I ended up calling it `history_metadata` after all rather than `exchange_metadata` because it not only includes exchange-related information (e.g., `currency`, `symbol`, `instrumentType`).
- Also, this is a new PR. The old one at https://github.com/ranaroussi/yfinance/pull/1198 can be closed.

(BTW, is it just me or is the code base pretty messy? For example, type hints are very inconsistent. And what is the purpose of `TickerBase` vs `Ticker` class?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,9 @@ msft.info
 # get historical market data
 hist = msft.history(period="max")
 
+# show meta information about the history (requires history() to be called first)
+msft.history_metadata
+
 # show actions (dividends, splits, capital gains)
 msft.actions
 
--- a/tests/ticker.py
+++ b/tests/ticker.py
@@ -136,7 +136,10 @@ class TestTickerHistory(unittest.TestCase):
         self.ticker = None
 
     def test_history(self):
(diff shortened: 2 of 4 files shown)


ValueRaider:
The try-catch is redundant, because "meta" is accessed on line 224 so code would already have blown up.

BTW: Code base is indeed messy, but to be fair it is OSS and very reliant on voluntary contributions. Code base is small so manageable.

ymyke:
Good point. I left the try/catch in but moved the block higher up and then used the new attribute to make further metadata accesses a little bit cleaner in that function.

ValueRaider:
Can I suggest replacing the epoch values with actual `pd.Timestamps`? E.g.

> _pd.Timestamp(regularMarketTime, unit='s').tz_localize("UTC").tz_convert(tz_exchange)

Can probably discard all other 'timezone' and 'gmtoffset' keys.

PR #1240, opened 2022-12-14
Using the opposite operator instead

marco-carvalho:
It is needlessly complex to invert the result of a boolean comparison. The opposite comparison should be made instead.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,16 @@
+name: ci
+on:
+  pull_request:
+    branches:
+      - master
+      - main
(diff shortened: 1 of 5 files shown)


ValueRaider:
This is just enforcing style guidelines. I think if doing that, update the entire codebase in one go. `flake8` helps with that.

marco-carvalho:
@ValueRaider I only found these 5 cases, I believe that the project does not have more than these

ValueRaider:
You are correct regarding the guidelines you checked, but there are others - whitespace around/after operators, unused variables.

Issue #1246, opened 2022-12-17
yfinance 0.1.*: Unable to load ticker information


untucked:
- Info about your system:
yfinance version: '0.1.90'
  - operating system: 
  -windows
- Simple code that reproduces your problem
```
ticker = 'AAPL'
print('ticker: %s'%(ticker))
yf_info = yf.Ticker(ticker)
bs = yf_info.balance_sheet
print(bs)
```
- The error message
Returns empty array: when I print bs
Empty DataFrame
Columns: [Open, High, Low, Close, Adj Close, Volume]
Index: []

BrsrkrBrown:
Having similar issue, when trying to print regularMarketPrice and volume.
printing yf.Ticker(ticker).info returns:
{'regularMarketPrice': None, 'preMarketPrice': None, 'logo_url': '', trailingPegRatio':None}

Travis-Barton:
Same problem for me on all stocks it seems.
I get `{'regularMarketPrice': None, 'preMarketPrice': None, 'logo_url': ''}` for any `info`

mrtomcollins:
Same problem for me

valankar:
Same here. Problem seems to have started today, perhaps some issue with Yahoo or API.

galashour:
same here

ValueRaider:
Looks like Yahoo has compressed or encrypted the data stores embedded in html that `yfinance` relied upon. E.g.

Before (via my session cache):
```
root.App || (root.App = {});
root.App.now = 1671025861087;
root.App.main = {"context":{"dispatcher":{"stores":{"PageStore":{"currentPageName":"quote", ...
```

After
```
> root.App || (root.App = {});
> root.App.now = 1671275144853;
> root.App.main = {"context":{"dispatcher":{"stores":"U2FsdGVkX18NwmzwZgyLozYcI3uwl2EXN\u002FWBnsSt9vGtBRZGAIiUIqV9mS1LKBRe42L1I1VjxC1qsKQ9g8KP4qeGBkrL28E0AG14...
```

Rogach:
They are encrypting the store data (it seems with AES) using `crypto-js` library.

Deobfuscated code is doing something like this:

```
var key = PBKDF2.create({ keySize: 8 }).compute(t._cs, JSON.parse(t._cr))
var plaintext = AES.decrypt("U2FsdGVk...", key)
var decoded = JSON.parse(decodeURIComponent(plaintext)
```

[...]

PR #1253, opened 2022-12-18
decode encrypted root.App.main.context.dispatcher.stores

Rogach:
WIP decryption code for issue #1246 

~~However this doesn't fix the issue completely, since QuoteSummaryStore is not found in the decrypted data - probably the data is now stored in some other place.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,7 @@ To install `yfinance` using `conda`, see
 -   [frozendict](https://pypi.org/project/frozendict) \>= 2.3.4
 -   [beautifulsoup4](https://pypi.org/project/beautifulsoup4) \>= 4.11.1
 -   [html5lib](https://pypi.org/project/html5lib) \>= 1.1
--   [pycryptodome](pycryptodome>=3.6.6) \>= 3.6.6
+-   [cryptography](https://pypi.org/project/cryptography) \>= 3.3.2
 
 ### Optional (if you want to use `pandas_datareader`)
 
--- a/README.md
+++ b/README.md
@@ -306,6 +306,7 @@ To install `yfinance` using `conda`, see
 -   [frozendict](https://pypi.org/project/frozendict) \>= 2.3.4
 -   [beautifulsoup4](https://pypi.org/project/beautifulsoup4) \>= 4.11.1
 -   [html5lib](https://pypi.org/project/html5lib) \>= 1.1
+-   [pycryptodome](pycryptodome>=3.6.6) \>= 3.6.6
 
 ### Optional (if you want to use `pandas_datareader`)
 
--- a/meta.yaml
+++ b/meta.yaml
@@ -26,7 +26,8 @@ requirements:
     - frozendict >=2.3.4
     - beautifulsoup4 >=4.11.1
     - html5lib >=1.1
-    - pycryptodome >=3.6.6
+    # - pycryptodome >=3.6.6
+    - cryptography >=3.3.2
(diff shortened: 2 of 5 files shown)


Rogach:
I was using an incorrect ticker value - with correct ticker value scraping seems to work without problems (for my limited use-cases).

Rogach:
But there seems to be some small differences - for example, if previously `yfinance.Ticker("^IXIC").info["regularMarketPrice"]` returned the value directly, now it returns a dict `{'raw': 10705.414, 'fmt': '10,705.41'}`.

havedill:
Confirming that i loaded this into my environment with `pip install 'yfinance @ git+https://github.com/ranaroussi/yfinance.git@refs/pull/1253/head'` 

It is working but agreed, I had to define ['raw'] in a decent number of places.

jesusignazio:
pycryptodome is now a dependency as well.

Rogach:
> pycryptodome is now a dependency as well

I'm not sure where to add it, so I need some help from the maintainers (maybe in `setup.py`? but what version boundary should I use?).

ValueRaider:
This looks great. I'll handle the rest.

Rogach:
Ah, so the `"raw"` troubles were due to me deleting bunch of code that was actually needed. Sorry about that.

ValueRaider:
I've just pushed a change to fix returned data structure, now the tests are passing. No worries @Rogach 

Minimum version should be 3.6.6 or later because [older contain a AES decryption bug](https://www.cve.org/CVERecord?id=CVE-2018-15560). I'll update setup.py etc, then give others a chance to question.

ValueRaider:
One more thing @Rogach . This also needs to be backported to branch `release/0.1`. I don't mind doing this, I just want to give you chance in case you want to be in commit log. You will still get full credit in PRs and release notes.

ValueRaider:
Regarding `cryptography` vs `pycryptodome`, is there any strong argument to choose one over other? `cryptography` has the high-level interface and larger community, but `pycryptodome` has been free of known CVEs for longest time. @fredrik-corneliusson

galashour:
Many thanks guys, this is much appreciated.

Is a new 'fixed' release is expected soon, or should we manually update a specific file on our local setup?

ValueRaider:
Wait for a fix, it will come soon. We just need to decide with crypto module to use.

fredrik-corneliusson:
> Regarding `cryptography` vs `pycryptodome`, is there any strong argument to choose one over other? `cryptography` has the high-level interface but this solution requires the low-level; it has larger community but `pycryptodome` has been free of known CVEs for longest time (unless `pycryptodome` has more unknown CVEs from less eyes). @fredrik-corneliusson
> 

Sorry for causing confusion, I worked on this fix before I saw @Rogach already had PR.
I only made my version as I had issues installing the crypto lib I thought the code in the comment used (required compiling).
But now that @Rogach has a working PR and pycryptodome seems to install on windows without issues my branch can be scrapped if there is no other compelling reason to use cryptography lib. 
I committed the code anyway in my repo but I was not planning to make a PR for it, forgot how good GitHub is at sniffing out commit messages in forks.

ValueRaider:
I decided to put both in the code but defaulting to `cryptography` just so I can get a release out. Will tidy when I get a good answer.

Rogach:
@ValueRaider Sorry, was away from computer. Thanks for backporting the code!

Issue #1274, opened 2023-01-04
sqlite database is locked when using multithread download


gogog22510:
- Info about your system:
  - yfinance version
    - 0.2.3
  - operating system
    - Windows WSL Ubuntu 20.04, running in Docker image from python:3.7-slim
- Simple code that reproduces your problem
```
import yfinance as yf
start_date = '2017-01-01'
end_date= '2022-04-29'
tickers = 'SPY TSLA NVDA MSFT'

data = yf.download(  # or pdr.get_data_yahoo(...
    # tickers list or string as well, cannot be an numpy.ndarray
    tickers=tickers,

    # start date
    start=start_date,

    # end date
    end=end_date,

    # fetch data by interval (including intraday if period < 60 days)
    # valid intervals: 1m,2m,5m,15m,30m,60m,90m,1h,1d,5d,1wk,1mo,3mo
    # (optional, default is '1d')
    interval="1d",

    # group by ticker (to access via data['SPY'])
    # (optional, default is 'column')
    group_by='ticker',

    # adjust all OHLC automatically
    # (optional, default is False)
    auto_adjust=False,

    # download pre/post regular market hours data
    # (optional, default is False)
    prepost=True,

    # use threads for mass downloading? (True/False/Integer)
    # (optional, default is True)
    threads=True,

    # proxy URL scheme use use when downloading?
    # (optional, default is None)
    proxy=None
)
```
- The error message
    - `- NVDA: OperationalError('database is locked')`

When running the multithread download from my flask server in Docker image, I sometime saw this error message, the failed symbol can be different one. The failing symbol will still produce empty dataset.

Any suggestion on how to know there is an error when calling `yf.download(threads=True)` or is there a way to fix the sqlite database lock issue? (Maybe by increasing the timeout?)

ValueRaider:
This *shouldn't* happen because database access is managed by a mutex, so is thread-safe. Are you running multiple Docker instances?

gogog22510:
I'm running one Docker instance with multiple threads, the command is from the Flask official guide, something like `gunicorn --bind :$PORT --workers 1 --threads 8 --timeout 0 'flaskr:create_app()'`

gogog22510:
[...]

PR #1276, opened 2023-01-04
Fix the database lock error in multithread download

gogog22510:
As per discussion in #1274, initialize the database in constructor to avoid database lock error in multithread environment.


gogog22510:
I tested locally with my app, looks good to me.

ValueRaider:
I've tidied the code, let me know if it broke again.

gogog22510:
The tidied change LGTM.

PR #1282, opened 2023-01-07
fix: Fix can't use proxy to access Yahoo api

yearnfar:
Fix can't use proxy to access Yahoo api

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -171,12 +171,6 @@ class TickerBase:
         if params["interval"] == "30m":
             params["interval"] = "15m"
 
-        # setup proxy in requests format
-        if proxy is not None:
-            if isinstance(proxy, dict) and "https" in proxy:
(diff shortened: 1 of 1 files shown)


yfinance/base.py
```diff
@@ -171,12 +171,6 @@ def history(self, period="1mo", interval="1d",
         if params["interval"] == "30m":
             params["interval"] = "15m"
 
-        # setup proxy in requests format
```
ValueRaider, line 174: Why delete this code?

jbpolle:
Hello,

You would also need to modify the function _get_proxy in data.py as sometimes the proxy variable is transformed in a "frozendict" and this doesn't work anymore. Something like that work for me:
if proxy is not None:
            if type(proxy) in [dict, frozendict] and "https" in proxy:

ValueRaider:
Superseded by #1371, closing.

PR #1283, opened 2023-01-07
Change default value to ignore_tz to False

### LINKED ISSUE #1275: Ticket data from different timezone are not aligned (bug introduced with #1085)
I am trying to find correlations between tickets in differnet timezone (one in the New York Stock exchange and the other in the London Stock Exchange). Due to changes to the timezone logic, the data in each row of `yfinance.download` are no longer the data of the Tickets at the same time. Using `ignore_tz=False` fixes this problem. This problem didn't exist with version `0.1.77` and previous. So I think by default `ignore_tz` should be set to `False` as that behaviour is consistent with the previous minor versions.

DE0CH:
With #1093, setting `ignore_tz=True` makes `download()` behave differently to previous version. This PR sets ignore_tz to False by default to bring the behavior of download() to be the same as 0.1.77 and prior.

Fixes #1275.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,8 +213,7 @@ data = yf.download(  # or pdr.get_data_yahoo(...
         interval = "5d",
 
         # Whether to ignore timezone when aligning ticker data from 
-        # different timezones. Default is True. False may be useful for 
-        # minute/hourly data.
+        # different timezones. Default is False.
         ignore_tz = False,
 
         # group by ticker (to access via data['SPY'])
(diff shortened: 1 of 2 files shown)


ValueRaider:
I've updated the docstring (that's what I meant by "doc", I should have been specific).

Can you `git squash` please.

DE0CH:
Thanks. Just rebased, squashed and pushed.

PR #1303, opened 2023-01-15
enum typing annotation for download period and inteval

rickturner2001:
I thought we could use enums to specify the period and interval values in the yf.download function. I usually struggle to remember all the legal options. Furthermore, we could use enums to leverage autocompletion.

Code example:

```python3
import yfinance as yf
from yfinance.types import TickerPeriod, TickerInterval

aapl = yf.download("AAPL", interval=TickerInterval.ONE_DAY, period=TickerPeriod.ONE_YEAR)
```


ValueRaider:
> I usually struggle to remember all the legal options

Doesn't the docstring handle this? And can deliberately enter illegal option to get Yahoo's message with intervals.

> leverage autocompletion

Fair request. I only wonder if Python `typing` hints can somehow allow this -  I don't know.

rickturner2001:
@ValueRaider  The docstring does indeed handle this, but say that I hover over the download method to get the definition. At this point, we can see that the type of both the period and interval parameters is either string or one of the newly implemented types, as opposed to simply being string. 

"And can deliberately enter illegal option to get Yahoo's message with intervals". Fair point but I would argue that it's not really an idea to force someone to pass an illegal option. Also, this is an optional approach, at the end of the day, we're using python for ease of programming. 

If enums don't look appealing, we might even do something like:


```python3
from typing import Union

def download(..., interval: Union["1m", "1d", ...]):
    ...
```

rickturner2001:
> > I usually struggle to remember all the legal options
> 
> Doesn't the docstring handle this? And can deliberately enter illegal option to get Yahoo's message with intervals.
> 
> > leverage autocompletion
> 
> Fair request. I only wonder if Python `typing` hints can somehow allow this - I don't know.


I forgot to address the last point in your response, so please allow me to do so.

I thought we could use the Literal type from the typing library like so:

```python3
ticker_period = Union[TickerPeriod, Literal["1d", "5d", "1mo", "3mo", "6mo", "1y", "2y", "5y", "10y", "ytd", "max"]]
```

**Calling the function in vsCode** ("python.analysis.typeCheckingMode": "basic")

```python3
import yfinance as yf
aapl = yf.download("AAPL", interval="am", period="am")
```

VsCode will alert us with a red squiggly line that am is not a legal Literal


**Warning**:
```
Argument of type "Literal['am']" cannot be assigned to parameter "period" of type "ticker_period" in function "download"
  Type "Literal['am']" cannot be assigned to type "ticker_period"
    "Literal['am']" is incompatible with "TickerPeriod"
    "Literal['am']" cannot be assigned to type "Literal['1d']"
    "Literal['am']" cannot be assigned to type "Literal['5d']"
    "Literal['am']" cannot be assigned to type "Literal['1mo']"
    "Literal['am']" cannot be assigned to type "Literal['3mo']"
    "Literal['am']" cannot be assigned to type "Literal['6mo']"
    "Literal['am']" cannot be assigned to type "Literal['1y']"
  ...Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues) 
```

The warning is very similar when we use **Pycharm**.

`Expected type 'Literal["1d", "5d", "1mo", "3mo", "6mo", "1y", "2y", "5y", "10y", "ytd", "max"] | TickerPeriod', got 'Literal["am"]' instead `

My first PR does not include this "Literal" string constraint. I'll wait for your feedback to update the original PR.

rickturner2001:
@ValueRaider added Literal type annotations and removed the enums. You should now be getting warnings when passing illegal options as interval or period.

ValueRaider:
Working now (was some PyCharm newbie user error).

rickturner2001:
It seems like literals are a good way to generate a type warning in this case. Honestly, I only used enums because I am used to working with typescript. I have a reason to believe that enums are not as appealing in python 😐

ValueRaider:
Can you rebase to `dev` and squash commits.

rickturner2001:
@ValueRaider  Pushed changes to dev. see #1316

ValueRaider:
? That pull request is completely different. You don't want this?

Issue #1295, opened 2023-01-13
Possible to get earnings for a specific date from https://finance.yahoo.com/calendar/earnings


sonso-1:
This function is currently available in yahoo_fin but that project is abandoned and broken due to recent changes in yahoo's encryption.  Looking to switch to from yahoo_fin to yfinance since this project is actively fixing bugs and keeping up with yahoo's changes.

Here's the function from yahoo_fin (https://github.com/atreadw1492/yahoo_fin), which is currently broken due to Yahoo's encryption changes:

def get_earnings_for_date(date, offset = 0, count = 1):

    '''Inputs: @date
       Returns a dictionary of stock tickers with earnings expected on the
       input date.  The dictionary contains the expected EPS values for each
       stock if available.'''
    
    base_earnings_url = 'https://finance.yahoo.com/calendar/earnings'
    
    if offset >= count:
        return []
    
    temp = pd.Timestamp(date)
    date = temp.strftime("%Y-%m-%d")

    dated_url = '{0}?day={1}&offset={2}&size={3}'.format(
        base_earnings_url, date, offset, 100)
    
    result = _parse_earnings_json(dated_url)
    
    stores = result['context']['dispatcher']['stores']
    
    earnings_count = stores['ScreenerCriteriaStore']['meta']['total']

    new_offset = offset + 100
    
    more_earnings = get_earnings_for_date(date, new_offset, earnings_count)
    
    current_earnings = stores['ScreenerResultsStore']['results']['rows']

    total_earnings = current_earnings + more_earnings

    return total_earnings

ValueRaider:
Start with `Ticker.earnings_dates`. Uses same Yahoo API but doesn't set the `day` argument. If you think `day` would improve then feel free to submit a pull request.

sonso-1:
I see how I could use Ticker.earnings_dates to get earnings dates for a specific ticker but what I'm looking for is to get a list of all tickers that have upcoming earnings dates.

ValueRaider:
Ah, that's not implemented.

mw66:
get a list of all tickers that have upcoming earnings dates.

Yes! this is what I'm looking for too. Is there a simple way to do this?  Thanks.

mw66:
e.g. 

https://finance.yahoo.com/calendar/earnings?from=2023-11-05&to=2023-11-11&day=2023-11-10

ValueRaider:
@mw66 Good suggestion. This works:
```
[...]

PR #1316, opened 2023-01-18
Get earnings by date 

rickturner2001:
Added the functionality to get earnings based on a given date #1295

### DIFF
--- a/yfinance/version.py
+++ b/yfinance/version.py
@@ -1,1 +1,1 @@
-version = "0.2.31b1"
+version = "0.2.31b2"


ValueRaider:
This looks awfully over-engineered (and in file). Can I suggest you look into test-driven development? Write tests first that specify behaviour, then implement.

rickturner2001:
> This looks awfully over-engineered (and in wrong file). Can I suggest you look into test-driven development? Write tests first that specify behaviour, then implement.

Would you mind being more specific as it relates to my changes being "over-engineered"?

I can definitely write some tests for this

ValueRaider:
`parse_date()` - this smells like over-engineering. Why not just copy how other functions handle user date e.g.:
```
def _get_earnings_by_date(date):
  timezone = self._get_ticker_tz(debug_mode=False, proxy=None, timeout=10)
  timestamp = utils._parse_user_dt(date, timezone)
  url = EARNINGS_URL + int(timestamp)
  ...
```

rickturner2001:
I don't think _get_ticker_tz is doing what you think it is doing. Furthermore, it can potentially send out a get request out to, url = "{}/v8/finance/chart/{}".format(self._base_url, self.ticker), which doesn't seem appropriate for our use case.

You said the implementation is **awfully over-engineered** but I honestly don't see it. The only thing that is maybe a bit unessential is the use of parse_date as a decorator. But then again, I think it works great in terms of code splitting. 

This takes out to my final point. It's important to be methodical when coding, especially when coding in a library. There's a difference between coding for yourself and adding features to an existing library. We need to split our code, of course, we can just create a function and put everything inside there, and call it a day. But that's neither good nor readable or maintainable.

However, if you do have something specific in mind that you would like to change, of course, feel free to do so. I'll be writing the test cases later today

ValueRaider:
Ignore timezone for now, `utils._parse_user_dt` is the important part, [it already splits out date parsing](https://github.com/ranaroussi/yfinance/blob/0c037ddd128f3ce5dee79ceb8a8571e5000fcd30/yfinance/utils.py#L291).

rickturner2001:
@ValueRaider I updated with _parse_user_dt. It seems to be much better this way. Where would you move the code? I thought base might be more appropriate

rickturner2001:
@ValueRaider  Updated with ```_parse_user_dt```. it looks like it's working nicely now. The function was missing an exception for invalid datetimes. Where do you think is best to put this code?

yfinance/utils.py
```diff
@@ -282,35 +306,48 @@ def camel2title(strings: List[str], sep: str = ' ', acronyms: Optional[List[str]
...
+
         dt = int(dt.timestamp())
-    return dt
+        return dt
+
+    except ValueError as e:
+        # Invalid date format or non-existing date (2023-20-20 or 2023-11-31)
+
+        print(f"Invalid date {dt}. {e}")
+        return None
```
ValueRaider, line 342: Don't discard exceptions at low-level. If one occurs it should be left to raise. Then at the API level `_get_earnings_by_date`, you can decide whether to replace with a different message.

yfinance/utils.py
```diff
@@ -282,35 +306,48 @@ def camel2title(strings: List[str], sep: str = ' ', acronyms: Optional[List[str]
...
+            # Should already be epoch, test with conversion:
+            dt = _datetime.datetime.fromtimestamp(dt)
+            # Convert str/date -> datetime, set tzinfo=exchange, get timestamp:
+        elif isinstance(dt, str):
             dt = _datetime.datetime.strptime(str(dt), '%Y-%m-%d')
-        if isinstance(dt, _datetime.date) and not isinstance(dt, _datetime.datetime):
+        elif isinstance(dt, _datetime.date) and not isinstance(dt, _datetime.datetime):
             dt = _datetime.datetime.combine(dt, _datetime.time(0))
-        if isinstance(dt, _datetime.datetime) and dt.tzinfo is None:
+        elif isinstance(dt, _datetime.datetime) and dt.tzinfo is None:
```
ValueRaider, line 326: Are you absolutely **sure** the logic hasn't changed by switching `if` -> `elif`? I deliberately used `if` for each to behave like a data pipeline.

ValueRaider (commented):
questions

ValueRaider:
Most of the diff changes aren't related to this feature, can you fix that? Rebase to `dev`, then revert the splitting code over lines. Makes review much easier.

yfinance/utils.py
rickturner2001: Yes, the behavior did not change at all. It was just a matter of semantics really. All the branches are checking the instance of dt and it seems to be more appropriate to have them at the same lever. But then again the behavior did not change

rickturner2001:
> Most of the diff changes aren't related to this feature, can you fix that? Rebase to `dev`, then revert the splitting code over lines. Makes review much easier.

Yes will do. I guess my formatter kicked in. Although I would say that my prettier is set up to use pep8 as a standard guideline for python formatting. Maybe someone should consider formatting the code on a different PR?

yfinance/utils.py
ValueRaider: I tested, it does change. If user provides a string the timezone isn't set. Please be careful.

rickturner2001: I guess we could just default it to "UTC" or whatever you think is best. If we don't set a default then we are forced to pass a timezone exchange even though the expression does not use it

for example

```python3
if isinstance(dt, int):
        # Should already be epoch, test with conversion:
        _datetime.datetime.fromtimestamp(dt)
```

what do you think should be the default? also, if you authored the function, could you write some tests so that we can track breaking changes? if you didn't author the function, I would gladly write the test cases

ValueRaider: Short answer: no default.

Long answer:
Presumably the `dt` value you're passing into that function is from the user, so you need to provide a timezone anyway because it's usually ISO 8601 ("2023-01-01").

Assuming UTC was cause of user experience issues before my fix - missing a day or receiving a day outside of requested range. Assuming the exchange timezone fixed this, and just makes sense.

I've started branch `clarify-datetime-handling` for answering these questions.

sonso-1:
Looks like this is only retrieving the first page of results (1-100).  1/30/2023 has 3 pages for example (209).  So it would need to combine results from:
page 1:  https://finance.yahoo.com/calendar/earnings?from=2023-01-29&to=2023-02-04&day=2023-01-30
page 2:  https://finance.yahoo.com/calendar/earnings?from=2023-01-30&to=2023-02-04&day=2023-01-30&offset=200&size=100
page 3: https://finance.yahoo.com/calendar/earnings?from=2023-01-30&to=2023-02-04&day=2023-01-30&offset=200&size=100

I was hoping to simply modify the URL to:  https://finance.yahoo.com/calendar/earnings?day=2023-01-30&size=1000
 but yahoo doesn't seem to accept anything larger than 100.

ValueRaider:
@sonso-1 Correct, `size` is capped at 100. [See how `get_earnings_dates()` handles it](https://github.com/ranaroussi/yfinance/blob/b17ad32a47b1e0d27573db084ac349a9af4ad2e7/yfinance/base.py#L1625)

sonso-1:
Do you think this will be added in a future update?

ValueRaider:
@sonso-1 No idea, all work here is voluntary.

rickturner2001:
We could send the request with beautiful soup and check whether `next` is an active link or not. If it is then we scan the page to which the link is pointing. If you guys like the idea I can send a pr later today

ValueRaider:
@rickturner2001 `get_earnings_dates()` just keeps requesting the next page until Yahoo returns nothing, works fine.

WesNeu:
The name could be improved by inserting the word "calendar".
I'm looking forward to this functionality for the exact same reasons as sonso-1.  
I believe the industry standard terminology is that getting "earnings" would mean getting dollars per share info, whereas what we are seeking here is known as getting "earnings calendar" data.  Notice the "calendar" in the link sonso-1 provided.

ValueRaider:
@sonso-1 @WesNeu Do either of you want to help finish implementing this?

WesNeu:
Thanks for the invitation.  Unfortunately I'm really not the right person for this.  I taught myself just enough python about a year ago to implement this app that calls into yahoo_fin to get what I needed.  I'm not very good at python and I think this note I'm writing now is the 5th time I've logged into github.  I don't know enough.  If I was retired and had time I would be happy to learn, but instead I'm overwhelmed with worldly responsibilities.  If someone else can do this I would be very appreciative.

sonso-1:
Thanks but unfortunately I'm not a developer either.

Issue #1300, opened 2023-01-14
Reduce `info[]` use


ValueRaider:
`Ticker.info` contains a lot of data, but some of it is accessible other faster ways:
- recent price data - `Ticker.history()` or `download()`
- exchange, timezone, currency - `Ticker.get_history_metadata()`
- share count (#1133 will prioritise merge)
- market cap (= price * share count)

`info` is a relatively expensive request, and may be contributing to Yahoo encrypting data. So I propose deleting these items from `info`, with accompanying documentation / warnings etc for a smooth transition.

Open to contrary views.

cmjordan42:
Caveat: I have limited familiarity with the architecture of yfinance; I recently switched over to it from my own yfinance-ish library when the previous cryptography issues hit a month or so ago. I'm assuming that a lot of the architecture is similar to how I built it, as it was dictated by how Yahoo APIs offered its data.

If someone is grabbing current price, previous close, and a handful of other non-pricing data points from `info` , wouldn't this change require to now call `info` AND `history()` and result in more load on Yahoo?

It seems this change would only yield improvement if the only reason that the person was calling info was to only get pricing data.

ValueRaider:
Basically `history` and `download` are fast, everything else is slower - `info` is the worst at 6.5x slower. Also you can call `history` 100-1000 times in a loop before being temporarily blocked, but `info[]` much less (maybe 10-100 times?).

> the person was calling info was to only get pricing data.

I've seen this in that large project. For individual users I often see them wanting just marketCap.

cmjordan42:
Got it. But isn't the proposed change to actually remove it from info? So for those who are being deliberate and not wasteful, this will increase the load / number of calls? For example, I will now have to call `history()` every time I currently call `info` as my goal when I call `info` is to get the summary-level data on a single security which `info` provides, not to get deep history on pricing.

It seems like the real solution is that the people who are being wasteful just need to not be wasteful :) Unless I'm missing something, this change will tangibly increase server load for those who ARE being resource-conscious, so we'd be going in the wrong direction.

ValueRaider:
The number of calls is less important than the amount of work those calls triggers. For whatever reason, `info` is ~6.5x more costly to Yahoo than `history`.

Another example: the proposed new way for getting market cap, using `history` and `share count`, is 3x faster despite 2 calls instead of 1.

cmjordan42:
[...]

PR #1317, opened 2023-01-18
`Ticker.basic_info` - fast but minimal alternative to `info[]`

ValueRaider:
Draft implementation for #1300 so feedback wanted. See that Issue for discussion of why ...

`Ticker.basic_info` provides fast access to ~25 values of `Ticker.info[]`, e.g. last close price. To encourage users to switch, these keys are removed from `info`. If user is using affected keys, then `yfinance` will print migration instructions eg:

`dat.info["open"]`
> Price data removed from info. Use Ticker.basic_info or history() instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,9 @@ msft.capital_gains
 msft.shares
 msft.get_shares_full()
 
+# calculate market cap
+msft.market_cap
+
 # show financials:
 # - income statement
 msft.income_stmt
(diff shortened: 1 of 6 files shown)


YieldPilot:
If beta and dividendRate were moved over I could switch to basic_info as well

ValueRaider:
Moved over from where?

ValueRaider:
@YieldPilot `basic_info` doesn't move anything from `info`, it fetches/calculates data via different methods. If you know a different way to get beta or dividendYield then you are free to implement and submit a Pull Request.

PR #1335, opened 2023-01-25
Add new yahoo key

ifel:
They just added a new key. Added it to the list, commented exception, the thing works now:
https://gist.github.com/ifel/5cdbf537a79da8a6e37203b1a9c30e4f


ValueRaider:
Superstar

ValueRaider:
Can you look into the feasibility of fetching & parsing the JS file to extract keys? Just to assess how difficult implementation would be, no need to actually do it.

ValueRaider:
Thanks. I'll create an Issue for anyone to implement.

PR #1336, opened 2023-01-25
Hardcode decryption keys

### LINKED ISSUE #1329: Exception: yfinance failed to decrypt Yahoo data response with hardcoded keys, contact developers
I use Python 3.8
Installed yfinance 0.2.4
I'm pulling the info of a ticker using this code:
`info = yf.Ticker(symbol).info`
And I'm getting this error message:

> Exception("Yahoo has again changed data format, yfinance now unsure which key(s) is for decryption: '80226cfb77c7'-><class 'str'> , '80226cfb77c7'-><class 'str'> , '80226cfb77c7'-><class 'str'> , '80226cfb77c7'-><class 'str'> , '80226cfb77c7'-><class 'str'> , '80226cfb77c7'-><class 'str'> , '80226cfb77c7'-><class 'str'> , '80226cfb77c7'-><class 'str'> ,
And it continues with a ton of <class 'str'> like above.

Does someone know how to fix it?

**EDIT from @ValueRaider**
In case you're confused by title not matching this top post - I edited title to reflect the latest error message. Discussion below still very relevent.

ValueRaider:
Should fix #1329

Thanks @ifel #1335


paulmcq (changes requested):
yfinance/data.py, line 168: suggest
`if password is not None:`

ValueRaider:
@paulmcq I think there are more pressing matters right now than linting, can fix that later.

PR #1363, opened 2023-01-26
fix typo causing AttributeError

dendiz:


ValueRaider:
Resolved in #1362, I'm just waiting short while in case more Issues get raised.

PR #1368, opened 2023-01-28
`fast_info` usability improvements

ValueRaider:
`fast_info` usability improvements:

- Add support camelCase dict keys, and officially support these, but secretly continue to also support snake_case keys
- Add `dict` functions `values()` and `items()`

### DIFF
--- a/tests/ticker.py
+++ b/tests/ticker.py
@@ -722,13 +722,16 @@ class TestTickerInfo(unittest.TestCase):
         key_rename_map["three_month_average_volume"] = "averageVolume"
 
         key_rename_map["market_cap"] = "marketCap"
-        key_rename_map["shares"] = "floatShares"
+        key_rename_map["shares"] = "sharesOutstanding"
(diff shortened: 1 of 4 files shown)


sanyearng:
I wasn't aware of the following dependency, but am posting here as FYI, as it surprised me:

I could not get Ticker.get_history_metadata() to function unless both a) Ticker.fast_info was first run; and additionally Ticker.fast_info.items() (for 0.2.9)

Delete or ignore this comment if this is otherwise obvious or intended.

ValueRaider:
Did you should see this messag?

> RuntimeError: Metadata was never retrieved so far, call history() to retrieve it

sanyearng:
> Did you should see this message?
> 
> > RuntimeError: Metadata was never retrieved so far, call history() to retrieve it
> 
> It doesn't make sense to request metadata before fetching the real data (e.g. `history`). If you think something different should happen, request in a new Issue.

Perfect.  My mistake.

PR #1385, opened 2023-01-31
Restore original download() timezone handling

ValueRaider:
`download()` original behaviour for handling timezones of different exchanges depended on interval:
- intraday = keep timezones
- interday = ignore timezones

This was lost with #1048 but no one noticed for long time. It treated all intervals same way - never ignore. Eventually users noticed and requested always ignore #1103, then never ignore again #1283.

@DE0CH

### DIFF
--- a/yfinance/multi.py
+++ b/yfinance/multi.py
@@ -29,7 +29,7 @@ from . import Ticker, utils
 from . import shared
 
 
-def download(tickers, start=None, end=None, actions=False, threads=True, ignore_tz=False,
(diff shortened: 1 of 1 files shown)


DE0CH:
Sorry I forgot about this for a while. But thanks for the change. I updated my code to use ignore_tz=False so I’m ok with it.Best wishes,DeyaoOn 7 Feb 2023, at 13:16, ValueRaider ***@***.***> wrote:﻿
Merged #1385 into dev.

PR #1397, opened 2023-02-05
Catch TypeError Exception

Matt-Seath:
Addresses recent issue where calling Ticker.info would occasionally result in a TypeError Exception at line 287.


Matt-Seath:
@ValueRaider Thanks for your assistance, I'm not very experienced with GitHub as this is my first PR. Hopefully this better aligns with contributor guidelines.

PR #1411, opened 2023-02-09
Fix `history_metadata` accesses for unusual symbols

sdeibel:
Fix format_history_metadata when firstTradeDate is None, as is the case for QCSTIX and probably others.  Submitting this on main since it seemed appropriate for that based on the instructions in https://github.com/ranaroussi/yfinance/discussions/1084 -- I can redo it if that is not what you want.

### DIFF
--- a/tests/ticker.py
+++ b/tests/ticker.py
@@ -52,12 +52,16 @@ class TestTicker(unittest.TestCase):
     def test_badTicker(self):
         # Check yfinance doesn't die when ticker delisted
 
-        tkr = "AM2Z.TA"
+        tkr = "DJI"  # typo of "^DJI"
         dat = yf.Ticker(tkr, session=self.session)
         dat.history(period="1wk")
         dat.history(start="2022-01-01")
         dat.history(start="2022-01-01", end="2022-03-01")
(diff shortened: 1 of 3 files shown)


ValueRaider:
`main` is fine for this. I've updated unit tests (they should have detected problem), which found some other issues. I don't have time to use, let me know if anything broken.

sdeibel:
I don't think there are new problems but found that requirements.txt was missing requests-cache and scipy so I updated this pull request.  Actually, I meant to make a new separate pull request but that didn't got as planned...

ValueRaider:
Those two modules aren't *necessary* so deliberately missing.
- `requests_cache` - if user smart enough to want to cache, they're smart enough to install
- `scipy` - my sense is usage of price repair is rare (where `scipy` used)

sdeibel:
There were needed to run the unit tests, which is what motivated me to add them.  I suppose another idea would be to add a requirements-test.txt but it's not a big deal since it's pretty obvious what's needed when trying to run the tests and only two things had to be installed.

ValueRaider:
I deleted that commit

sdeibel:
Great, thanks.

PR #1423, opened 2023-02-16
No print

### LINKED ISSUE #1378: Library using print rather than logger
The library is using print statements to warn the user, e.g.:

https://github.com/ranaroussi/yfinance/blob/ac6e047f0dc2e6ec446f95359bc55684f08370f4/yfinance/scrapers/quote.py#L129

This should be replaced by standard logs (logging.Logger) to allow the library client to process the logs correctly (e.g., redirect them to files, stdout, stderr, live notification streams, ...).

flaviovs:
This resolves  #1378.
Here's what this PR does:
- Replace ordinary debug/info/warnings/error messages using `print()` with proper `logger.XXXX()` calls.
- Replace deprecation warnings issued with `print()` with proper Python warnings using [the _warnings_ module](https://docs.python.org/3/library/warnings.html).

NOTE: Some `print()` calls were guarded by `if debug:` tests. This PR does **not** change those, because debug is toggled on/off inside the library sometimes, so there's some logic involved that I am not fully aware of. Ideally this internal debug flag should be removed altogether, and the logic replaced with `logger.debug()` calls. I can provide a PR for that if desired.

### DIFF
--- a/test_yfinance.py
+++ b/test_yfinance.py
@@ -15,6 +15,9 @@ Sanity check for most common library uses all working
 
 import yfinance as yf
 import unittest
+import logging
+
+logging.basicConfig(level=logging.DEBUG)
 
 symbols = ['MSFT', 'IWO', 'VFINX', '^GSPC', 'BTC-USD']
 tickers = [yf.Ticker(symbol) for symbol in symbols]
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -22,6 +22,7 @@
 from __future__ import print_function
 
 import logging
+import warnings
 import time as _time
 import datetime as _datetime
 import dateutil as _dateutil
@@ -1487,7 +1488,7 @@ class TickerBase:
 
     @property
     def basic_info(self):
-        print("WARNING: 'Ticker.basic_info' is renamed to 'Ticker.fast_info', hopefully purpose is clearer")
+        warnings.warn("'Ticker.basic_info' is renamed to 'Ticker.fast_info', hopefully purpose is clearer", DeprecationWarning)
         return self.fast_info
 
     def get_sustainability(self, proxy=None, as_dict=False):
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -595,6 +595,9 @@ class TickerBase:
                 exceptions instead of printing to console.
         """
 
+        if raise_errors:
+            debug = True
+
         if start or period is None or period.lower() == "max":
             # Check can get TZ. Fail => probably delisted
             tz = self._get_ticker_tz(debug, proxy, timeout)
@@ -607,7 +610,9 @@ class TickerBase:
                     if raise_errors:
                         raise Exception('%s: %s' % (self.ticker, err_msg))
                     else:
-                        print('- %s: %s' % (self.ticker, err_msg))
+                        logger.error('- %s: %s' % (self.ticker, err_msg))
+                else:
+                    logger.debug('- %s: %s' % (self.ticker, err_msg))
                 return utils.empty_df()
 
             if end is None:
@@ -703,7 +708,9 @@ class TickerBase:
                 if raise_errors:
                     raise Exception('%s: %s' % (self.ticker, err_msg))
                 else:
-                    print('%s: %s' % (self.ticker, err_msg))
+                    logger.error('%s: %s' % (self.ticker, err_msg))
+            else:
+                logger.debug('%s: %s' % (self.ticker, err_msg))
             return utils.empty_df()
         
         # parse quotes
@@ -721,7 +728,9 @@ class TickerBase:
                 if raise_errors:
                     raise Exception('%s: %s' % (self.ticker, err_msg))
                 else:
-                    print('%s: %s' % (self.ticker, err_msg))
+                    logger.error('%s: %s' % (self.ticker, err_msg))
+            else:
+                logger.debug('%s: %s' % (self.ticker, err_msg))
             return shared._DFS[self.ticker]
 
         # 2) fix weired bug with Yahoo! - returning 60m for 30m bars
@@ -842,7 +851,9 @@ class TickerBase:
                 if raise_errors:
                     raise Exception('%s: %s' % (self.ticker, err_msg))
                 else:
-                    print('%s: %s' % (self.ticker, err_msg))
+                    logger.error('%s: %s' % (self.ticker, err_msg))
+            else:
+                logger.debug('%s: %s' % (self.ticker, err_msg))
 
         if rounding:
             df = _np.round(df, data[
@@ -876,9 +887,6 @@ class TickerBase:
 
         # Reconstruct values in df using finer-grained price data. Delimiter marks what to reconstruct
 
-        debug = False
-        # debug = True
-
         if interval[1:] in ['d', 'wk', 'mo']:
             # Interday data always includes pre & post
             prepost = True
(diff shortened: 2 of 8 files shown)


asafravid (commented):
You sometimes use logger.warning() and sometimes warning.warn() - can you elaborate why?

flaviovs:
@asafravid, `logger.warning()` (from the _logging_ module) and `warnings.warn()` serve two different purposes.

Logging and `logger.*()` calls are geared towards end users, who can configure the logging system in the app to handle log messages issued by libraries in several ways. For example, they can set warnings and errors to go to stderr, info and debug to stdout, also log warnings or above to a file, etc.). The app can even silence log messages from specific libraries/levels.

Conceptually, `logger.*()` is used to tell **the user** what is going on **in the library** at runtime. In the context of this PR, `logger.warning()` is just logging messages at _warning_ level.

On the other hand, `warnings.warn()` (from the [_warnings_ module](https://docs.python.org/3/library/warnings.html)) is geared towards developers. These warnings are totally unrelated to the logging system. There's no way to redirect warnings, and to silence them, other than the developer [explicitly filtering them out](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) (which is an indication that they know what they're doing — at least in theory :wink:).

Conceptually `warnings.warn()` is used to tell **the developer** what is going on with **the code**. 

 In this PR, for example, I used `warnings.warn()`  with [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) to replace `print()` calls warning about deprecated calls in yfinance. This is the precise use case for `warnings.warn(..., DeprecationWarning)`.

Let me know if that does not answer your question.

asafravid:
> @asafravid, `logger.warning()` (from the _logging_ module) and `warnings.warn()` serve two different purposes.
> 
> Logging and `logger.*()` calls are geared towards end users, who can configure the logging system in the app to handle log messages issued by libraries in several ways. For example, they can set warnings and errors to go to stderr, info and debug to stdout, also log warnings or above to a file, etc.). The app can even silence log messages from specific libraries/levels.
> 
> Conceptually, `logger.*()` is used to tell **the user** what is going on **in the library** at runtime. In the context of this PR, `logger.warning()` is just logging messages at _warning_ level.
> 
> On the other hand, `warnings.warn()` (from the [_warnings_ module](https://docs.python.org/3/library/warnings.html)) is geared towards developers. These warnings are totally unrelated to the logging system. There's no way to redirect warnings, and to silence them, other than the developer [explicitly filtering them out](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) (which is an indication that they know what they're doing — at least in theory 😉).
> 
> Conceptually `warnings.warn()` is used to tell **the developer** what is going on with **the code**.
> 
> In this PR, for example, I used `warnings.warn()` with [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) to replace `print()` calls warning about deprecated calls in yfinance. This is the precise use case for `warnings.warn(..., DeprecationWarning)`.
> 
> Let me know if that does not answer your question.

Thanks for the details

asafravid (approved):
LGTM

ValueRaider:
I haven't looked through code yet, just want to say: rebase to `dev` then change target to `dev`. #1084

ValueRaider:
This doesn't work for me. @asafravid did you actually run this?

I put this at top of `base.py`:
```
logger.setLevel(logging.DEBUG)
```

Nothing prints.

asafravid:
@ValueRaider Try this: https://stackoverflow.com/questions/41796934/turn-on-debug-logging-in-python

ValueRaider:
@asafravid That SO answer is fine if I want to print *everything*, but not if I want to reduce verbosity to e.g. CRITICAL. As I explained already. Have you successfully restricted `logger` output to above DEBUG? [Logging levels](https://docs.python.org/3/library/logging.html#logging-levels)

flaviovs:
> This doesn't work for me. @asafravid did you actually run this?
> 
> I put this at top of `base.py`:
> 
> ```
> logger.setLevel(logging.DEBUG)
> ```
> 
> ... then nothing prints.
> 
> If instead I put this at top, everything prints when only CRITICAL should print (according to their documentation)
> 
> ```
> logging.basicConfig(level=logging.CRITICAL)
> ```

@ValueRaider if you put `logger.setLevel(logging.DEBUG)` at the top of `base.py` then you'll be changing only the _yfinance.base_ logger.

However, I strongly advise against adding `setLevel()` calls in modules, because this hinder the possibility of app users to control debug levels on the app using yfinance.

Notice that the distinction between module and app calls is important in the _logging_ module context. I'm not sure what your use case is, but if you're just trying to enable DEBUG+ in _yfinance_ **only**, while perhaps seeing WARNING+ from other libraries too (e.g. urllib), the you can do this on your script:

```python
import logging

# Basic configuration to log everything at WARNING, ERROR and CRITICAL
# to stderr. This will print WARNING+ messages from **all** modules (not
# only yfinance).
logging.basicConfig(level=logging.WARNING)

# Now configure yfinance logger to display DEBUG and above messages.
logger = logging.getLogger('yfinance')
logger.propagate = False 
handler = logging.StreamHandler()
handler.setFormatter(logging.Formatter('%(levelname)s:%(name)s:%(message)s'))
logger.addHandler(handler)
logger.setLevel(logging.DEBUG)
```

A simplified version of the above that logs DEBUG+ from all modules  was added already to `test_finance.py`.

All that said, I want to remind you guys that **I did not replace debugging `print()`s with `logger.debug()` calls**. The reason as I stated in the PR is because the `print()`s are guarded by `if debug: ` logic, and I didn't want to go out of scope in this PR by changing programming logic.

ValueRaider:
Thanks @flaviovs , I think I understand now.

> I did not replace debugging `print()`s with `logger.debug()` calls

That's fine but I think these are in scope so I want to replace these too, `logging` should be able to replace that logic. I can handle this.

flaviovs:
> And reminder: you need to rebase to dev. Should solve that merge conflict below, don't waste time on it.

@ValueRaider, I just did it (no conflicts).

Let me know if you need any more help with the PR.

ValueRaider:
@flaviovs I have a request: does `logging`enable printing a specific message only once? I.e. it remembers? This will be useful for warning about deprecated arguments, most users run `yfinance` in a loop and don't need to see same message many times.

asafravid:
@ValueRaider possibly we can insert this statefulness per logging message within the yfinance code (if not possible via the logging infrastructure)
Otherwise there’s a hack which can be used as here: https://stackoverflow.com/questions/31953272/logging-print-message-only-once

flaviovs:
> @flaviovs I have a request: does `logging`enable printing a specific message only once? I.e. it remembers? This will be useful for warning about deprecated arguments, most users run `yfinance` in a loop and don't need to see same message many times.

Hi @ValueRaider, _logging_ does not "remember". I think you can find (or develop) a log handler that does that but, as far as I know, it does not that by default.

However, I think that for this particular use case (deprecation), you want  `warnings.warn(MESSAGE, DeprecationWarning)` .  This is the correct tool for the job and it indeed does suppress repeated messages. From https://docs.python.org/3/library/warnings.html:

> Repetitions of a particular warning for the same source location are typically suppressed.

ValueRaider:
@flaviovs Thanks for suggestion but `warnings` "once" filter doesn't work. Internet filled with 'not working' threads.

@asafravid LRU cache idea looks good.

flaviovs:
> @flaviovs Thanks for suggestion but `warnings` repeat suppression doesn't work, even if using "once" filter. I'm not alone in having trouble.

That's interesting. I've never depended on this behaviour so actually never tested it. Thanks for bringing me awareness to this.

> @asafravid LRU cache idea works perfect.

I'm glad to hear that you guys found a solution.

Let me know if I can help with anything else.

Issue #1436, opened 2023-02-28
problem with 30 mins Time frame for ^NSEI (national stock exchange of india)


AbhishekSRaut:
There is problem with 30 mins time frame with indian stocks.
the 30 mins time interval data is showing wrong open and close.
for example:
in nifty, the first 30 mins candle is starts from 9:15 and, ends at 9:30.
and after that it is continueing from 9:30 to 10:00.
but, because of this, our all strategies getting destroied, as normal candle starts at 9:15, and ends at 9:45.
the last candle of nifty should be of 15 mins, i.e. starts at 15:15, and ends at 15:30.
Please solve this issue as soon as possible, because due to this, our all strategies / scanner gets destroied.
Thank you.
regards,
Abhishek Raut.

ValueRaider:
Only because you say this:

> our all strategies / scanner gets destroied

Try `yfinance`, price history fetches work perfect. I know it has other issues but they are separate.

ValueRaider:
This is because `yfinance` resamples 30m data from 15m, because apparently the 30m data Yahoo returned was bad years ago, but the resampling is lazy. But today Yahoo's data looks fine, so potentially the resampling is not needed.

If someone wants to solve, do this:
- resample properly, by offsetting the intervals times by market open time (get this from `history_metadata`)
- move this resampling into a unit test, and verify that the resampling is now not necessary (test on variety of stock exchanges)

ValueRaider:
Regarding 60m - I can't reproduce, all the hourly intervals start/end at hh:15

AbhishekSRaut:
> This is because yfinance resamples 30m data from 15m, but naive resampling assumes alignment at hh:00. Why resample? Because apparently the requesting Yahoo for 30m data returned 60m, but this was years ago. Today Yahoo's data looks fine, so potentially the resampling is not needed.
If someone wants to solve, do this:
• resample properly, by offsetting the intervals times by market open time (get this from history_metadata)
• move this resampling into a unit test, and verify that the resampling is now not necessary (test on variety of stock exchanges)

AbhishekSRaut:
I do not understand your solution.
can you please provide python code to solve this?

ValueRaider:
> I do not understand your solution.

That's ok. Someone else will, someone familiar with `yfinance` code.

AbhishekSRaut:
> That's ok. Someone else will, someone familiar with yfinance code.

It means, We have to wait for the update of yfinance.
[...]

PR #1447, opened 2023-03-09
Don't fix_Yahoo_returning_prepost_unrequested on Indian exhanges and 15m interval

ivan23kor:
Addresses #1436.


ivan23kor:
@ValueRaider, the only problem I have is that `fix_Yahoo_returning_prepost_unrequested` won't be run for `15m` intervals either, because we Ticker._history() overwrites user-specified interval of `30m` with `15m`.

ValueRaider:
> the order of `fix_Yahoo_returning_prepost_unrequested` and downsampling doesn't matter in this case, as trading start date is 9:15 but the downsampling interval is `30m`.

I don't understand, please elaborate.

ivan23kor:
For `30m` interval (`15m`, actually), the timestamps in the data just after fetching by requests look like this: `8:45, 9:00, 9:15, 9:30, 9:45, 10:00 ...`

If you downsample first, you get `9:00, 9:30, 10:00, ...`, then `fix_Yahoo_returning_prepost_unrequested` will remove 9:00.

If you `fix_Yahoo_returning_prepost_unrequested` first, you get `9:15, 9:30, 9:45, 10:00, ...`, then downsampling removes 9:15.

In either case, 9:00 will be removed.

ivan23kor:
@ValueRaider, changed the interval check to the user-provided interval, instead of the interval passed to `requests`, so now `fix_Yahoo_returning_prepost_unrequested` will be skipped for NSI and BSE only for the true `30m` interval.

ValueRaider:
> In either case, 9:00 will be removed.

Wrong. I am running code now, if I reorder then `yfinance` returns interval 09:00 -> 09:30.

ivan23kor:
You are right, that works!
Pushed reordering.

ValueRaider:
Good, that fixes the first interval. But I'm still not convinced aligning to HH:00 is the right choice. You said:

> Yahoo Finance's frontend returns HH:00 timestamps for some reason and yfinance should follow that behaviour.

This is a terrible reason - Yahoo is US-centric and doesn't trade, so their reason is [overruled by an Indian trader](https://github.com/ranaroussi/yfinance/issues/1436#issuecomment-1459468997). I don't trade intraday so don't know a good authority on this.

ivan23kor:
Tradingview for ^NSEI [starts with 9:15](https://www.tradingview.com/symbols/NSE-NIFTY/)

AbhishekSRaut:
agreed with @ValueRaider other factors, like regional trading practices, should be considered when deciding on the best approach.

> Tradingview for ^NSEI starts with 9:15. That's a good enough argument for me to shift the beginning of 30m data for NSE and BSE to HH:15.

this officialy mention [here.](https://www.nseindia.com/resources/exchange-communication-holidays#:~:text=Normal%20market%20%2F%20Exercise%20market%20open,Collateral%20value%3A%2016%3A15%20hrs)
you can read:

> Normal / Limited Physical Market Open: 09:15 hrs
> Normal / Limited Physical Market Close: 15:30 hrs

thanks @ivan23kor and @ValueRaider to understand this and fix it.

ivan23kor:
@AbhishekSRaut do you want the intervals to be `[9:15AM, 3:15PM]` or `[9:15AM, 3:30PM]`?

@ValueRaider, I [added](5f379c36d8a03e857553965c9cc75b40e364e236) `[9:15AM, 3:15PM]` for `(NSI, BSE)` and kept `[9:30AM, 3:30PM]` for other exchanges.

ValueRaider:
Where in the code does 3:15pm vs 3:30pm matter?

AbhishekSRaut:
> do you want the intervals to be [9:15AM, 3:15PM] or [9:15AM, 3:30PM]?

if you talking about starting time of each candle, then its surely starts from 9:15 AM, and last is 3:15PM
3:30 PM is closing time i.e. the "last candle" which starts at 3:15 PM, will end at 3:30 PM.

PR #1460, opened 2023-03-22
Add failback for decryption error in info interface

qianyun210603:
Due to the persistent failure of `info` property of `Ticker`s due to decryption error of yahoo finance web pages, I added a failback to get some information from unofficial api `https://query1.finance.yahoo.com/v7/finance/quote`. which can get the ticker short/long name, listed date, exchange, timezone etc.

Also did some refactoring of `base.py` and `data.py`:
- merge duplicate imports;
- remove unused local variables
- fix some (potential) reference before defined error
- reformatting using black.


qianyun210603:
@ValueRaider This makes sense. I removed all reformatting and created a new PR #1461 for easy review. Thanks.

Issue #1399, opened 2023-02-05
Ticker.get_isin method does not work with proxy 


vidalmarco:
I spent some time on debugging the get_isin method after having applied `hotfix/proxy` (see also #1387)

After some failures (always getting HTTPS Tunnel 407 failure) I've tried to test it via curl.

**First** (as generated by get_isin)
`curl -x 10.xx.xxx.xx:pppp https://markets.businessinsider.com/ajax/SearchController_Suggest?query=IBM`
curl: (56) Received HTTP code 407 from proxy after CONNECT

**Second** (specify a user)
`curl -x 10.xx.xxx.xx:pppp https://markets.businessinsider.com/ajax/SearchController_Suggest?query=IBM -U <my_username>`
Enter proxy password for user '<my_username>':
curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092012) - The revocation function was unable to check revocation for the certificate.

**Third** (specify a user and a ssl cert verification check = False)
`curl -x 10.xx.xxx.xx:pppp https://markets.businessinsider.com/ajax/SearchController_Suggest?query=IBM -U <my_username> --ssl-no-revoke`

Enter proxy password for user '<my_username>':
[...]

PR #1371, opened 2023-01-28
Fix proxy arg passthrough

ValueRaider:
`proxy` argument was broken in recent versions, this should fix it. 

Fixes #998 #1282 #1386, and maybe #1387 (were they using proxy?)

### Status

`Ticker.isin()` doesn't work with proxy, can someone with a proxy help investigate? #1399

PR #1472, opened 2023-04-01
Omit of params of the get_fn

Qiqipixel:
I've tried a lot of time to change my params in method .history, but the result is always"No data found for this date range, symbol may be delisted". I don't know if the reason is that I use a proxy because I'm not in America. But when I add param: {proxy and user_agent_headers}, the problem is soluted. Hope this will be helpful.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -644,7 +644,8 @@ class TickerBase:
 
         # Getting data from json
         url = "{}/v8/finance/chart/{}".format(self._base_url, self.ticker)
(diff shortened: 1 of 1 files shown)


ValueRaider:
Thanks for contribution. Previous PR #1371 should fix this, but not merged because need a proxy user to test - can you test? Specifically help with this issue: #1399

ricardoprins:
@ValueRaider should this be closed because of #1371?

PR #1480, opened 2023-04-04
'info' fetch now gets same data as scrape 

### LINKED ISSUE #1471: 0.2.14 `yf.Ticker("<ticker").info` No Longer has Sector, Industry, Description Data as in 0.2.12
In version 0.2.12, running the code 

```
import yfinance as yf

msft = yf.Ticker("MSFT")

# get all stock info (slow)
msft.info
```

yielded sector, industry, and description, taken from the profile page, as seen below. However, this is not the case on 0.2.14. If this is intentional and the information is moved somewhere else, it is not well documented. Else, this is a bug.

image

kennykos:
Closes #1471 
* Changed base url to "https://query2.finance.yahoo.com/v10/finance/quoteSummary"
* instead of just getting the quote from `"https://query1.finance.yahoo.com/v7/finance/quote"`, we now get
	* ``` items = ['summaryProfile', 'financialData', 'quoteType', 'defaultKeyStatistics', 'assetProfile', 'summaryDetail'] ``` which is the same as in the srape function


ValueRaider:
More consideration needed of data layout & formatting. Can you improve it? OK if you cannot, I can merge into a branch and someone else can improve.

kennykos:
> More consideration needed of data layout & formatting. Can you improve it? OK if you cannot, I can merge into a branch and someone else can improve.

My most recent commit turns the response into a linear dictionary structure, which can now be formatted in the format parser that is currently in the `_fectch()` function. Is there any other formatting required?

ValueRaider:
Better but looks like some data was lost, e.g. "companyOfficers" was a populated list.

And should check for duplicate keys across the sub-dicts and handle smartly e.g. "companyOfficers" (might be why the list disappeared).

kennykos:
> Better but looks like some data was lost, e.g. "companyOfficers" was a populated list.
> 
> And should check for duplicate keys across the sub-dicts and handle smartly e.g. "companyOfficers" (might be why the list disappeared).

I addressed this issue by only adding non empty values to the info dict. This way, if a value is overwritten, it will be the same (assuming yahoo's info is well defined, which if not, why even bother), and in the case were a key had an empty value, such as for company officers, nothing will be overwritten.

ValueRaider:
It's still missing from `info`.

ValueRaider:
Looks good to me now. I'll give others a chance to comment before merging.

ValueRaider:
I've squashed all the commits together.

Issue #1531, opened 2023-05-18
Yahoo sometimes fails to apply stock split to past prices, YF could repair


ValueRaider:
Symbol CNE.L had a stock split recently, but clearly most prices before split have not been updated (except May 9):

![image](https://github.com/ValueRaider/yfinance-cache/assets/96923577/c81ccc9f-5393-497b-9fd2-357a6991d0fd)

Because the problem is obvious, implies a technical solution is possible.

~Will need to recalculate `Adj Close` - dividend is split-adjusted but Yahoo using pre-split price to calculate. May only be necessary when dividend on same day as split. Adjustment = 1 - dividend / (close day before).~ <- Not possible because impossible to know if dividend is split-adjusted or not.

Implement under the `repair` flag so default is disabled. Search in `base.py`, should be obvious.

Watch out for multiple stock splits, e.g. last week and 2 years ago. Probably the older split is properly processed, only the recent one is bad.

ValueRaider:
I've drafted a fix but I want more examples for testing.

lcrmorin:
Noticed the same problem during the M6 competition. I was mainly identifying problems from mostly symetrical log returns if that helps. I will report if I see another problem and will be able to help to check the solution.

ValueRaider:
I've since found more examples for testing, and I feel confident in the algorithm, so really all it needs now is people to use and check.

PR #1543, opened 2023-05-31
Price repair update: fix Yahoo messing up dividend and split adjustments

ValueRaider:
- Dividends

Rarely, Yahoo sets `Adj Close` = `Close` despite a dividend meaning they should be different. Easy to fix.

- Stock splits

Sometimes, Yahoo fails to apply a stock split to pre-split price data. And rarely, they apply the adjustment twice! See Issue #1531. Harder to fix properly so this has a unit test.

This is implemented under the `repair` flag so nothing happens unless user requests it.

**Related changes**

- Yahoo often messes up price data on stock split day - all equal or simply missing. Incredibly unlikely that price didn't move/trade on stock split day so should be repaired.
- Use split-repair logic to fix some 100x errors.
- When repairing 1d 'Adj Close' (with 1h data), extract correct div-adjustment from the good 1d data.
- Limit price reconstruction recursion depth to 2.


ValueRaider:
@lcrmorin FYI

PR #1545, opened 2023-06-03
fix: Readme cache-ratelimit. Limiter parenthesis was never closed

SnoozeFreddo:
The example in the docs will not work out of the box due to a syntax error.


ValueRaider:
Should README explain how to rate-limit without a cache? It's just this:
``` Python
from requests_ratelimiter import LimiterSession
session = LimiterSession(limiter=limiter)
```

SnoozeFreddo:
First i think it's more important to fix the readme, because it's not working :P

Issue #1534, opened 2023-05-21
Passing request session to yf.download


jkant:
Currently it is only possible to pass the request session to ticker objects. This means it is not possible for yf.download to use rate limiting and avoid pulling down corrupted data.

bveber:
I added a PR that should fix this. 

I have it working locally, but I'm not seeing the expected behavior from the sample code with the custom `CachedLimiterSession`. I was expecting for a cache hit to bypass the rate-limiter, but in my testing I don't see any speedup for cache hits. I have verified the caching works without the rate-limiter (using only requests_cache.CacheMixin) but when I include the LimiterMixin from requests_ratelimiter my response time is always bottlenecked by the limiter. 

My problem is obviously external to this package but I'm curious if anyone has seen the same behavior.

PR #1547, opened 2023-06-06
add session to download

bveber:
Allows user to pass a session object to the download method.

Fixes [#1534](https://github.com/ranaroussi/yfinance/issues/1534)

### DIFF
--- a/yfinance/multi.py
+++ b/yfinance/multi.py
@@ -33,7 +33,7 @@ from . import shared
 def download(tickers, start=None, end=None, actions=False, threads=True, ignore_tz=None,
              group_by='column', auto_adjust=False, back_adjust=False, repair=False, keepna=False,
              progress=True, period="max", show_errors=None, interval="1d", prepost=False,
-             proxy=None, rounding=False, timeout=10):
+             proxy=None, rounding=False, timeout=10, session=None):
     """Download yahoo tickers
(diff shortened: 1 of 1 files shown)


bveber:
I have it working locally, but I'm not seeing the expected behavior from the sample code with the custom CachedLimiterSession. I was expecting for a cache hit to bypass the rate-limiter, but in my testing I don't see any speedup for cache hits. I have verified the caching works without the rate-limiter (using only requests_cache.CacheMixin) but when I include the LimiterMixin from requests_ratelimiter my response time is always bottlenecked by the limiter.

My problem is obviously external to this package but I'm curious if you've seen the same behavior.

ValueRaider:
Sounds like `logging` should be able to help with that. Look how `yfinance` already uses logging, and have it log the url + params to "debug". If url is changing then it will never hit cache. Can combine with this PR.

bveber:
I think I resolved my caching issue. In my download method calls I was using the default `end` parameter which sets that value to the current time which bumps against this [conditional check](https://github.com/ranaroussi/yfinance/blob/d9bfd291131ca70121c0fd0b027512a268b2fc8b/yfinance/base.py#LL201C28-L201C28). Setting the end date to some value more than 30 minutes ago seems to have fixed it.

PR #1562, opened 2023-06-15
Fix logging behaviour

ValueRaider:
Various changes to improve alignment of logging output, particularly DEBUG logs.

`MultiLineFormatter` improves appearance of multi-line messages in any logging mode. Adds leading indent to all lines below first, to align with the first line which is indented by %levelname%.

Only add stream handler if no root logger, this fixes message being double-printed.

Then specifically for DEBUG logs:

1 - Whitespace padding is added to end of %levelname% string up to 8 characters. This resolves different level strings e.g. INFO & DEBUG creating different indents. But this only automatically applies to first line of message - for multi-line messages, `MultiLineFormatter` extracts amount of padding and applies to all other lines.

2 - Add leading indent proportional to function depth, because DEBUG generates a lot of messages particularly when repairing price data - same function can be called recursively. Implemented in `IndentLoggerAdapter`, a simple wrapper around the logger object.

3 - `log_indent_decorator` inserts 'Entering/Exiting %function%' messages, helps organise.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -38,7 +38,7 @@ Yahoo Finance free service has rate-limiting depending on request type - roughly
 **Delete these instructions** and replace with your bug report, providing the following as best you can:
 
 - Simple code that reproduces your problem, that we can copy-paste-run.
-- Run code with [debug logging enabled](https://github.com/ranaroussi/yfinance/tree/dev#logging) and post the full output.
+- Run code with [debug logging enabled](https://github.com/ranaroussi/yfinance#logging) and post the full output.
 - If you think `yfinance` returning bad data, give us proof.
 - `yfinance` version and Python version.
 - Operating system type.
--- a/README.md
+++ b/README.md
@@ -188,14 +188,7 @@ Review the [Wiki](https://github.com/ranaroussi/yfinance/wiki) for more options 
 
 ### Logging
 
-`yfinance` now uses the `logging` module. To control the detail of printed messages you simply change the level:
-```
-import logging
-logger = logging.getLogger('yfinance')
-logger.setLevel(logging.ERROR)  # default: only print errors
-logger.setLevel(logging.CRITICAL)  # disable printing
-logger.setLevel(logging.DEBUG)  # verbose: print errors & debug info
-```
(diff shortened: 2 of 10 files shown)


ValueRaider:
@flaviovs Curious to hear any opinions you have. GPT helped me a lot, but I'm still a noob with logging.

flaviovs:
Hi @ValueRaider - sorry for taking too long to give feedback. Been incredibly busy here...

I'm now using the newest version with logging enabled, and it is certainly working as expected. The only thing to think about is: my app started to issue double logging messages to stderr from _yfinance_, and I tracked the issue down to `utils.get_yf_logger()`, which adds a handler and formatter to the logging system. This is called on `base.py`, indirectly added imported by `__init__.py`, so effectively that function is always called when _yfinance_ is imported. This explains the double logging, but it is usually not a good thing because now applications need to add heuristics to prevent that specifically logger for interfering with their logging configuration.

Also, while browsing the code I saw some `logger.setLevel()` calls which would break the logging config done by any application anyway.

Usually logging *setup* calls (i.e. those that add handler, formatter, filters, set levels, etc.) should be done in application-space only, and not by library code. The reason is because libraries don't know how the app want's to do logging. I understand that when we want to debug _yfinance_ we need a way to enable debug for (and just for) _yfinance_, but in this case the setup for that should be done in application-space (perhaps in `tests/__init__.py` or similar).

I also struggled with logging at the beginning (no ChatGTP back them :wink:), but at some point I realized that if I just wanted to use logging in my apps, or make my packages logging-compatible, there's just a handful of things I need to make:

In my apps:
```python
import logging

# Here I configure logging for the entire app, plus ALL modules it imports, including
# modules imported by other modules I don't know about. I can just simply call
# basicConfig() here to setup a basic logging configuration, but note that this can be as
# complicated as reading the entire logging configuration from a .ini file
# (see https://docs.python.org/3/library/logging.config.html), and/or configure the
# framework to log to a SQL database, direct some modules to a file, suppress
# uninteresting messages, reformat messages from specific packages, etc.
logging.basicConfig()
```

In my packages, just do the below on every module. That's it.

```python
import logging

logger = logging.getLogger(__name__)

# And then use logger.info(), logger.error(), etc., everywhere.
```

It works like a charm.

ValueRaider:
Thanks for the details, and I hear you! I'm just thinking of the users that haven't used `logging`, and our request that they review/provide debug logs with Issues. So I've moved this logic in function `yf.enable_debug_mode()`.

This latest branch should work for you now.

flaviovs:
Amazing, thanks @ValueRaider

Issue #1554, opened 2023-06-09
Missing ETF weights 


lcrmorin:
First of all let me take this opportunity to thank you for greatly improving the time it takes to get ticker info. Some of my code that took 15+ mins to run now run in a minute. 

However something else changed in the code. More specifically I was able to get sector weights for ETF. This data seems to still be available on yahoo (see: https://finance.yahoo.com/quote/IVV/holdings?p=IVV, sector weighing). But this info doesn't seems to be available in the info. 

Is this an expected change ? Am i missing something ?

ValueRaider:
`info` fetch was completely redone in PRs #1461 and #1480 - converting webpage scraping to API requests.

I had a quick look and fix super-easy, so can I get you to submit a patch to `dev` branch #1084? It's simply adding `'topHoldings'` to `modules` in `quote.py` - look at code and should be obvious.

ValueRaider:
Also add `'price'`, this will restore more data. I think there's another issue about this, let me find it ...

AKwoKWH:
> `info` fetch was completely redone in PRs #1461 and #1480 - converting webpage scraping to API requests.
> 
> I had a quick look and fix super-easy, so can I get you to submit a patch to `dev` branch #1084? It's simply adding `'topHoldings'` to `modules` in `quote.py` - look at code and should be obvious.

Not sure if I added the patch correctly.
Can we also have sector/asset class and country allocation of mutual funds?
eg: https://finance.yahoo.com/quote/HGIYX/holdings?p=HGIYX

AKwoKWH:
I further check the codes, 
It seems it is getting the page "Holders" instead of "Holdings"

PR #1571, opened 2023-06-24
Update quote.py

AKwoKWH:
To add topHoldings and price for ETF weighting
https://github.com/ranaroussi/yfinance/issues/1554


ValueRaider:
The reason there are other changes is you created your branch from `main`, but it should be from `dev`. Branch from `dev` to merge into `dev`, understand?

ValueRaider:
And move `major_holdings` into a separate pull request, because it's a different feature. Also, note that for non-ETF symbols e..g 'AMZN', `major_holdings` returns price information instead, maybe should return `None`.

ValueRaider:
@AKwoKWH Do you plan to complete this?

dotJoel:
i have a branch that replicates and finishes this work.
unless #1784 is the preferred direction.

ValueRaider:
@bot-unit

PR #1608, opened 2023-07-15
Fix to enable fetching all (including last) rows of intraday data

e9henrik:
This pull request enables fetching all intraday data, even the last, or "close" row, which was previously not fetched or dropped.

I don't know why, but without useYfid=true, the last row is dropped from the server response.


ValueRaider:
Can you add a unit test for this scenario, because I'm not sure what this is fixing.

Issue #898, opened 2021-11-27
Lookup by ISIN doesn't work with yf.download (Version 0.1.67)


Marc9988:
Hi Ran,

Lookup by ISIN seems not to work with yf.download. I use version 0.1.67 of YFinance and Python 3.6.9.

```
msft = yf.download('US5949181045', start='2020-01-01', end='2020-01-15')
```

I receive this error message:

```
File "C:\Users\Marc\anaconda3\envs\borse36\lib\site-packages\yfinance\multi.py", line 138, in download
    return shared._DFS[shared._ISINS.get(ticker, ticker)]

KeyError: 'US5949181045'
```

But with yf.Ticker it works properly.

```
msft2 = yf.Ticker('US5949181045')
```

Would be nice to find a solution for the problem. Anyway thank you so much for your great tool!

Marc9988:
Thank you very much!! :-)

PR #1611, opened 2023-07-17
[BUG] Fix failure when using single ISIN as a ticker (#1525)

### LINKED ISSUE #1525: Falure when using single ISIN as a ticker (fix included)
Multiple tickers with ISINs mixed in do work, but a single ISIN fails. Here is the culprit (multi.py):

>     if len(tickers) == 1:
>        ticker = tickers[0]
>        return shared._DFS[shared._ISINS.get(ticker, ticker)]

The last line should be just

> return shared._DFS[ticker]

as at that point in time both _DFS and tickers refer to Yahoo ID instead of ISIN, and lookup by original ISIN in _DFS is failing. 

Testcase:

>import yfinance as yf
>yf.download("SE0000594111", period="7d", repair=True, progress=False, threads=False, timeout=30)

(corresponding Yahoo ID is '0P00000FYR.ST'.

ricardoprins:
Fix #1525, #898.

I am submitting this directly to main since this bug is already in production.

### DIFF
--- a/yfinance/multi.py
+++ b/yfinance/multi.py
@@ -207,7 +207,7 @@ def download(tickers, start=None, end=None, actions=False, threads=True, ignore_
 
     if len(tickers) == 1:
         ticker = tickers[0]
-        return shared._DFS[shared._ISINS.get(ticker, ticker)]
+        return shared._DFS[ticker]
 
     try:
         data = _pd.concat(shared._DFS.values(), axis=1, sort=True,


ValueRaider:
Great. Can you search through [other ISIN issues](https://github.com/ranaroussi/yfinance/issues?q=is%3Aissue+is%3Aopen+isin) and add if fixed

ricardoprins:
Sure, I'll take a look at it.

ricardoprins:
So, I found 8 related issues; I'll go through them one at a time, and I'll post the corresponding conclusions for the tests inside each of the issues. It will take me a few days to set up the proxy for the ones involving proxy connection, but I'll try to get these done as well. Hope it helps!

ValueRaider:
Oh don't worry about the proxy ones, I just meant see if this PR fixes other issues. Duplicate issues are common.

ricardoprins:
That works for me. From a quick glance, though, it seems that the proxy issues are already fixed. I have some free time on my hands, I'll "sit around" in this repo for a while :)

ricardoprins:
Fixes #898 as well.

Issue #1616, opened 2023-07-18
Security issue: bump requests to 2.31 because of CVE-2023-32681


ricardoprins:
Link to [CVE-2023-32681](https://devhub.checkmarx.com/cve-details/CVE-2023-32681/).
TLDR: `Exposure of Sensitive Information to an Unauthorized Actor vulnerability pending CVSS allocation`

This was fixed on requests v.2.31 [see info](https://github.com/advisories/GHSA-j8r2-6x86-q33q).

So, the fix would be bumping requests from >=2.26 to >=2.31 in all corresponding places.

I can send a pull request to make the fix and test to see if this might require any modifications in the code (which I don't think will be the case, from what I've seen so far on the codebase here).

ricardoprins:
Closing; fixed with #1625.

PR #1625, opened 2023-07-20
Bump requests to 2.31 and removes cryptography.

ricardoprins:
As suggested in #1616. I messed up something here and reverted the whole thing.

- Bumps requests to 2.31
- Removes cryptography (unnecessary)


ValueRaider:
Can this go into `dev` branch instead? #1084

ricardoprins:
It sure can. I've only sent this to main because it's mostly a security flaw, and I thought it would be more urgent. However, if it preferred for me to send it there, I can do it.

ValueRaider:
Good point, this is serious for any proxy users.

Issue #1624, opened 2023-07-20
PEP 8 Coding Style Adjustments


ricardoprins:
I know this is not exactly an urgent matter...however, cleaner code is usually easier to maintain (and often easy to understand). 

There are a total of 633 PEP 8 coding style violations. I can submit a huge PR fix that adjusts the entire code or separate PRs for each code. Here's my current list (considering the main branch)

- [x] base.py (102 PEP 8 coding style violations)
- [ ] const.py (363 PEP 8 coding style violations)
- [ ] context.py (10 PEP 8 coding style violations)
- [ ] data.py (1 PEP 8 coding style violations)
- [ ] fundamentals.py (3 PEP 8 coding style violations)
- [ ] holders.py (1 PEP 8 coding style violations)
- [ ] multi.py (5 PEP 8 coding style violations)
- [x] prices.py (64 PEP 8 coding style violations)
- [ ] quote.py (29 PEP 8 coding style violations)
- [ ] setup.py (1 PEP 8 coding style violations)
- [ ] ticker.py (5 PEP 8 coding style violations)
- [ ] tickers.py (1 PEP 8 coding style violations)
- [ ] fundamentals.py (47 PEP 8 coding style violations)

After that, another suggestion would be to add a linter action that automatically scans the submitted code with each PR.

ValueRaider:
Generally this is a good idea, but just remember these are guidelines not strict rules, so be thoughtful. My personal preference is to exempt `E702: multiple statements on one line` - this is done rarely but for good reason. May also need to exempt some `E501 line too long`, or adjust the max length.

ricardoprins:
Yeah, we can filter some of those. However there are some more serious things such as imports in the middle of the code, bare excepts, and things like that. 

I can send a sample PR for one of the files, just to have an idea of the kind of modifications I'm talking about.

ValueRaider:
Sure go ahead

PR #1627, opened 2023-07-21
Adjust PEP 8 + minor improvement

ricardoprins:
As per discussion in #1624.

- Adjusted prices.py according to PEP 8 coding style suggestions, except:

	- E501 line too long
	- E701 multiple statements in one line

- Changed tkr_div_dates declaration/assignment style within TestPriceHistory::test_dailyWithEvents (this is the only thing non-PEP8 here)

The bare except changes merit another future discussion about how exceptions are being handled in the tests; I believe that with some tweaks to the code, exceptions can be handled more gracefully (this is something I also noted in other files such as base.py)

Alternatively, test_intraDayWithEvents and test_intraDayWithEvents_tase are being defined twice...is this on purpose?  (perhaps this should be another issue?)

### DIFF
--- a/tests/prices.py
+++ b/tests/prices.py
@@ -59,7 +59,7 @@ class TestPriceHistory(unittest.TestCase):
             dt1 = df.index[-1]
             try:
                 self.assertNotEqual(dt0.hour, dt1.hour)
-            except:
+            except AssertionError:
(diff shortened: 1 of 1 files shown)


ricardoprins:
Closing; will submit to dev

PR #1630, opened 2023-07-22
Adjust PEP 8 + minor improvements + f-strings in base.py

ricardoprins:
As per discussed in #1624.

- `prices.py`: minor correction 
- `base.py`: 
   - multiple PEP 8 adjustments (ignoring E702 and E501)
   - removal of old .format-styled strings

----
In [`base.py:867`](https://github.com/ranaroussi/yfinance/blob/dev/yfinance/base.py#L867) there is an import mid-code that I have preserved; not sure what to do with it since it is conflicting with the comment right above it:

```python
        # Only import scipy if users actually want function. To avoid
        # adding it to dependencies.
        from scipy import ndimage as _ndimage
```

If this is handled by some other PR, or if I should just ignore this for now, please let me know.

There are a few other issues with this file, but for the sake of organization, I will submit a future issue/PR touching these other warnings/errors.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -196,7 +196,7 @@ class TickerBase:
         tz = self._get_ticker_tz(proxy, timeout)
         for k in ["period1", "period2"]:
             if k in params_pretty:
-                params_pretty[k] = str(_pd.Timestamp(params[k], unit='s').tz_localize("UTC").tz_convert(tz))
+                params_pretty[k] = str(pd.Timestamp(params[k], unit='s').tz_localize("UTC").tz_convert(tz))
         logger.debug(f'{self.ticker}: Yahoo GET parameters: {str(params_pretty)}')
 
         # Getting data from json
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -21,33 +21,31 @@
 
 from __future__ import print_function
 
-import warnings
-import time as _time
 import datetime as _datetime
-import dateutil as _dateutil
+import json as _json
+import logging
+import time as _time
+import warnings
 from typing import Optional
-
-import pandas as _pd
-import numpy as _np
-import pandas as pd
-
-from .data import TickerData
-
 from urllib.parse import quote as urlencode
 
-from . import utils
+import dateutil as _dateutil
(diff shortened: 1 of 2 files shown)


yfinance/base.py
```diff
@@ -21,34 +21,32 @@
...
+import time as _time
+import warnings
 from typing import Optional
+from urllib.parse import quote as urlencode
 
-import pandas as _pd
+import dateutil as _dateutil
 import numpy as _np
+import pandas as _pd
 import pandas as pd
```
ValueRaider, line 35: Can one of these be removed?

ValueRaider:
Good but I'll leave PR open this time, guessing you plan on editing other files.

ValueRaider:
`from scipy import ndimage as _ndimage`

Could be moved to top if wrapped in a try-catch. But if style is only complaint, then can leave.

ricardoprins:
I think it could stay there for now then. I think maybe with some refactoring done in the file in the future, other options could be considered.

yfinance/base.py
ricardoprins: Totally. I avoided messing with most non-PEP 8 related stuff, save those I saw were already adopted, such as converting old .format to f-string. 

I don't fully understand why using `_pd` (or some of the other underscored imports in other parts of the codebase). I can go ahead and remove it if that's okay.

ValueRaider: Feels like it should be a PEP 8 violation, basically a duplicate import.

ricardoprins: It isn't a PEP 8 violation, technically. 

The underscore usage that is seen here is weird to me (along with a few other things I've been seeing on this codebase)... usually, _s are indicative of private modules...however, it doesn't make sense since pandas is also being imported/used within the same module with open usage (and used outside of the module as well)

Anyways, if that's okay, I'd simply remove the _pd (and replace all its usage for regular pd).

ValueRaider: Underscores were here before me, and you care more about style than me, so do what you think best.

ricardoprins:
I think these commits take care of it.

PR #1643, opened 2023-08-01
Add option to extract trailing data out of yahoo finance.

init-js:
Adds ticker.ttm_income_stmt and ticker.ttm_cashflow

Unlike the yahoo website, the API outputs two timestamped columns. One column is dated from the last annual financial release, and the other one is dated from the last quarterly financial report. The TTM calculation on the website appears to be using the yearly information. The calculation appears to be slightly different, otherwise.

https://github.com/ranaroussi/yfinance/discussions/1596


ValueRaider:
Thanks for implementing this. But can you move into `dev` branch instead of `main`? #1084

PR #1690, opened 2023-09-21
Update utils.py for position-based indexing

svraghu:
Updated utils.py to use .iloc[] for position-based indexing since integer-based indexing ["Open"][n - 1] will be deprecated in future Pandas


ValueRaider:
Thanks for contribution but is already fixed in `dev` branch #1672. Planning to merge and release next few days.

Issue #1700, opened 2023-09-25
Importing yf fails if ~/.cache directory doesn't exist


blackary:
### Describe bug

Even though yf has a .set_tz_cache_location, simply importing yfinance fail if `~/.cache` directory doesn't exist

Simply importing yf should not have side-effects like creating a database, and especially should not fail before there is a chance to overwrite the cache directory.

This is particularly problematic when it's not possible to create the .cache directory as the app is being deployed to a 3rd-party system.

### Simple code that reproduces your problem

```python
# Use a non-existing directory for ease of testing
import appdirs as ad

ad.user_cache_dir = lambda *args: ".cache"

# If the ad.user_cache_dir directory doesn't exist, this fails
import yfinance as yf

# This line is never reached
yf.set_tz_cache_location(".cache")
```

### Debug log

```
python test.py
Traceback (most recent call last):
  File "/private/tmp/repos/yf/test.py", line 5, in <module>
    import yfinance as yf  # noqa
  File "/private/tmp/repos/yf/.direnv/python-3.10/lib/python3.10/site-packages/yfinance/__init__.py", line 23, in <module>
    from .ticker import Ticker
  File "/private/tmp/repos/yf/.direnv/python-3.10/lib/python3.10/site-packages/yfinance/ticker.py", line 29, in <module>
    from .base import TickerBase
  File "/private/tmp/repos/yf/.direnv/python-3.10/lib/python3.10/site-packages/yfinance/base.py", line 38, in <module>
    from . import utils
  File "/private/tmp/repos/yf/.direnv/python-3.10/lib/python3.10/site-packages/yfinance/utils.py", line 991, in <module>
    class _KV(_peewee.Model):
  File "/private/tmp/repos/yf/.direnv/python-3.10/lib/python3.10/site-packages/yfinance/utils.py", line 995, in _KV
    class Meta:
  File "/private/tmp/repos/yf/.direnv/python-3.10/lib/python3.10/site-packages/yfinance/utils.py", line 996, in Meta
    database = _DBManager.get_database()
  File "/private/tmp/repos/yf/.direnv/python-3.10/lib/python3.10/site-packages/yfinance/utils.py", line 953, in get_database
    cls._initialise()
  File "/private/tmp/repos/yf/.direnv/python-3.10/lib/python3.10/site-packages/yfinance/utils.py", line 972, in _initialise
    _os.mkdir(cls._cache_dir)
[...]

PR #1705, opened 2023-09-27
Fix TZ cache exception blocking import

ValueRaider:
Hopefully fix TZ cache exception blocking import #1700

The Singleton pattern makes unit testing this tricky, so can someone affected try branch? @rickturner2001 

Also:
- relocate cache init lock
- add test for setTzCacheLocation()

### DIFF
--- a/tests/ticker.py
+++ b/tests/ticker.py
@@ -152,21 +152,9 @@ class TestTicker(unittest.TestCase):
         tkr = "IBM"
         dat = yf.Ticker(tkr, session=self.session)
 
-        dat._fetch_ticker_tz(proxy=self.proxy, timeout=5, debug_mode=False, raise_errors=False)
-        dat._get_ticker_tz(proxy=self.proxy, timeout=5, debug_mode=False, raise_errors=False)
+        dat._fetch_ticker_tz(proxy=self.proxy, timeout=5)
+        dat._get_ticker_tz(proxy=self.proxy, timeout=5)
         dat.history(period="1wk", proxy=self.proxy)
-
-        v = dat.stats(proxy=self.proxy)
-        self.assertIsNotNone(v)
-        self.assertTrue(len(v) > 0)
-
-        v = dat.get_recommendations(proxy=self.proxy)
-        self.assertIsNotNone(v)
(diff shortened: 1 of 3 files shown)


rickturner2001:
Just tested the code on a readonly docker container and it works just fine. I wanted to ask you bout _TzCacheDummy

since the tz cache is registered by `_get_ticker_tz` in `base.py`, what made you want to initialize a cache with an instance of either _TzCache or _TzCache dummy instead of simply returning the tz value in the evnet  that we don't have r/w access

what made you want to crate the Dummy class instead of doing something like this? 

```python
 def _get_ticker_tz(self, proxy, timeout):
        if self._tz is not None:
            return self._tz

        # assume this exists
        has_rw_permissions = utils.can_rw_to_fs()
        
        tz = None
        cache = None
        
        if has_rw_permissions:
            cache = utils.get_tz_cache()
            tz = cache.lookup(self.ticker)

            if tz and not utils.is_valid_timezone(tz):
                # Clear from cache and force re-fetch
                cache.store(self.ticker, None)
                tz = None

        if tz is None:
            tz = self._fetch_ticker_tz(proxy, timeout)

            if utils.is_valid_timezone(tz):
                # info fetch is relatively slow so cache timezone
                if has_rw_permissions:
                    cache.store(self.ticker, tz)
            else:
                tz = None

        self._tz = tz
        return tz
```

I think maybe the class makes it easier to test, but maybe you can shed some light on this. Thanks

ValueRaider:
I didn't implement the sqlite architecture so couldn't answer. The person that did, seemed to know more about Python engineering than me so I didn't question design, but they aren't active anymore.
First implementation - #1105 
`_TzCacheDummy` - #1113

If I had to guess, maybe it's to avoid checking filesystem (`can_rw_to_fs`) every time need timezone. I believe the Python philosophy is, just try and catch exceptions, instead of meticulous if-else https://stackoverflow.com/a/7604717

rickturner2001:
is anyone still working on this?

ValueRaider:
I half forgot, half wasn't sure if you wanted changes.

PR #1707, opened 2023-09-29
Test Fix: Check for type and expect exceptions in tests

rickturner2001:
# Reason for pull request

Some of the attributes in `base.py`, `ticker.py` and the other main scripts may throw a `YFNotImplemented` error when accessing properties. This was not previously being checked for. Furthermore, I added some type checking 

## In addition 

I would like to improve the DX  for contributors in the repository by adding automated testing with github actions 
and maybe we could use this to have black and isort refactor the code appropriately and in a uniform way

If this proposal is well received, I will gladly take care of the integrations


ValueRaider:
When `assert_attribute_type()` catches a `YFNotImplementedError` exception, why does it call `getattr` again? It already knows it raised `YFNotImplementedError`. Sounds circular.

Otherwise ok, but better to intergrate changes into `tests/ticker.py` and delete file.

> I would like to add automated testing with github actions & refactor

Good proposal but first review `tests/context.py`, not sure how it'll work with github actions.

rickturner2001:
Tests are now passing in tests/ticker.py. Nice catch on the try/except, I meant to catch an `Exception` and then have the attribute raise it once again to confirm that it is an instance of `YFNotImplemented`

With my latest commit I changed the expected url which was raising an errorr in `TestTickerHistory.test_no_expensive_calls_introduced`

In addition I added a `proxy` attribute which can now be passed to `yf.Ticker` so that when we fetch for data, we don't have to keep on passing the proxy around.

rickturner2001:
> Although, thinking more, I'd prefer effort go to open pull requests abandoned by authors, to finish them.

Yes, that's fair enough. We'll put a pin on that

rickturner2001:
@ValueRaider I am done on tests/ticker.py you may push if you're confident with the changes

ValueRaider:
Move that proxy stuff into another pull request, keep these focused. Because maybe `proxy` argument needs to be deprecated from all other functions & README updated, I wasn't aware it could be set through `Ticker`.

Then git squash

rickturner2001:
I'll open a new Pr for the Proxy integration in Ticker. `test_goodTicker_withProxy` remains broken until then

ValueRaider:
Must this go into `main`? I prefer `dev`.

PR #1711, opened 2023-10-02
Refactor/ticker proxy

rickturner2001:
Added proxy attribute to the TickerBase class. Note that all the getters still allow to pass a proxy in the event that a user wishes to use different proxies for different actions.


rickturner2001:
@ValueRaider can you please take a quick look at this? I think you might be able to merge it into dev as is.

PR #1724, opened 2023-10-12
Bugfix/data types 2

mreiche:
Fixes the bug, when you want to perform a pandas operation on an empty list, which is returned when no data is present:
```
ERROR:yfinance:XXX: No timezone found, symbol may be delisted
ERROR:xxx:'list' object has no attribute 'loc'
```
This change does 2 things:
- It returns an static empty pandas series instead of a list
- It annotates the return type for some functions


mreiche:
I've updated the description. Please review.

yfinance/base.py
```diff
@@ -45,6 +45,9 @@
 from .const import _BASE_URL_, _ROOT_URL_
 
 
+_empty_series = pd.Series()
```
ValueRaider, line 48: Why?
- cost of `pd.Series()` is negligible
- `_empty_series` is more characters

requirements.txt
```diff
@@ -1,11 +1,13 @@
...
 multitasking>=0.0.7
 lxml>=4.9.1
 appdirs>=1.4.4
 pytz>=2022.5
 frozendict>=2.3.4
 beautifulsoup4>=4.11.1
 html5lib>=1.1
-peewee>=3.16.2
\ No newline at end of file
+peewee>=3.16.2
```
ValueRaider, line 13: revert changes

tests/ticker.py
```diff
@@ -798,7 +814,7 @@ def test_info(self):
 def suite():
     suite = unittest.TestSuite()
     suite.addTest(TestTicker('Test ticker'))
-    suite.addTest(TestTickerEarnings('Test earnings'))
+    #suite.addTest(TestTickerEarnings('Test earnings'))
```
ValueRaider, line 817: why comment?

ValueRaider (changes requested):
Nit-picky comments, but IMO every code change should be relevant and matter.

requirements.txt
mreiche: I reverted the explicit dependencies but kept the newline at the end.

tests/ticker.py
mreiche: I branched from the wrong base I guess. Reverted.

ValueRaider:
You might have a better time with Git if you rebase your changes.

mreiche:
> You might have a better time with Git if you rebase your changes.

I don't get what happend acutally. I rebased from master to dev and git messed something up. This is the second branch already. I usually don't use rebase because of these issues.

Anyway, is there something you request a change for?

mreiche:
> I'll merge this time, but in future:
> 
> * reference [Network Graph](https://github.com/ranaroussi/yfinance/network), keep it tidy
> * avoid merging in from both `main` and `dev`, it usually creates a mess

Alright

ValueRaider (approved):

PR #1736, opened 2023-11-05
feat/retry with fallback cookie and crumb

Jshen123:
retry with fallback cookie and crumb when default strategy failed

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -1656,7 +1656,8 @@ class TickerBase:
                 # info fetch is relatively slow so cache timezone
                 cache.store(self.ticker, tz)
             else:
-                tz = None
+                # tz = None
+                raise Exception(f"_fetch_ticker_tz() has returned None for '{self.ticker}', will cause problems. Investigate")
 
         self._tz = tz
         return tz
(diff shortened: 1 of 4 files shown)


ValueRaider:
This just looks like moving code around, nothing actually changed.

Jshen123:
> This just looks like moving code around, nothing actually changed.

@ValueRaider Previously, the fallback only kicks in [when `get_cookie_basic` fails to get crumb ](https://github.com/ranaroussi/yfinance/blob/a064937dcdf14a99bc3aff5cb6a883f870a2017e/yfinance/data.py#L283C42-L283C42)

However, even when you successfully get the crumb (with default strategy), the request will still get 401 response for some users (including myself). And in this case, code will never attempt fallback.

What the change does is, in the event default crumb is retrieved but the response fails, [it will attempt the fallback cookie and crumb](https://github.com/Jshen123/yfinance/blob/b1022ae228088032c7fb408c0b5dd7f70c728e74/yfinance/data.py#L335)

ValueRaider:
You're right, I thought `if response.status_code >= 400 ...` was just refactoring. Will merge after I squash my commits.

ValueRaider:
My squash has broken the PR. Rather that recreate, can you post a git patch file to my PR and I'll integrate? Means I can keep squashing.

Issue #1685, opened 2023-09-18
Running `.institutional_holders` give the following error - ValueError: time data "..." doesn't match format "%B %d, %Y", at position 1


Valachio:
### Describe bug

I was scraping the ticker ACST and getting institutional holders when I ran into this issue.

The issue can be replicated like this:

```
import yfinance
stock = yfinance.Ticker('ACST')
stock.institutional_holders # Gives error
```

The full error is:

```
ValueError: time data "Jun 29, 2023" doesn't match format "%B %d, %Y", at position 1. You might want to try:
    - passing `format` if your strings have a consistent format;
    - passing `format='ISO8601'` if your strings are all ISO8601 but not necessarily in exactly the same format;
    - passing `format='mixed'`, and the format will be inferred for each element individually. You might want to use `dayfirst` alongside this.
```

I'm using Python 3.11, and yfinance v0.2.28

### Simple code that reproduces your problem

import yfinance
stock = yfinance.Ticker('ACST')
stock.institutional_holders # Gives error

### Debug log

/Users/jack/Projects/ais/aisenv/lib/python3.11/site-packages/yfinance/scrapers/holders.py:39: FutureWarning: Passing literal html to 'read_html' is deprecated and will be removed in a future version. To read from a literal string, wrap it in a 'StringIO' object.
  holders = pd.read_html(resp.text)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/jack/Projects/ais/aisenv/lib/python3.11/site-packages/yfinance/ticker.py", line 114, in institutional_holders
    return self.get_institutional_holders()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jack/Projects/ais/aisenv/lib/python3.11/site-packages/yfinance/base.py", line 1662, in get_institutional_holders
    data = self._holders.institutional
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jack/Projects/ais/aisenv/lib/python3.11/site-packages/yfinance/scrapers/holders.py", line 26, in institutional
    self._scrape(self.proxy)
  File "/Users/jack/Projects/ais/aisenv/lib/python3.11/site-packages/yfinance/scrapers/holders.py", line 63, in _scrape
    self._mutualfund['Date Reported'] = pd.to_datetime(
                                        ^^^^^^^^^^^^^^^
[...]

PR #1740, opened 2023-11-09
Fix pandas FutureWarning: "Passing literal html to 'read_html' is deprecated"

mikez:
This addresses #1685 (`institutional_holders`) and also `get_earnings_dates()`.

Pandas issue is found here:
https://github.com/pandas-dev/pandas/issues/53767
and the change in code here:
https://github.com/pandas-dev/pandas/blob/5cedf87cccd77c7b4b6aaa64bfec98b32b512f68/pandas/io/html.py#L1238

As for legacy Python 2.7 support: `io.StringIO` seems to be supported in the versions I tested. See https://docs.python.org/2/library/io.html

### DIFF
--- a/tests/ticker.py
+++ b/tests/ticker.py
@@ -325,76 +325,77 @@ class TestTickerHistory(unittest.TestCase):
         self.assertFalse(data.empty, "data is empty")
 
 
-# Below will fail because not ported to Yahoo API
-# class TestTickerEarnings(unittest.TestCase):
-#     session = None
+class TestTickerEarnings(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


ValueRaider:
Looks fine. I've restored `earnings_dates` unit tests. If no complaint then I'll merge.

mikez:
👍 Sounds good to me.
Do you want to remove or change the comment "Below will fail because not ported to Yahoo API"?

ValueRaider:
Good spot. I've moved comment, and git-squashed so remember to pull/reclone.

PR #1741, opened 2023-11-15
fix: update unsupported operand in request_args

Jshen123:
address an issue where operand `|` is not supported in python.

### DIFF
--- a/yfinance/data.py
+++ b/yfinance/data.py
@@ -358,7 +358,7 @@ class YfData(metaclass=SingletonMeta):
 
         request_args = {
             'url': url,
-            'params': params | crumbs,
+            'params': {**params, **crumbs},
             'cookies': cookies,
             'proxies': proxy,
             'timeout': timeout,


yfinance/data.py
```diff
@@ -358,7 +358,7 @@ def get(self, url, user_agent_headers=None, params=None, proxy=None, timeout=30)
 
         request_args = {
             'url': url,
-            'params': params | crumbs,
```
Jshen123, line 361: @ValueRaider 
`|` is not supported and throws `unsupported operand type(s) for |: 'dict' and 'dict` when I try to make a request

ValueRaider:
This logic used several more times in data.py.

And here: https://github.com/ranaroussi/yfinance/blob/1b0d8357d6edf6c994ee63fe8536ee6aee6f1b10/yfinance/utils.py#L697

Jshen123:
> This logic used several more times in data.py.
> 
> And here:
> 
> https://github.com/ranaroussi/yfinance/blob/1b0d8357d6edf6c994ee63fe8536ee6aee6f1b10/yfinance/utils.py#L697

it looks like the issue is [caused by me using python 3.8](https://pyfr.discourse.group/t/mesh-import-error-unsupported-operand-type-s-for-dict-and-dict/576) 😅

PR #1762, opened 2023-12-07
Pandas future proofing

### LINKED ISSUE #1739: Some updates on pandas to avoid incoming errors
### Describe bug

Pandas 2.x will raise those errors in the future. The proposed solutions are also available in version 1.x so I think they must be implemented.

N/A

### Debug log

yfinance/base.py:707: FutureWarning: Series.fillna with 'method' is deprecated and will raise in a future version. Use obj.ffill() or obj.bfill() instead.
div_adjusts = div_adjusts.fillna(method='bfill').fillna(method='ffill')

yfinance/base.py:723: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
div_adjusts[idx] = div_adjusts[idx - 1]

yfinance/base.py:723: FutureWarning: Series.__setitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To set a value by position, use `ser.iloc[pos] = value`
div_adjusts[idx] = div_adjusts[idx - 1]

### Bad data proof

_No response_

### `yfinance` version

0.2.31

### Python version

3.10

### Operating system

linux

JuliaLWang8:
Fixes #1739 : 
- FutureWarning deprecation warning for pandas resolved by changing fillna and iloc.
- Added a unit test in test_goodTicker_withProxy for _reconstruct_intervals_batch, and fixed the previous test.


coskos-ops:
excellent change

PR #1766, opened 2023-12-08
Pandas future proofing

### LINKED ISSUE #1739: Some updates on pandas to avoid incoming errors

JuliaLWang8:
Fixes #1739 : 
- FutureWarning deprecation warning for pandas resolved by updating fillna and iloc in the `reconstruct_intervals_batch` function.
- Added a unit test to tests/ticker.py in TestTickerHistory called `test_reconstruct_intervals_batch`, also validated against current output.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -706,33 +706,33 @@ class TickerBase:
                     # But in case are repairing a chunk of bad 1d data, back/forward-fill the 
                     # good div-adjustments - not perfect, but a good backup.
                     div_adjusts[f_tag] = np.nan
-                    div_adjusts = div_adjusts.fillna(method='bfill').fillna(method='ffill')
+                    div_adjusts = div_adjusts.ffill().bfill()
                     for idx in np.where(f_tag)[0]:
                         dt = df_new_calib.index[idx]
                         n = len(div_adjusts)
                         if df_new.loc[dt, "Dividends"] != 0:
                             if idx < n - 1:
                                 # Easy, take div-adjustment from next-day
-                                div_adjusts[idx] = div_adjusts[idx + 1]
+                                div_adjusts[idx] = div_adjusts.iloc[idx + 1]
                             else:
                                 # Take previous-day div-adjustment and reverse todays adjustment
                                 div_adj = 1.0 - df_new_calib["Dividends"].iloc[idx] / df_new_calib['Close'].iloc[
                                     idx - 1]
-                                div_adjusts[idx] = div_adjusts[idx - 1] / div_adj
+                                div_adjusts[idx] = div_adjusts.iloc[idx - 1] / div_adj
                         else:
                             if idx > 0:
                                 # Easy, take div-adjustment from previous-day
-                                div_adjusts[idx] = div_adjusts[idx - 1]
+                                div_adjusts[idx] = div_adjusts.iloc[idx - 1]
                             else:
                                 # Must take next-day div-adjustment
-                                div_adjusts[idx] = div_adjusts[idx + 1]
+                                div_adjusts[idx] = div_adjusts.iloc[idx + 1]
                                 if df_new_calib["Dividends"].iloc[idx + 1] != 0:
                                     div_adjusts[idx] *= 1.0 - df_new_calib["Dividends"].iloc[idx + 1] / \
                                                         df_new_calib['Close'].iloc[idx]
                     f_close_bad = df_block_calib['Close'] == tag
                     df_new['Adj Close'] = df_block['Close'] * div_adjusts
                     if f_close_bad.any():
-                        df_new.loc[f_close_bad, 'Adj Close'] = df_new['Close'][f_close_bad] * div_adjusts[f_close_bad]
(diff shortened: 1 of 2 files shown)


JuliaLWang8:
I also had to manually pip install packages not in `requirements.txt`: scipy, requests_cache, and requests_ratelimiter. Is there a reason those aren't included in the reqs / should they be added?

yfinance/base.py
```diff
@@ -706,33 +706,33 @@ def _reconstruct_intervals_batch(self, df, interval, prepost, tag=-1):
...
-                                div_adjusts[idx] = div_adjusts[idx + 1]
+                                div_adjusts[idx] = div_adjusts.iloc[idx + 1]
                                 if df_new_calib["Dividends"].iloc[idx + 1] != 0:
                                     div_adjusts[idx] *= 1.0 - df_new_calib["Dividends"].iloc[idx + 1] / \
                                                         df_new_calib['Close'].iloc[idx]
                     f_close_bad = df_block_calib['Close'] == tag
                     df_new['Adj Close'] = df_block['Close'] * div_adjusts
                     if f_close_bad.any():
-                        df_new.loc[f_close_bad, 'Adj Close'] = df_new['Close'][f_close_bad] * div_adjusts[f_close_bad]
+                        df_new.loc[f_close_bad, 'Adj Close'] = df_new['Close'].iloc[f_close_bad] * div_adjusts[f_close_bad]
```
ValueRaider, line 735: Was there a warning/error on this? Because this lookup isn't like the others.

ValueRaider (changes requested):
question

ValueRaider:
> Maybe we should add this requirements with pip[optional]? [docs](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) [stockoverflow](https://stackoverflow.com/questions/46775346/what-do-square-brackets-mean-in-pip-install)

Fine with me. I propose naming `pip install yfinance[nospam]`

yfinance/base.py
JuliaLWang8: You're right, this one doesn't really need it now, I removed it in [0bcd2dc](https://github.com/ranaroussi/yfinance/pull/1766/commits/0bcd2dc725e9a40a306db96370c3be60b8454a33).

JuliaLWang8:
> > Maybe we should add this requirements with pip[optional]? [docs](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) [stockoverflow](https://stackoverflow.com/questions/46775346/what-do-square-brackets-mean-in-pip-install)
> 
> Fine with me. I propose naming `pip install yfinance[nospam]`

Should we include scipy as an optional req too? I noticed it was also purposely excluded in [base.py#L872](https://github.com/ranaroussi/yfinance/blob/f08fe83290136d103d46d67524f5b6e7b6b827ff/yfinance/base.py#L872)

ValueRaider:
> Should we include scipy as an optional req too? I noticed it was also purposely excluded in [base.py#L872]

I don't mind, but if yes then I propose naming `[repair]`, because only used by price repair.

> pandas_datareader

[No](https://github.com/pydata/pandas-datareader/issues/977#issuecomment-1786841251)

ValueRaider (approved):

JuliaLWang8:
> I don't mind, but if yes then I propose naming `yfinance[repair]`, because only used by price repair.
> 
> 
> 
> > pandas_datareader
> 
> 
> 
> [No](https://github.com/pydata/pandas-datareader/issues/977#issuecomment-1786841251)
> 
> 
> 
> 
> 
> 

Sounds good, I can make the mentioned changes to `setup.py`. Should I make a new PR for that or just do it here? 

Also with respect to documentation, does this need to be added somewhere like the readme?

ValueRaider:
Create a new PR, reference this for background discussion. Should have done this earlier tbh.

Update README too, maybe move install section to top, and replace requirements with link to file.

ValueRaider (approved):

ValueRaider:
I've made some small tweaks. If happy, then Git squash (#1084) then I will merge.

Issue #1737, opened 2023-11-07
Getting the data for revenue & earning


BenResCyber:
Hi, I need to get the data to show up like the chart below.
<img width="332" alt="Screenshot 2023-11-07 at 10 48 16" src="https://github.com/ranaroussi/yfinance/assets/150105166/f11c7757-8c9a-4139-b691-cd7612679d8c">
https://finance.yahoo.com/quote/AAPL?p=AAPL

Is there any possible way to do it? Please help

BenResCyber:
Found out by the endpoint `/quoteSummary?module=earnings`

reisenmachtfreude:
Is it possible to access this endpoint using yfinance? If yes, how?

ValueRaider:
Any endpoint can be accessed by yfinance, just need to write the Python code #1084

HWilson65:
Did you manage to get the data using `ticker.get_earnings(freq="quarterly")` or did you ended up implementing the endpoint directly?

The function currently has the not very helpful message of `yfinance.exceptions.YFNotImplementedError: Have not implemented fetching 'earnings' from Yahoo API`, so if you ended up using the mentioned endpoint it may be worthwhile to add it to the project.

I imagine to implement the endpoint something like the steps detailed in #1592 had to be done to get the cookie and crumb for the earnings module request. If this is the case it could be added to the wiki for future reference.

Graeme22:
For those who only need recent EPS, that's included in the income statement, so no need to make a separate call.

Tikam02:
How to make to the separate call to get EPS and expected earnings?

dokson:
Opened #2823 which restores `Ticker.earnings` and adds `Ticker.eps_history`. The data was actually available the whole time via `quoteSummary?modules=earnings` — the deprecation note *"not available via API"* turned out to be inaccurate (the old HTML-page scraper had silently broken, and the property was bypassed rather than ported to the JSON endpoint). cc @HWilson65 @Tikam02 @BenResCyber @Graeme22

PR #1770, opened 2023-12-10
Feature/added financials plot

JuliaLWang8:
Fixes issue #1737 , the plot should be the same as the one seen on the [yahoo financials page](https://finance.yahoo.com/quote/AAPL/financials?p=AAPL).

This can be tested by running the below function and switching between yearly and quarterly.
```
apple = yf.Ticker('AAPL')
apple.get_financials_plot(freq='yearly')
```

Note:
- This does use matplotlib which needs to be added to the dependencies for whoever wants to use it. We could also add this as optional via `pip install yfinance[optional]`, let me know if you need me to add that into setup.py 
- I decided against developing another get request to yahoo for the plot (as mentioned in the issue) since we already have the data for it which can be extracted from the income statement we already have developed. 
- I also decided against returning the plot and just using `plt.show()` directly in the function. This is because I'm unsure that users will know to call `plt.show()` on the output if it was just the figure, and I personally find it more user-friendly just have the figure open after running the function. Open to other people's opinions though!

### DIFF
--- a/yfinance/scrapers/fundamentals.py
+++ b/yfinance/scrapers/fundamentals.py
@@ -69,7 +69,7 @@ class Financials:
             res[freq] = self._fetch_time_series("cash-flow", freq, proxy)
         return res[freq]
     
-    def get_financials_plot_from_ics(self, freq="yearly", proxy=None) -> plt.figure:
+    def get_financials_plot_from_ics(self, freq="yearly", proxy=None) -> None:
         if len(self._income_time_series) == 0:
             self._income_time_series = self.get_income_time_series(freq, proxy)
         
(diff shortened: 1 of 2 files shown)


ValueRaider:
Plots are beyond the scope of this package, need to stay focused on Yahoo. Also this assumes user running in a GUI.

I'm not against implementing another get to `/quoteSummary?module=earnings`, it *might* be less load on Yahoo than Income Statement = faster.

yfinance/scrapers/fundamentals.py
paulmcq, line 1: This is adding a blank line at the end of the code. Do not do that.

PR #1771, opened 2023-12-10
Feat/adding extra dependencies

JuliaLWang8:
Implemented extra dependencies as discussed in #1766.

This can be tested locally via the command `pip install -e .[extras]` by replacing extras with:
- `nospam`: requests dependencies for custom requests and testing
    -  requests_cache, requests_ratelimiter
- `repair`: for price repairing functionality 
    - scipy

Maybe also add to #1084 that `nospam` is required to run tests and link the Installation instructions in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,21 +50,15 @@ Install `yfinance` using `pip`:
 $ pip install yfinance --upgrade --no-cache-dir
 ```
 
-To install `yfinance` using `conda`, see
-[this](https://anaconda.org/ranaroussi/yfinance). 
+[With Conda](https://anaconda.org/ranaroussi/yfinance).
 
-Test new features by installing betas, provide feedback in [corresponding Discussion](https://github.com/ranaroussi/yfinance/discussions):
+To install with optional dependencies, replace `optional` with: `nospam` for [caching-requests](#smarter-scraping), `repair` for [price repair](https://github.com/ranaroussi/yfinance/wiki/Price-repair), or `nospam,repair` for both:
+
 ``` {.sourceCode .bash}
-$ pip install yfinance --upgrade --no-cache-dir --pre
+$ pip install yfinance[optional]
 ```
 
-To install with extra dependencies, replace `extra` with `nospam` for custom requests or `repair` for price repairing functionality:
-``` {.sourceCode .bash}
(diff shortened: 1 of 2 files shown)


JuliaLWang8:
Regarding the naming, I think `nospam` makes sense concerning development because it's only needed for testing. However, since users could also install via the pip cmd to make custom requests (as in the readme), should we change it to reflect its use case?

ValueRaider:
> should we change it to reflect its use case?

Isn't the use case same? Reduce spam on Yahoo.

JuliaLWang8:
> > should we change it to reflect its use case?
> 
> Isn't the use case same? Reduce spam on Yahoo.

The way I interpreted the readme 'Smarter scraping' section was that the use case of `requests_cache` is a custom way to make requests via modifying User-agent. I guess there's also caching and the uses of `requests_ratelimiter` include the rate-limiting to avoid triggering Yahoo's blocker, so nospam makes sense.

ValueRaider:
> The way I interpreted the readme 'Smarter scraping' section was that the use case of `requests_cache` is a custom way to make requests via modifying User-agent.

That's a README problem so my fault. To keep it concise I combined 2 examples:
- custom User-agent on a session
- replace `requests` with `requests-cache`

Maybe User-agent example should be removed.

JuliaLWang8:
> Maybe User-agent example should be removed.

Should we change/remove it in the readme for this PR or just keep as is for now?

ValueRaider:
Can lower scipy minimum to 1.6.3.

> Should we change/remove it in the readme for this PR or just keep as is for now?

I don't mind if in this PR. As you had confusion, probably you're best to decide whether/how to change.

JuliaLWang8:
> Can lower scipy minimum to 1.6.3.
> I don't mind if in this PR. As you had confusion, probably you're best to decide whether/how to change.

Added these changes in my latest commit.

ValueRaider:
I've made some 'personal preference' tweaks to README. If no complaint, then I'll merge.

JuliaLWang8:
> I've made some 'personal preference' tweaks to README. If no complaint, then I'll merge.

Looks good to me!

Issue #1280, opened 2023-01-06
How can i get Growth Estimates ? 


mzaeemnasir:
**Describe the problem**
Is there any way to get Growth Estimates (Next 5 Years ? )

g0t0wasd:
Hi, were you able to get the Estimates?

Issue #575, opened 2021-01-31
Growth Estimates


darth57:
Is there a way to get the Yahoo Finance growth estimates specifically the "Next 5 Years (per annum)"?

JulioAlamill:
Was about to ask the same,  would be nice to have it in yfinance.Tickers.tickers[symbol].info

asafravid:
I will check
Recently the proper trailingPegRatio was fixed so this should be similar

kocielnik:
Related to #1280

dementedhedgehog:
This would be good to have for calculating Graham's Valuation (for example).

ValueRaider:
@dementedhedgehog Can you finish #1778 ?

dementedhedgehog:
https://github.com/ranaroussi/yfinance/pull/1778 is not my PR and I doubt I can edit it.   If @coskos-ops has dropped off the radar I suppose I could resurrect it in a new PR?

ValueRaider:
I've merged it into a new branch `feature/growth-estimates`, you can fork and pull-request that.

dementedhedgehog:
https://github.com/ranaroussi/yfinance/pull/1828

Is this what you're after?  Once you start pulling at this thread you start to wonder where to stop.   That original PR removes fields that aren't implemented but that are referenced in the rest of the code and some of the tests.

Also I've discovered I'm really rusty with git.

PR #1778, opened 2023-12-11
Added support for getting growth estimates

coskos-ops:
Added support for getting growth estimates. Some people wanted this feature as seen in #1280 and #575. Added basic scraping logic for the analysis section. 

Only growth estimates are added for now but it can easily be extended to support some of the other data in that section. 

Added a test case to check if the growth estimates are properly being retrieved.

Edit: Updated the PR to save all the other information available on the analysis page. The code already scrapes all of the data on the analysis page so it is wasteful to discard all but one entry


coskos-ops:
@bot-unit Should I save all the data available from the scraping? I only saved Growth Estimates because people were specifically asking for this feature. It would not be much work to extract the values of Earnings Estimate, Revenue Estimate, Earnings History, EPS Trend, and EPS revisions. They are already available in the analysis variable.

coskos-ops:
@bot-unit I saved all the other data on the analysis page. No point discarding data for which we have already paid the tax of retrieving

ValueRaider:
Benchmark with `time.perf_counter`?

ValueRaider:
@coskos-ops Just looking at code there's probably opportunity to format types etc. Correct me if wrong. See `Ticker.get_earnings_dates` for inspiration.

ValueRaider:
Changed target to branch `feature/growth-estimates` so someone else can finish. @dementedhedgehog

PR #1782, opened 2023-12-13
Drop empty dataframes before merge with appropriate error handling

### LINKED ISSUE #1382: Dataframe with incorrect index type returned when combining two tickers (one without data)
`import yfinance as yf
df = yf.download( tickers = ["^NDX", "^TWII"],       
                  start="2023-01-22", end="2023-01-29",
                  interval = "1m",
                  group_by = 'column',
                  auto_adjust = False,
                  prepost = True,
                  threads = True,
                  proxy = None
                  )'
[*********************100%***********************]  2 of 2 completed

1 Failed download:
- ^TWII: No data found for this date range, symbol may be delisted

It's not delisted, it's just there as there is no data due to the lunar new year.
The problem is the returned dataframe index is not a datetime index, but has type object)

yfinace version 0.2.9, python 3.10.8 on Linux ubuntu

ChNgineer:
Resolves #1382 and Implements suggested filtering of empty DataFrames before merge and realignment to address TypeError of object column when dataframe is empty. Implemented appropriate error logging to inform user that merge with certain tickers was unsuccessful, but does not kill the process outright and returns a merged DataFrame of only the non-empty DataFrames.

### DIFF
--- a/yfinance/multi.py
+++ b/yfinance/multi.py
@@ -216,8 +216,21 @@ def download(tickers, start=None, end=None, actions=False, threads=True, ignore_
         return shared._DFS[ticker]
 
     try:
+        for key , df in shared._DFS.items():
+            assert not df.empty
         data = _pd.concat(shared._DFS.values(), axis=1, sort=True,
                           keys=shared._DFS.keys())
+    except AssertionError:
+        dfs = {}
+        empty_dfs = []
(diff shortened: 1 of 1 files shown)


ValueRaider:
What is output when one DataFrame empty, or how does user find out that happened?

PR #1783, opened 2023-12-13
updated bookValue to be TangibleBookValue

### LINKED ISSUE #936: stock.info["bookValue"] returns Book Value per Share
In the info dictionary, "bookValue" returns the Book Value per Share.
```
stock = yf.Ticker("AAPL")
stock.info["boookValue"]
```
returns `3.841`, which is according to https://finance.yahoo.com/quote/AAPL/key-statistics?p=AAPL the Book Value Per Share (mrq). I'd suggest either renaming the variable or returning the actual book value of the company to prevent confusion.

ChNgineer:
Resolves #936 by retroactively changing `bookValue` from book value per share to `TangibleBookValue` from `get_balance_sheet()` and changed original `bookValue` to `bookValuePerShare`

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -1751,6 +1751,9 @@ class TickerBase:
     def get_info(self, proxy=None) -> dict:
         self._quote.proxy = proxy or self.proxy
         data = self._quote.info
+        data['bookValuePerShare'] = data['bookValue']
+        bal_sheet = self.get_balance_sheet()
+        data['bookValue'] = bal_sheet.loc['TangibleBookValue'][0]
         return data
 
     def get_fast_info(self, proxy=None):


yfinance/base.py
```diff
@@ -1751,6 +1751,9 @@ def get_mutualfund_holders(self, proxy=None, as_dict=False):
     def get_info(self, proxy=None) -> dict:
         self._quote.proxy = proxy or self.proxy
         data = self._quote.info
+        data['bookValuePerShare'] = data['bookValue']
+        bal_sheet = self.get_balance_sheet()
```
ValueRaider, line 1755: I don't like this, introduces another GET. Most people fetching `info` don't want book value - let those that do directly request balance sheet.

PR #1785, opened 2023-12-13
Fixed start/end issue not downloading and giving error the ticker is unavailable

Onerafaz:
using the start/end arguments is outputting an error that the ticker is unavailable

CL1_data = yf.download('CL=F', start='2023-12-01', end='2023-12-02')

error:
CL=F: No data found for this date range, symbol may be delisted

my solution might be wrong, please review and assess if the problem is on my side only.


ValueRaider:
I can't reproduce, and I don't see how the code changes help, so raise an Issue.

PR #1806, opened 2023-12-30
Ticker.history() to raise HTTP request excs if raise_errors args is True

puntonim:
I'd like to be able to handle any exception that might happen while `Ticker.history()` performs the HTTP request.
I currently do it in a hackish way using a monkey patch.
With this PR I'd like to introduce a hook for a custom function that, when given, it is invoked when such exception happens.

Most basic example I can think of:
```py
import requests
import yfinance as yf
from pandas import DataFrame

def reraise(exc):
    raise exc

ticker = yf.Ticker("AAPL")

try:
    data: DataFrame = ticker.history(
        interval="1m",
        start=...,
        end=...,
        timeout=10,
        raise_errors=True,
        request_exc_hook=reraise,
    )
except requests.ConnectTimeout as exc:
    ...
```

Note that my actual case a bit more complex to explain and boring (meaning not really useful for this proposal).

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -218,7 +218,8 @@ class TickerBase:
 
             data = data.json()
         except Exception:
-            pass
+            if raise_errors:
+                raise
 
         # Store the meta data that gets retrieved simultaneously
         try:


ValueRaider:
To have any chance of this being merged you're gonna need to explain your use case, because maybe there is a better way.

puntonim:
@ValueRaider thanks for the prompt reply 😄 ! I'll try  to explain better.

Let's consider this:
```py
import requests
import yfinance as yf
from pandas import DataFrame

def get_prices(symbol, timeout=(10, 10)):
    ticker = yf.Ticker(symbol)
    data: DataFrame = ticker.history(
        period="5d",
        interval="1d",
        timeout=timeout,
    )
    return data

# Network timeout (simulated by using a really short read-timeout).
prices = get_prices("AAPL", timeout=(10, 0.01))
assert prices.empty

# Non-existent symbol.
prices = get_prices("XXXAAPLXXX")
assert prices.empty
```

The [current code](https://github.com/ranaroussi/yfinance/blob/8975689bd19f96ea4655688611b0d853822eb5ec/yfinance/base.py#L218) swallows any exception that `YfData.get()` might raise, including all [Requests' exception](https://requests.readthedocs.io/en/latest/api/#exceptions).
And it returns an empty DataFrame.
So it is impossible for the consumer code to tell if the empty DataFrame is due to a (read or connect) timeout, a DNS failure, refused connection, any other network issue at client or Yahoo-side, JSON-encoding error, etc. In all these cases an empty DataFrame is returned exactly like the "legit" case when the symbol does not exist or there is no data for the requested period/interval.
My use case is exactly this: handling network errors.

With my proposed code change this would be possible:

```py
def reraise(exc):
    raise exc

def get_prices(symbol, timeout=(10, 10)):
    ticker = yf.Ticker(symbol)
    data: DataFrame = ticker.history(
        period="5d",
        interval="1d",
        timeout=timeout,
        request_exc_hook=reraise,
    )
    return data

# Network timeout (simulated by using a really short read-timeout).
try:
    prices = get_prices("AAPL", timeout=(10, 0.01))
except requests.Timeout:
    print("Timeout, retrying one more time...")
    ...
```

My proposed code change is backward compatible and the new feature is disabled by default, so no change is required by current users, unless they want to use the new feature.

I see 2 possible alternatives that do not require introducing the `request_exc_hook` arg:
 - expecting all "reasonable" exceptions (like network errors) and raising custom exceptions: to me the best one, but it's not the style in this repo and it would not be backward compatible.
 - do not swallow exceptions but let them bubble-up, so to let the consumer code decide if ignoring them or not: it would not be backward compatible.

ValueRaider:
Seems best fix is much simpler - add `raise_errors` logic to that swallowed exception. Ctrl+F "raise_errors"

puntonim:
@ValueRaider done 😉!

ValueRaider:
Squash commits then rebase to `dev`, guide here #1084

puntonim:
> Squash commits then rebase to `dev`, guide here #1084

@ValueRaider 👍

PR #1823, opened 2024-01-12
Update README.md for better copy-ability

molpcs:
Wrap yfinance[optional] code snippet with quotes to avoid conflict with zsh globbing. Remains compatible with bash.


ValueRaider:
This can go straight into main but then probably you need to rebase. Your decision.

molpcs:
👍 let's keep it going into dev.

PR #1828, opened 2024-01-15
Add type info.

dementedhedgehog:
Is this what people want.. r.e type info?

feature/growth-estimate removes some unimplemented fields from Analysis do we want to remove this stuff all the way up to the tests?, or is this info supposed to come from somewhere else and we should leave the hooks?

```
         self._earnings_trend = None
        self._analyst_trend_details = None
        self._analyst_price_target = None
        self._rev_est = None
        self._eps_est = None
```

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -1988,42 +1988,42 @@ class TickerBase:
             return data.to_dict()
         return data
 
-    def get_earnings_estimate(self, proxy=None, as_dict=False):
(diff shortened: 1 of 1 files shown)


ValueRaider:
Looks more like renaming https://github.com/ranaroussi/yfinance/pull/1778/files#diff-87ec0a62e5cd9f9570c4396b107fa8cb014070a905504e50e80773ecca4902ea

PR #1844, opened 2024-01-29
adding upgrade for pandas deprecation warning, adding pyarrow>=0.17.0…

power-edge:
… at minimum requirement as defined in dev requirements for pandas==1.3.0 version

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,4 +9,3 @@ frozendict>=2.3.4
 beautifulsoup4>=4.11.1
 html5lib>=1.1
 peewee>=3.16.2
-pyarrow>=0.17.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ frozendict>=2.3.4
 beautifulsoup4>=4.11.1
 html5lib>=1.1
 peewee>=3.16.2
+pyarrow>=0.17.0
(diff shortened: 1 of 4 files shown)


power-edge:
trying to get rid of a pandas deprecation warning. fixed up some tests that weren't working. adding requirements_dev for us (based on what else I was required to install to run tests)

ValueRaider:
Why is pyarrow needed?

power-edge:
When I was running test, I saw this warning without pyarrow installed. This PR was to remove annoying pandas warnings, wdyt? I have removed it from now. I also dropped the requirements_dev, on closer inspection those requirements were covered by the extras already.

Here is the warning I saw when running tests in some new virtual envs with project just installed.

```
./yfinance/yfinance/ticker.py:27: DeprecationWarning: 
  Pyarrow will become a required dependency of pandas in the next major release of pandas (pandas 3.0),
  (to allow more performant data types, such as the Arrow string type, and better interoperability with other libraries)
  but was not found to be installed on your system.
  If this would cause problems for you,
  please provide us feedback at https://github.com/pandas-dev/pandas/issues/54466
          
    import pandas as _pd

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```
image

FYI I have tested these changes while trying to keep requirements down to a minimum version (replace `>=` with `~=`) and the tests were passing in this state the same as with `>=`

ValueRaider:
This one should be left for user & Pandas to handle.

power-edge:
it is removed

PR #1872, opened 2024-02-27
Add scrape of ETF top/major holdings

dotJoel:
This PR is a a completion of the work started in https://github.com/ranaroussi/yfinance/pull/1571

To address the issue with returning price information when checking the holdings of a non-fund ticker, it does the following:
1. check for the presence of a redirect away from `/holdings` to identify when they are n/a.
2. return empty list in that case


ValueRaider:
Can it be converted to a DataFrame? More usable. Use `get_earnings_dates` for inspiration.
- symbol as index
- convert `% Assets` to actual number

Then add a unit test, in case someone in future breaks your code.

dotJoel:
It seems that this data is no longer a table on that page?
Top holdings that were a table now seem to be divs.
I swear this worked last week...

I'd rather help with the quoteSummary than parse more html.

ValueRaider:
Does the webpage fetch from API? Firefox -> F11 -> Network -> refresh

dotJoel:
Not directly.
There is a JS script block that has the data that comes from the API.
Seems to be the same API endpoint that the Fund data PR is using.

ValueRaider:
btw your commit history is a mess, I can't merge that.

ValueRaider:
I've fixed formatting but only tested on `IVV`. Can people try more ETFs?

PR #1915, opened 2024-04-23
Add error classes for symbol delisting errors, closes #270

### LINKED ISSUE #270: [FEATURE REQUEST] allow supression of "delisted" tickers
when using `yf.download`, have a keyword parameter that will allow us to hide "symbol may be delisted" errors among other errors.
## proposal 
Using `hide_errors=set()` or `suppress_errors=set()` you should check if the error has been specified to be suppressed and then output only if it has not been suppressed

elibroftw:
You will need to bump version to `0.3.0` as this might be a breaking change for some people who string comparing the previous error message.


yfinance/exceptions.py
```diff
@@ -1,12 +1,49 @@
 class YFinanceException(Exception):
```
ValueRaider, line 1: Can we shorten all instances of "YFinance" in exception class names to "YF"?
Can make other shortenings eg:
- timezone -> TZ
- priceData -> prices

ValueRaider (commented):
1 suggestion.
Also, rebase to dev branch #1084

elibroftw:
Going to reopen a clean PR

PR #1922, opened 2024-04-28
Fix: datetime.datetime.utcnow() is deprecated ...

ValueRaider:
I upgraded Python 3.11 -> 3.12 and get this new warning:

> DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).

Simple to fix.


ValueRaider:
Problem: this breaks yfinance if Python older than 3.11.

PR #1955, opened 2024-06-01
Fix 'T' is deprecated

lp177:
site-packages/yfinance/scrapers/history.py:239: FutureWarning: 'T' is deprecated and will be removed in a future version, please use 'min' instead.

### DIFF
--- a/yfinance/scrapers/history.py
+++ b/yfinance/scrapers/history.py
@@ -236,7 +236,7 @@ class PriceHistory:
         # 2) fix weired bug with Yahoo! - returning 60m for 30m bars
         if interval.lower() == "30m":
             logger.debug(f'{self.ticker}: resampling 30m OHLC from 15m')
-            quotes2 = quotes.resample('30T')
+            quotes2 = quotes.resample('30min')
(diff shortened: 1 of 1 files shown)


ValueRaider:
Does pandas 1.3.0 support change? That's our minimum version.

lp177:
@ValueRaider This seems supported, I've test with this example on pandas 1.3.0 and python 3.9.19 :
```python
import pandas as pd
import numpy as np

date_range = pd.date_range(start='2023-01-01 00:00:00', end='2023-01-01 23:59:59', freq='S')
temperature_data = np.random.normal(loc=20, scale=5, size=len(date_range))
nan_indices = np.random.choice(len(date_range), size=100, replace=False)
temperature_data[nan_indices] = np.nan
df = pd.DataFrame({'Date': date_range, 'Temperature': temperature_data})
df.set_index('Date', inplace=True)
df['Temperature'].fillna(method='ffill', inplace=True)
minute_avg_temperature = df.resample('Min').mean()
print(minute_avg_temperature)
```

PR #1959, opened 2024-06-09
Sustainability Fetching Implementation

MohamedAlaa201:
This PR does the following: 

- implements the fetching of the sustainability data 
- Enables tests related to sustainability fetching


MohamedAlaa201:
> Can you squash commits? Looks like you've been trying but the history still there.

Done.

Issue #1916, opened 2024-04-24
Limited period auto adjust with "max" period value


LoGo124:
### Describe bug

In some cases you're interested to download all data u can download  in a specific interval. 
I suppose the correct method, is use the "max" period option, but it only works with "1m" value as interval.

### Simple code that reproduces your problem

```
import yfinance as yf

ticker = yf.Ticker("BTC-USD")

print(ticker.history(interval = "1m", period = "max"))
print(ticker.history(interval = "1h", period = "max"))
```

### Debug log

```
                                   Open          High           Low         Close    Volume  Dividends  Stock Splits
Datetime                                                                                                            
2024-04-17 11:16:00+00:00  63106.847656  63106.847656  63106.847656  63106.847656         0        0.0           0.0
2024-04-17 11:17:00+00:00  63181.269531  63181.269531  63181.269531  63181.269531         0        0.0           0.0
2024-04-17 11:18:00+00:00  63179.203125  63179.203125  63179.203125  63179.203125   1630208        0.0           0.0
2024-04-17 11:19:00+00:00  63152.871094  63152.871094  63152.871094  63152.871094   4526080        0.0           0.0
2024-04-17 11:20:00+00:00  63122.992188  63122.992188  63122.992188  63122.992188   8736768        0.0           0.0
...                                 ...           ...           ...           ...       ...        ...           ...
2024-04-24 11:08:00+00:00  66438.062500  66438.062500  66438.062500  66438.062500         0        0.0           0.0
2024-04-24 11:09:00+00:00  66430.968750  66430.968750  66430.968750  66430.968750  61263872        0.0           0.0
2024-04-24 11:10:00+00:00  66431.734375  66431.734375  66431.734375  66431.734375         0        0.0           0.0
2024-04-24 11:11:00+00:00  66432.492188  66432.492188  66432.492188  66432.492188         0        0.0           0.0
2024-04-24 11:12:00+00:00  66437.500000  66437.500000  66437.500000  66437.500000         0        0.0           0.0

[8740 rows x 7 columns]
BTC-USD: 1h data not available for startTime=1410912000 and endTime=1713957318. The requested range must be within the last 730 days.
Empty DataFrame
Columns: [Open, High, Low, Close, Adj Close, Volume]
Index: []
```

### `yfinance` version

0.2.38

### Python version

3.10.12

### Operating system

Pop!_OS 22.04 LTS

Terry-50000:
How can i get 5m interval data of over 60 days?

PR #1967, opened 2024-06-25
fixed history fetching when period='max'

SnowCheetos:
#1916 Fixed limited period auto adjust with "max" period value

accidentally misspelled 'period' for branch name...

### DIFF
--- a/yfinance/scrapers/history.py
+++ b/yfinance/scrapers/history.py
@@ -98,7 +98,19 @@ class PriceHistory:
             if start is None:
                 if interval == "1m":
                     start = end - 604800  # Subtract 7 days
+                
+                # 1 hour interval supports up to 730 days
+                elif interval == "1h":
+                    max_start_datetime = pd.Timestamp.utcnow().floor("D") - _datetime.timedelta(days=729)
+                    start = int(max_start_datetime.timestamp())
+                
+                # 5m to 90m interval supports up to 60 days
+                elif interval in ["5m", "15m", "30m", "60m", "90m"]:
+                    max_start_datetime = pd.Timestamp.utcnow().floor("D") - _datetime.timedelta(days=59)
+                    start = int(max_start_datetime.timestamp())
+
                 else:
+                    # 99 years is not available for intervals less than 1d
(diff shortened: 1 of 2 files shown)


ValueRaider:
I haven't tested but 2 considerations:
- floor vs ceil
- UTC vs exchange timezone

yfinance/scrapers/history.py
```diff
@@ -98,6 +99,14 @@ def history(self, period="1mo", interval="1d",
...
+                
+                elif interval == "1h":
+                    start = end - 63072000 # Subtract 730 days
+                
+                # 5m to 90m interval supports up to 60 days
+                elif interval in ("5m", "15m", "30m", "60m", "90m"):
+                    start = end - 5184000 # Subtract 60 days
+
                 else:
                     max_start_datetime = pd.Timestamp.utcnow().floor("D") - _datetime.timedelta(days=99 * 365)
```
SnowCheetos, line 111: This was the current implementation on the main branch, left it as is

yfinance/scrapers/history.py
```diff
@@ -98,6 +99,14 @@ def history(self, period="1mo", interval="1d",
             if start is None:
                 if interval == "1m":
                     start = end - 604800  # Subtract 7 days
+                
+                elif interval == "1h":
```
SnowCheetos, line 103: 1h can fetch up to 730 days of data, subtracted 730 * 24 * 60 * 60 = 63,072,000 seconds

yfinance/scrapers/history.py
```diff
@@ -98,6 +99,14 @@ def history(self, period="1mo", interval="1d",
...
+                    start = end - 63072000 # Subtract 730 days
+                
+                # 5m to 90m interval supports up to 60 days
+                elif interval in ("5m", "15m", "30m", "60m", "90m"):
```
SnowCheetos, line 107: I tested these intervals with the main branch, and they all support up to 60 days, including 60m interval which should be equivalent to 1h, unless there is another reason for it?

ValueRaider: Last I checked, Yahoo treats `1h` different, which could be different to `60m`.

ValueRaider: 60m is equivalent to 1h, so I've fixed.

PR #1995, opened 2024-07-19
Fix error on empty options chain

### LINKED ISSUE #1969: option_chain() function is returning error while calling it.
KeyError                                  Traceback (most recent call last)
Cell In[31], line 1
----> 1 yca.option_chain()

File ~/anaconda3/envs/prepJob/lib/python3.11/site-packages/yfinance/ticker.py:98, in Ticker.option_chain(self, date, tz)
     94     date = self._expirations[date]
     95     options = self._download_options(date)
     97 return _namedtuple('Options', ['calls', 'puts', 'underlying'])(**{
---> 98     "calls": self._options2df(options['calls'], tz=tz),
     99     "puts": self._options2df(options['puts'], tz=tz),
    100     "underlying": options['underlying']
    101 })

KeyError: 'calls'

stevenbischoff:
Fixes #1969 

Error occurs when Yahoo has no calls or puts for a ticker. I wasn't sure whether y'all would prefer an exception in this case instead (first pull request!), but I chose to return an "empty" Options namedtuple since _download_options() returns an empty dict.

### DIFF
--- a/yfinance/ticker.py
+++ b/yfinance/ticker.py
@@ -92,6 +92,11 @@ class Ticker(TickerBase):
             date = self._expirations[date]
             options = self._download_options(date)
 
+        if not options:
+            return _namedtuple('Options', ['calls', 'puts', 'underlying'])(**{
+                "calls": None, "puts": None, "underlying": None
+            })
+
         return _namedtuple('Options', ['calls', 'puts', 'underlying'])(**{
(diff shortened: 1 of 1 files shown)


ValueRaider:
I don't use options so no opinion. Try tagging the few people that use options (search here).

BiggRanger:
Returning empty works perfectly for me. I'd rather check for nothing than have to handle an exception.

PR #2008, opened 2024-08-03
Update README.md

jpcofr:
Removed a member from the msft object causing the example to fail.


ValueRaider:
It's just a typo, don't remove it.

Issue #2007, opened 2024-08-02
Request: Support for SEC filings module


rgeronimi:
_Originally posted by @dmoklaf in https://github.com/ranaroussi/yfinance/discussions/2004_

It seems a stock's SEC filings can be retrieved through the `secFilings`module:
```
filings = quote._fetch(  # pylint: disable=W0212
    quote.proxy, modules=["secFilings"]
)
if filings is not None:
    filings = filings["quoteSummary"]["result"][0]["secFilings"]["filings"]
    for filing in filings:
        url = filing["edgarUrl"]
        type_ = filing["type"]
        title = filing["title"]
        timestamp = filing["epochDate"]
        date = datetime.datetime.fromtimestamp(timestamp, tz=datetime.UTC)
        symbols = filing.get("relatedTickers")
        # Do something with the filing...
```

Would it be possible to include this as a new feature within `yfinance`?

PR #2009, opened 2024-08-03
Implement fetch sec-filings

ValueRaider:
Implement #2007 . @dmoklaf give feedback - can structure be improved more? How do you use the data?


dmoklaf:
I use these  exactly like news - to suggest recent changes that could explain sudden moves in the stock price. Thus what matters are primarily the date, type of form / title, and URL to get more details (in this case, the EDGAR URL).

ValueRaider:
Try it #1080

dmoklaf:
Works perfectly ! Bravo

PR #2010, opened 2024-08-03
Fix reading timezone from earning dates

mreiche:
Reading earning dates returns index timestamps without timezone. That's been fixed.


ValueRaider:
But it already has a timezone. How did you determine it was missing?

mreiche:
Hmm... Maybe I was working on an outdated version. Let me check and get back to this PR if necessary.

mreiche:
I improved the test for the bug fix

tests/test_ticker.py
```diff
@@ -323,6 +325,15 @@ def test_earnings_dates_with_limit(self):
     #     data_cached = self.ticker.earnings_trend
     #     self.assertIs(data, data_cached, "data not cached")
 
+    def test_ticker_has_tz(self):
+        test_data = {"O2D.DE": "UTC", "LHA.DE": "Europe/Berlin"}
```
ValueRaider, line 329: I can't see O2D.DE on Yahoo - is it real?

yfinance/base.py
```diff
@@ -589,20 +599,20 @@ def get_earnings_dates(self, limit=12, proxy=None) -> Optional[pd.DataFrame]:
...
 
+        def map_date(time_str: str):
+            tz_match = re.search('([AP]M)([a-zA-Z]*)$', time_str)
+            tz_str = tz_match.group(2).strip()
+            # - remove AM/PM and timezone from date string
+            time_str = time_str.replace(tz_str, "")
+            try:
+                tz = pytz.timezone(tz_str)
+            except pytz.UnknownTimeZoneError:
+                tz = get_ticker_tz()
```
ValueRaider, line 611: If `get_ticker_tz()` is returning None then the ticker isn't listed. So the returned earnings dates for legitimate tickers should already be localized.

ValueRaider (changes requested):

tests/test_ticker.py
mreiche: It may be delisted now, but returned earning dates with UTC timestamp. Let me check this and find another test related symbol.

yfinance/base.py
mreiche: As far as I remember, timezones like „EDT“ are unknown to Python. Let me point that out in a dedicated test.

ValueRaider: Exactly, that's why you shouldn't use `tz_str`.

mreiche: But the timezones from the earnings list can differ from the ticker itself. I would prefer them and use the general ticker tz as fallback.

ValueRaider: Now I understand your position. I've pushed a commit to improve your proposal, easier than discussing here, can always rollback.

ValueRaider (approved):

PR #2016, opened 2024-08-07
Fix datetime conversion with mixed timezones when ignore_tz is False

mreiche:
When querying multi symbols at the same time, the Pandas Timestamp to datetime64 conversion will fail, when `utc=False`.
It looks like this is the most stable implementation, because Pandas suggests to use this feature in the future.
Downside: The original timezones are lost. But every datetime has now `tz=timezone.utc`
Upside: `ignore_tz` may be obsolete.

I added a test for testing Pandas datetime conversion as well.

### DIFF
--- a/tests/test_prices.py
+++ b/tests/test_prices.py
@@ -1,13 +1,13 @@
-from .context import yfinance as yf
-from .context import session_gbl
-
+import datetime as _dt
+import os
 import unittest
 
-import os
-import datetime as _dt
-import pytz as _tz
 import numpy as _np
 import pandas as _pd
+import pytz as _tz
+
+from .context import session_gbl
+from .context import yfinance as yf
 
 
 class TestPriceHistory(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


tests/test_prices.py
```diff
@@ -1,13 +1,14 @@
-from .context import yfinance as yf
-from .context import session_gbl
-
+import datetime
+import datetime as _dt
```
ValueRaider, line 2: Pick one.

ValueRaider, line 2: Pick one

ValueRaider (changes requested):

ValueRaider:
> Downside: The original timezones are lost. But every datetime has now `tz=timezone.utc`

That's fine as long as 9:30am EST is aligned with 2:30pm EU (or whatever the Atlantic offset is). `ignore_tz` means 9:30am EST aligned with 9:30am EU.

mreiche:
> > Downside: The original timezones are lost. But every datetime has now `tz=timezone.utc`
> 
> That's fine as long as 9:30am EST is aligned with 2:30pm EU (or whatever the Atlantic offset is). `ignore_tz` means 9:30am EST aligned with 9:30am EU.

That's covered by the test `PandasTest.test_mixed_timezones_to_datetime()` where the `datetime64` utc timestamps are compared to the tz-aware `Timestamps`.

ValueRaider (approved):

ValueRaider:
Approved. Just squash the commits #1084

mreiche:
> Approved. Just squash the commits #1084

Can't you just enable squash commits in your repo? https://docs.github.com/de/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

ValueRaider:
Hmm. Could, but then the squashed commit message lists each individual commit e.g. `Add assertion for timezone`. Why would you not want to rewrite the message?

mreiche:
> Hmm. Could, but then the squashed commit message lists each individual commit e.g. `Add assertion for timezone`. Why would you not want to rewrite the message?

It's not something I usually do manually. I have to go back, rebase, commit everything again, force push. I don't even know how to do that. I need to read documentation, have several tries and so on. It just takes time. In Github, you can configure, if the MR title should be taken as Squash commit message. That takes less than 10 seconds.

Issue #2015, opened 2024-08-07
.news will raise JSONDecodeError 


cyc00518:
### Describe bug

When I use `.news`, it raise `JSONDecodeError`.

But it worked well before. 

### Simple code that reproduces your problem

```
import yfinance as yf
ticker = yf.Ticker('SPY')
print(ticker.news)
```

### Debug log

```
---------------------------------------------------------------------------
JSONDecodeError                           Traceback (most recent call last)
File ~/.local/lib/python3.8/site-packages/requests/models.py:971, in Response.json(self, **kwargs)
    970 try:
--> 971     return complexjson.loads(self.text, **kwargs)
    972 except JSONDecodeError as e:
    973     # Catch JSON-related errors and raise as requests.JSONDecodeError
    974     # This aliases json.JSONDecodeError and simplejson.JSONDecodeError

File /opt/conda/lib/python3.8/json/__init__.py:357, in loads(s, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    354 if (cls is None and object_hook is None and
    355         parse_int is None and parse_float is None and
    356         parse_constant is None and object_pairs_hook is None and not kw):
--> 357     return _default_decoder.decode(s)
    358 if cls is None:

File /opt/conda/lib/python3.8/json/decoder.py:337, in JSONDecoder.decode(self, s, _w)
    333 """Return the Python representation of ``s`` (a ``str`` instance
    334 containing a JSON document).
    335 
    336 """
--> 337 obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    338 end = _w(s, end).end()

File /opt/conda/lib/python3.8/json/decoder.py:355, in JSONDecoder.raw_decode(self, s, idx)
    354 except StopIteration as err:
--> 355     raise JSONDecodeError("Expecting value", s, err.value) from None
    356 return obj, end

JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

JSONDecodeError                           Traceback (most recent call last)
Input In [16], in <cell line: 3>()
      1 import yfinance as yf
      2 ticker = yf.Ticker('SPY')
----> 3 print(ticker.news)

File ~/.local/lib/python3.8/site-packages/yfinance/ticker.py:262, in Ticker.news(self)
    260 @property
    261 def news(self) -> list:
--> 262     return self.get_news()

[...]

PR #2021, opened 2024-08-10
Handle faulty response object when getting news.

ericpien:
- Fixes: https://github.com/ranaroussi/yfinance/issues/2015
  - The trace provided in the issue shows the exception occurs due to trying to parse faulty response object `JSONDecodeError: Expecting value: line 1 column 1 (char 0)`
  - try except catches the error and returns empty set instead of raising the exception. 
  - The error is logged and the API user sees empty list of news instead of the raised exception.
- Add check if returned value is None before accessing its text to avoid AttributeError

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -512,11 +512,16 @@ class TickerBase:
         # Getting data from json
         url = f"{_BASE_URL_}/v1/finance/search?q={self.ticker}"
         data = self._data.cache_get(url=url, proxy=proxy)
-        if "Will be right back" in data.text:
+        if data is None or "Will be right back" in data.text:
             raise RuntimeError("*** YAHOO! FINANCE IS CURRENTLY DOWN! ***\n"
                                "Our engineers are working quickly to resolve "
                                "the issue. Thank you for your patience.")
-        data = data.json()
+        try:
+            data = data.json()
+        except (_json.JSONDecodeError): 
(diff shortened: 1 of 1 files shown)


yfinance/base.py
```diff
@@ -512,11 +512,16 @@ def get_news(self, proxy=None) -> list:
...
+        if data is None or "Will be right back" in data.text:
             raise RuntimeError("*** YAHOO! FINANCE IS CURRENTLY DOWN! ***\n"
                                "Our engineers are working quickly to resolve "
                                "the issue. Thank you for your patience.")
-        data = data.json()
+        try:
+            data = data.json()
+        except (_json.JSONDecodeError): 
+            logger = utils.get_yf_logger()
+            logger.error(f"{self.ticker}: Failed to retireve the news and received faulty response instead.")
```
ValueRaider, line 523: `retireve` typo

ValueRaider (changes requested):

yfinance/base.py
ericpien: Thanks. Addressed in: https://github.com/ranaroussi/yfinance/pull/2021/commits/3574cf4bdc7706ce0dcf87cd4f67b16e781c2217

ValueRaider (approved):

ValueRaider:
Approved, just do a git squash #1084

Issue #2026, opened 2024-08-15
Fix 30min interval of ^NSEI


aleksfasting:
# Fix 30min interval of ^NSEI and indian exchange

Regarding #1436 and #1447 

### Summary

The Indian stock exchange opens at 09.15, so when requesting the history of '^NSEI' by 30 minute intervals, the resampling will not include the first 15 minutes of the day. ***Note:** This only happens to 30 min intervals because of the specific handling of 30 min.*

-------------------

### Solution

@ivan23kor created a pull request some time ago to solve the issue, but the solution is now outdated.

I propose a solution that checks what time the exchange opens and adds that to the offset that ivan suggested:

```
exchangeStartTime = pd.Timestamp(self._history_metadata["tradingPeriods"][0][0]["start"], unit='s')
offset = str(exchangeStartTime.minute % 30)+"min"
quotes2 = quotes.resample('30min', offset=offset)
```

aleksfasting:
#2027 was merged. Issue is closed

PR #2027, opened 2024-08-15
Solves #2026 

aleksfasting:
# Solves #2026 to fix interval offset of the indian stock exchange

Solution that checks the offset of the opening time of the stock exchange (mod 30 minutes) and adds it to the quotes

That issue addresses  #1436 and #1447.

-----------

### Solution

The only change lies in `yfinance/scrapers/history.py`. 

If ìnterval='30m'`, we find the time that the stock exchange opens. If it opens at a time `t.minutes % 30 != 0` we add that offset to the quote.

### DIFF
--- a/yfinance/scrapers/history.py
+++ b/yfinance/scrapers/history.py
@@ -227,7 +227,9 @@ class PriceHistory:
         # 2) fix weired bug with Yahoo! - returning 60m for 30m bars
         if interval.lower() == "30m":
             logger.debug(f'{self.ticker}: resampling 30m OHLC from 15m')
-            quotes2 = quotes.resample('30min')
+            exchangeStartTime = pd.Timestamp(self._history_metadata["tradingPeriods"][0][0]["start"], unit='s')
+            offset = str(exchangeStartTime.minute % 30)+"min"
+            quotes2 = quotes.resample('30min', offset=offset)
             quotes = pd.DataFrame(index=quotes2.last().index, data={
                 'Open': quotes2['Open'].first(),
                 'High': quotes2['High'].max(),


ValueRaider:
Hypothetical - can `self._history_metadata["tradingPeriods"][0][0]["start"]` throw a `KeyError` or `IndexError`?

aleksfasting:
### The `'tradingPeriods'` key
The data from Yahoo Finance will always come with metadata, and it will include the `'tradingPeriods'` key as long as there was a ticker to be found. If the ticker wasn't found, that would have been handled before the call to this line.

### The first 0 index

The entries of `self._histoy_metadata['tradingPeriods']` will be a list of all the opening-times, end-times, etc. of each day succeding the start date.

If the start date is after the current date there will not have been generated any price data, and the code has handled this before this line of code. *The same goes for any scenario where no price data has been generated, like weekends, holidays or such.*

### The second 0 index

I have no idea why Yahoo Finance supply their metadata with this wrapper list. The list always contains a single `dict`. Artistic choice by Yahoo finance. I could not find a counter example.

### The `'start'` index

This is a key in the standard dict in `self._history_metadata["tradingPeriods][0][0]`. I have not found a dict that does not contain this key.

----------

So i think the keys and indexes will always exist if this line is being executed. **No exceptions will be thrown**

ValueRaider:
The exception was possible, I encountered it:
```
    exchangeStartTime = pd.Timestamp(self._history_metadata["tradingPeriods"][0][0]["start"], unit='s')
                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^

KeyError: 0
```

Please submit a patch.

ValueRaider:
Reproduce:

```
dat = yf.Ticker('0001.HK')
df = dat.history(start=1719763200, end=1720540800, interval='30m', prepost=True)
```

aleksfasting:
Shall I create another PR for the patch?

PR #2029, opened 2024-08-19
Patch keyError in #2027

aleksfasting:
Solves ´KeyError´ occurrence in #2027.

Adds Try/Except statement to find keyerror and return to old code if the keyerror is triggered.


ValueRaider:
That's not the solution. Look at the metadata.

aleksfasting:
The same information as in the original PR, #2026, can be fetched from the `'currentTradingPeriod'` key in the metadata. The metadata from the code that triggered the KeyError,

```
dat = yf.Ticker('0001.HK')
dat.history(start=1719763200, end=1720540800, interval='30m', prepost=True)
```

will have the same format as most other stocks.

I don't know if Yahoo Finance may give differently formated metadata for the ´'currentTradingPeriod'´ key.

ValueRaider:
Just add 2 unit tests: one with prepost=False, other with True. Then rebase to dev branch #1084

Issue #2093, opened 2024-10-21
f-string: unmatched '(' when importing yfinance


FX196:
### Describe bug

Syntax error in screener with Python 3.10. Error is coming from https://github.com/ranaroussi/yfinance/blob/main/yfinance/screener/screener.py#L101

### Simple code that reproduces your problem

```python
import yfinance
```
using Python 3.10

### Debug log
No debug logs but have stack trace of the import error
```
Traceback (most recent call last):
  File "/home/runner/work/.../libs.py", line 2, in <module>
    import yfinance as yf
  File "/home/runner/work/.../venv/lib/python3.10/site-packages/yfinance/__init__.py", line 30, in <module>
    from .screener.screener import Screener
  File "/home/runner/work/.../venv/lib/python3.10/site-packages/yfinance/screener/__init__.py", line 1, in <module>
    from .screener import Screener
  File "/home/runner/work/.../venv/lib/python3.10/site-packages/yfinance/screener/screener.py", line 101
    logger.error(f"Failed to get screener data for '{self._body.get('query', "query not set")}' reason: {e}")
                                                                              ^^^^^
SyntaxError: f-string: unmatched '('
Error: Process completed with exit code 1.
```

### `yfinance` version

0.2.45

### Python version

3.10

sing-tam:
same issue here

mrimahajan:
Can we take some action on this asap. This is impacting the daily production jobs for us.

FX196:
> Can we take some action on this asap. This is impacting the daily production jobs for us.

@mrimahajan you can pin your version to the previous release by changing your `requirements.txt` to say `yfinance==0.2.44`. That's what I did to fix mine

ValueRaider:
I'm confused how
> from .screener import Screener

is causing `Screener._fetch_and_parse()` to execute. I can't reproduce.

@ericpien

fled-dev:
Same issue

**System Details**
- MacOS 15.1
- Python 3.11.9

abdulmuhaiminmk:
> > Can we take some action on this asap. This is impacting the daily production jobs for us.
> 
> @mrimahajan you can pin your version to the previous release by changing your `requirements.txt` to say `yfinance==0.2.44`. That's what I did to fix mine

I solved the issues using jupyter notebook by pinned the previous version upon installing

'''!pip3 install yfinance==0.2.44'''

abdulmuhaiminmk:
[...]

PR #2094, opened 2024-10-21
Fix malformed f-string in release `0.2.45`

FX196:
Quick fix for issue https://github.com/ranaroussi/yfinance/issues/2093
This broke yfinance for me after I automatically upgraded to the latest release.

Existing tests should suffice. Directly targeting `main` because the release is broken

### DIFF
--- a/yfinance/screener/screener.py
+++ b/yfinance/screener/screener.py
@@ -98,7 +98,7 @@ class Screener:
             self._response = response['finance']['result'][0]
         except Exception as e:
             logger = utils.get_yf_logger()
-            logger.error(f"Failed to get screener data for '{self._body.get('query', "query not set")}' reason: {e}")
+            logger.error(f"Failed to get screener data for '{self._body.get('query', 'query not set')}' reason: {e}")
             logger.debug("Got response: ")
             logger.debug("-------------")
             logger.debug(f" {response}")


sushrutkhandelwal (approved):

fled-dev:
Just tried it on my system (macOS 15.1 / Python 3.11.9) and it works like a charm. Thanks for putting in the effort to address this issue.

DavidVilem:
Hello good morning.

I am testing from databricks and it generates the same error:

Syntax error: string f: mismatch '(' (screener.py, line 101)

vwbusguy:
I just tested against an image environment that failed tests from this error and this patch works:

```
pip install --force  git+https://github.com/FX196/yfinance@patch-1
python -c 'import yfinance'
```

Without the patch on the same image environment:

```
python -c 'import yfinance'
  File "<string>", line 1, in <module>
  File "/opt/conda/lib/python3.11/site-packages/yfinance/__init__.py", line 30, in <module>
    from .screener.screener import Screener
  File "/opt/conda/lib/python3.11/site-packages/yfinance/screener/__init__.py", line 1, in <module>
    from .screener import Screener
  File "/opt/conda/lib/python3.11/site-packages/yfinance/screener/screener.py", line 101
    logger.error(f"Failed to get screener data for '{self._body.get('query', "query not set")}' reason: {e}")
                                                                              ^^^^^
SyntaxError: f-string: unmatched '('
```

vwbusguy (approved):

PR #2096, opened 2024-10-21
Fix syntax error in logger error message in _fetch_and_parse method

shkabrii:
Fixed a syntax error in the `_fetch_and_parse` method. The logger.error call had mixed single and double quotes, which caused a syntax issue.

Updated the line to:
```python
logger.error(f"Failed to get screener data for '{self._body.get('query', 'query not set')}' reason: {e}")


shkabrii:
Closing this pull request as it's a duplicate of #2094, which addresses the same issue

PR #2099, opened 2024-10-25
fixed a bug throwing an error when trying to get a ticker from certain isin

DamienDuv:
When making a request to "{_BASE_URL_}/v1/finance/search?q={isin}", the response doesn't always has a 'longname' field and this throws an error.
Example with this isin : IE00BDBRDM35


ValueRaider:
Rebase to `dev` will tidy your commit history #1084

PR #2120, opened 2024-11-09
Improve Sphinx docs

ValueRaider:
Reorganise new documentation, add some missing content.

@ericpien can you sanity-check this? The changes are mostly AI - lots of warnings but seems to work, not sure how the Action will cope.


paulmcq (changes requested):
in doc/source/advanced/caching.rst line 48, suggest being a little less terse:
To reduce the number of requests to Yahoo servers, yfinance caches some data locally: timezones to localize dates, and the cookie & crumb. The cache location depends on the OS:

ericpien:
hey @ValueRaider I noticed this was converted to draft. Should I still review as is?

I see the vision in 1. Splitting out Ticker to underlying functions and 2. Changing the `class.rst` template but the implementation will take some work to rid of the warnings and unnecessary methods like `__dict__` and duplicate `__init__`.

ValueRaider:
I misuse draft state - we really need another state to mean ready for review but no merge.

All warnings fixed.

PR #2131, opened 2024-11-16
Massively simplify README by linking to new doc website

ValueRaider:
With the new doc website online, the README can be massively reduced, bringing its structure inline with other popular packages.

@ericpien @ranaroussi

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,23 +41,17 @@ Yahoo! finance API is intended for personal use only.**
 
 ---
 
-- [Installation](#installation)
-- [Quick start](#quick-start)
-- [Advanced](#logging)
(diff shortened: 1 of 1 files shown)


ericpien:
If we are planning on triggering the action on push to `main` in the future, `dev-documented` already has simplified [`README`](https://github.com/ranaroussi/yfinance/blob/dev-documented/README.md) that can we can merge into `main` after changing the `on-push-branches` to include `main`. Do you want to target that instead for the PR?

ValueRaider:
> dev-documented already has simplified README

I missed this.

`dev-documented` will merge into `dev` first, then into `main` for a release.

Issue #1932, opened 2024-05-14
earnings_date failing (KeyError: 'Earnings Date')


yunhongz:
### Describe bug

The earnings_dates record is not available

### Simple code that reproduces your problem

import yfinance as yf
msft = yf.Ticker('MSFT')
msft.earnings_dates

### Debug log

![image](https://github.com/ranaroussi/yfinance/assets/132848568/5e4cf65d-b1a8-4f3e-9af5-ee7376c42f82)

### `yfinance` version

yfinance 0.2.38

### Python version

python3.10

### Operating system

ubuntu

ValueRaider:
Yahoo has changed table structure

```
                                  Event Name  Earnings Call Time  EPS Estimate  Reported EPS  Surprise(%)
0                   Q1 2024 Earnings Release   Time Not Supplied           NaN           NaN          NaN
1                      Q1 2024 Earnings Call                 TAS           NaN           NaN          NaN
2                   Q1 2024 Earnings Release   Time Not Supplied           NaN           NaN          NaN
3   at Bank of America Healthcare Conference                 TAS           NaN           NaN          NaN
4                Annual Shareholders Meeting                 TAS          1.95          1.59      -0.1846
5                   Q1 2024 Earnings Release   Time Not Supplied           NaN           NaN          NaN
6                   Q3 2023 Earnings Release  After Market Close           NaN           NaN          NaN
7                   Q4 2023 Earnings Release  After Market Close           NaN           NaN          NaN
8   at Bank of America Healthcare Conference                 TAS         -0.74           NaN          NaN
9                   Q1 2024 Earnings Release   Time Not Supplied           NaN           NaN          NaN
10               Annual Shareholders Meeting   Time Not Supplied           NaN           NaN          NaN
11                  Q4 2023 Earnings Release  After Market Close           NaN           NaN          NaN
```

Event name is nice, but losing dates a problem. Hoping for a way to get old structure.

hjlgood:
Looks like yahoo fixed their problem. yfinance works fine now. I suggest you close this issue.

ValueRaider:
I suspect it will return, I've seen this alternative table randomly rarely over last few weeks. Yahoo is up to something...

michaelmoopenn:
For some tickers, it works and some others it doesn't. At this time, ENB.TO works, TSLA still gives the issue.

yunhongz:
It returned to normal, interesting!

hjlgood:
[...]

PR #2153, opened 2024-11-26
Add cookies as a workaround to fix earnings dates

ForeverZyh:
As discussed in #1932 , it seems we need to login to apply the filter, so the idea is to add additional cookies to the API. Usage:
```
yf.Ticker(symbol).get_earnings_dates(add_cookies={
        'T': Your_cookies1,
        'Y': Your_cookies2})
```
I tested that if we supply `T` and `Y` cookies to the `requests.get`, we can successfully get the desired response.

I asked ChatGPT to see how to get my cookies manually: https://chatgpt.com/share/67453e22-bba0-8013-9d97-b8d320b7e24f in Chrome.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -557,7 +557,7 @@ class TickerBase:
         return self._news
 
     @utils.log_indent_decorator
-    def get_earnings_dates(self, limit=12, proxy=None) -> Optional[pd.DataFrame]:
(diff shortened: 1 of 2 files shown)


ValueRaider:
`add_cookies` should be higher and modify `data.py`. At least up to the `Ticker` class, and maybe on `yfinance` itself.

And save it in cache via `YfData._save_cookie_basic` or a similar function, because manually extracting cookie sounds annoying.

Issue #2158, opened 2024-11-28
Change `Screener.set_predefined_body` and `Screener.set_body` to respond with the class


R5dan:
I want to change `Screener.set_predefined_body` and `Screener.set_body` to respond with the class.

This would reduce

```python
s = yf.Screener()
s.set_predefined_body("day_gainers")
r=s.response
```
to
```python
r = yf.Screener().set_predefined_body("day_gainers").response
```
or to be able to set it on initialisation
```python
r = yf.Screener(body="day_gainers").response
```

which I think is cleaner and more concise. I would also suggest editing the docs of `predefined_bodies` as they come above `set_predefined_body` to include how to set the Screener to them

ValueRaider:
@ericpien

Issue #2152, opened 2024-11-26
Screener


24thronin:
### Describe bug

Hello,
I have been trying to find a way to create a list of all securities on yahoo finance. I thought that screener with a custom query would be able to do this but the screener.response only returns up to 250 results. Is there a way to retrieve all values?
Thank you for the package.

### Simple code that reproduces your problem

import yfinance as yf
yf.enable_debug_mode()
r1 = yf.EquityQuery('eq', ['sector', 'Basic Materials'])
r2 = yf.EquityQuery('eq', ['region', 'us'])
crit = yf.EquityQuery('and', [r1,r2])
screener.set_default_body(crit,size=10000)
res=screener.response

### Debug log

DEBUG    Entering post()
DEBUG     Entering _make_request()
DEBUG      url=https://query2.finance.yahoo.com/v1/finance/screener
DEBUG      params={'corsDomain': 'finance.yahoo.com', 'formatted': 'false', 'lang': 'en-US', 'region': 'US'}
DEBUG      Entering _get_cookie_and_crumb()
DEBUG       cookie_mode = 'basic'
DEBUG       Entering _get_cookie_and_crumb_basic()
DEBUG        reusing cookie
DEBUG        reusing crumb
DEBUG       Exiting _get_cookie_and_crumb_basic()
DEBUG      Exiting _get_cookie_and_crumb()
DEBUG      response code=400
DEBUG      toggling cookie strategy basic -> csrf
DEBUG      Entering _get_cookie_and_crumb()
DEBUG       cookie_mode = 'csrf'
DEBUG       Entering _get_crumb_csrf()
DEBUG        Failed to find "csrfToken" in response
DEBUG       Exiting _get_crumb_csrf()
DEBUG       toggling cookie strategy csrf -> basic
DEBUG       Entering _get_cookie_and_crumb_basic()
DEBUG        loaded persistent cookie
DEBUG        reusing cookie
DEBUG        crumb = '5CmrIa/1bJ9'
DEBUG       Exiting _get_cookie_and_crumb_basic()
DEBUG      Exiting _get_cookie_and_crumb()
DEBUG      response code=400
DEBUG     Exiting _make_request()
DEBUG    Exiting post()
[...]

PR #2168, opened 2024-12-04
Update Screener

ericpien:
- Setting Body to Screener returns the Screener Object to enable method chaining.
    - i.e. `r = yf.Screener().set_predefined_body("day_gainers").response`
    - Addresses: https://github.com/ranaroussi/yfinance/issues/2158
    
- Limit query size to 250 and raise error if larger. This also avoids unnecessary network calls and makes the issue clearer.
    - Addresses: https://github.com/ranaroussi/yfinance/discussions/2165 and https://github.com/ranaroussi/yfinance/issues/2152#issuecomment-2499499627
    
- Unit test and doc updates for both changes

### DIFF
--- a/tests/test_screener.py
+++ b/tests/test_screener.py
@@ -13,7 +13,7 @@ class TestScreener(unittest.TestCase):
         self.query = EquityQuery('gt',['eodprice',3])
 
     def test_set_default_body(self):
-        self.screener.set_default_body(self.query)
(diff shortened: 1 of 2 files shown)


ValueRaider:
When you say `Setting Body to Screener returns the Screener object`, do you mean `Screener.set_predefined_body() returns the Screener object`?

PR #2172, opened 2024-12-08
Market summary

R5dan:
- New class for accessing the market summary of an area
    - As discussed in #2157 
    ```python
    import yfinance as yf
    summary = yf.MarketSummary(region)
    index = summary[index]
    status = index.status
    etc  
    ```
- New class for accessing data from `MarketSummary` to allow for types

PR #2175, opened 2024-12-10
Add market Summary

R5dan:
Supersedes #2172

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ Yahoo! finance API is intended for personal use only.**
 - `Ticker`: single ticker data
 - `Tickers`: multiple tickers' data
 - `download`: download market data for multiple tickers
+- `Market`: get infomation about a market
 - `Search`: quotes and news from search
 - `Sector` and `Industry`: sector and industry information
 - `EquityQuery` and `Screener`: build query to screen market
--- a/doc/source/reference/examples/market.py
+++ b/doc/source/reference/examples/market.py
@@ -1,0 +1,6 @@
+import yfinance as yf
+
+EUROPE = yf.Market("EUROPE")
+
+status = EUROPE.status
+summary = EUROPE.summary
--- a/doc/source/reference/index.rst
+++ b/doc/source/reference/index.rst
@@ -15,6 +15,7 @@ The following are the publicly available classes, and functions exposed by the `
 
 - :attr:`Ticker <yfinance.Ticker>`: Class for accessing single ticker data.
 - :attr:`Tickers <yfinance.Tickers>`: Class for handling multiple tickers.
+- :attr:`MarketSummary <yfinance.MarketSummary>`: Class for accessing market summary.
 - :attr:`Search <yfinance.Search>`: Class for accessing search results.
 - :attr:`Sector <yfinance.Sector>`: Domain class for accessing sector information.
 - :attr:`Industry <yfinance.Industry>`: Domain class for accessing industry information.
@@ -33,6 +34,7 @@ The following are the publicly available classes, and functions exposed by the `
    yfinance.stock
    yfinance.financials
    yfinance.analysis
+   yfinance.marketsummary
    yfinance.search
    yfinance.sector_industry
    yfinance.functions
--- a/doc/source/reference/yfinance.market.rst
+++ b/doc/source/reference/yfinance.market.rst
@@ -1,0 +1,16 @@
+=====================
+Market Summary
+=====================
+.. currentmodule:: yfinance
+Class
+------------
+The `Market` class, allows you to access market data in a Pythonic way.
+.. autosummary::
+   :toctree: api/
+   Market
+
+Market Sample Code
+--------------------------
+The `Market` class, allows you to access market summary data in a Pythonic way.
+.. literalinclude:: examples/market.py
+   :language: python
--- a/yfinance/__init__.py
+++ b/yfinance/__init__.py
@@ -30,6 +30,7 @@ from .domain.sector import Sector
 from .domain.industry import Industry
 from .screener.screener import Screener
 from .screener.screener_query import EquityQuery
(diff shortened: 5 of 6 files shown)


ValueRaider:
What is Summary class providing beyond just returning the dict?

R5dan:
Sorry, thought thats how it was wanted as other features (like the search) basically return just the dict. Will add a proper return class

R5dan:
@ValueRaider let me know what you want it to return and then I'll mark it for review

ValueRaider:
My mistake, I meant to question `MarketSummary` class.

R5dan:
> My mistake, I meant to question the `MarketSummary` class.

Yes I suppose it could be a typed dict. It was meant to help provide type hints and the keys available. It also changes some of the keys names.

ValueRaider:
Python variables are already typed, don't need to keep the hints they're for parsing the JSON.

ValueRaider:
> I can only see the type of dict. Not what each of the value types are?

Here:

```
>>> d = {'a':1}
>>> type(d['a'])
<class 'int'>
```

Reserve the Python type hints for function arguments (and return type).

R5dan:
> > I can only see the type of dict. Not what each of the value types are?
> 
> Here:
> 
> ```
> >>> d = {'a':1}
> >>> type(d['a'])
> <class 'int'>
> ```

Sorry misunderstood you.  I thought you meant static type hints in the editor.

> Reserve the Python type hints for function arguments (and return type).

Ok, can I ask why though? Surely more type hints and documentation is better?

ValueRaider:
> Ok, can I ask why though? Surely more type hints and documentation is better?

Our code yes, but not the data structures returned by Yahoo - that's Yahoo's job. They can change their data structure/naming at any time without warning breaking your previous code.

R5dan:
Ok, anyway all done. Is there anything else or shall I mark ready for review

ValueRaider:
I'm not sure on the refresh logic, especially because `the data is very unlikely to change`. I'm sensing this class can be replaced with a simple fetcher function, maybe unpack the data a bit.

R5dan:
> I'm not sure on the refresh logic, especially because `the data is very unlikely to change`. I'm sensing this class can be replaced with a simple fetcher function

I disagree as it does have data that changes daily: `change percentage`, `price confidence`, `price hint`, `price`, `price change`, `previous close`, `tradable`, `marketState` etc. These don't change from minute to minute (except maybe `market price` which does return a different value than `Ticker`) but from hour to hour, day to day, so caching is useful for this.

ValueRaider:
This what I have in mind (only thought to combine your branches today): https://github.com/ranaroussi/yfinance/blob/draft/market-info/yfinance/domain/market.py

```
>>> mkt = yf.Market("US")
>>> print(type(mkt.status))
<class 'dict'>
>>> print(type(mkt.summary))
<class 'dict'>
```

The periodic refresh logic is better handled by `requests_cache` session https://ranaroussi.github.io/yfinance/advanced/caching.html

R5dan:
Surely
``python
if self._status is not None:
   return self._status
```
and
```python
if self._summary is not None:
    return self._summary
```
defeat the purpose of caching: It is only requested once not once an hour or once a day?

I like what you did with `summary`. 

I think storing the data in `_raw_data` or something would help as it will further reduce the requests to Yahoo but also allow `status` to also use that data meaning they are the same.

I would also want to parse the data in `status` like how I have currently got.

What do you think?

> The periodic refresh logic is better handled by `requests_cache` session https://ranaroussi.github.io/yfinance/advanced/caching.html

Forgot about that...

ValueRaider:
> It is only requested once not once an hour or once a day

That's how lazy initialization works, nice side effect as a spam guard. Use other tools to refresh, yfinance focuses on the fetch.

Parse status like this? https://github.com/ranaroussi/yfinance/commit/63698c56c3f3f64da8b43a1e837c893b0100d6b3

Btw any way to get status fetch working with other regions beside US?

R5dan:
> Parse status like this? [63698c5](https://github.com/ranaroussi/yfinance/commit/63698c56c3f3f64da8b43a1e837c893b0100d6b3)

Sorry, no. Was getting confused with #2176 



> Btw any way to get status fetch working with other regions beside US?

**I believe** region doesn't effect the results and is only used for tracking, so shouldn't matter to the user. It might be that certain regions don't have access to certain data (geo blocking) but I don't know.

R5dan:
@ValueRaider You have merged this pr with #2176. Is that what you want or do you want them as 2 separate classes?

ValueRaider:
Merge was deliberate - only way to justify a new custom class, and make sense.

R5dan:
What would you think of market being a class - `market.status.is_open` etc? Otherwise I will put them all in a dict @ValueRaider

ValueRaider:
Abandon `is_open`, wait for Yahoo to complete their backend for `markettime` (other countries). Maybe they discovered that different markets have different times even in same country so only return US. Field `status` might stop being a simple `open` or `closed`.

R5dan:
@ValueRaider How about this?

ValueRaider:
> # Fetch both to ensure they are at the same time

Interesting. Why do they need to be synced?

R5dan:
If you have some data, I think it would be best if they were synced. It seems weird to have summary from time x and then maybe 30 minutes later status. With them being synced it allows better compatibility with the data. @ValueRaider

ValueRaider:
Just noticed the docs for `Market` sucks: https://ranaroussi.github.io/yfinance/reference/index.html . Can you add something? Use the other pages for inspiration.

ValueRaider:
I just noticed the doc page is bare/broken: https://ranaroussi.github.io/yfinance/reference/yfinance.market.html . Can you fix? Use other pages for inspiration. And build the docs yourself to check #1084

PR #2180, opened 2024-12-15
Raise Exception if Rate Limited

dhruvan2006:
### Changes
- New `YfRateLimitError` exception class, which is raised regardless of `raise_errors`
- Raises the above exception if the response from Yahoo! Finance returns `429` status code.

### Usage
Tricky. Can test it with any `yfinance` function. Would need to spam Yahoo! Finance until you get rate limited. But here is an example of how it looks:

image

### DIFF
--- a/yfinance/data.py
+++ b/yfinance/data.py
@@ -10,6 +10,8 @@ from frozendict import frozendict
 from . import utils, cache
 import threading
 
+from .exceptions import YFRateLimitError
+
 cache_maxsize = 64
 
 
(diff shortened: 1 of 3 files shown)


dhruvan2006:
@ValueRaider

ValueRaider:
I'd like to see more comments from users, maybe crosspost this into relevant Issues.

ValueRaider:
Will this raise in case of issue #2185? If yes then definite merge. Btw this should raise regardless of `raise_errors` argument, the user needs to take immediate action! So revert the change to `raise_errors`.

Situation is different to a loop where one ticker is delisted, the loop should continue, this is why `raise_errors` exists.

dhruvan2006:
> Will this raise in case of issue #2185? If yes then definite merge. Btw this should raise regardless of `raise_errors` argument, the user needs to take immediate action! So revert the change to `raise_errors`.
> 
> Situation is different to a loop where one ticker is delisted, the loop should continue, this is why `raise_errors` exists

Yes it will raise for #2185, I was testing it. Changes to `raise_errors` reverted and rate limit will always be raised regardless.

PR #2186, opened 2024-12-22
pass proxy and sesseion to default PriceHistory object

pata00:
fix this case:
`
import yfinance as yf

n225 = yf.Ticker("^N225", proxy="socks5://127.0.0.1:1080")

df = n225.history()

print(df)
`


ValueRaider:
Better: pass proxy to the data object like session https://github.com/ranaroussi/yfinance/blob/5adddf3e88f39720ef24a04927738848dee76761/yfinance/base.py#L67

Issue #154, opened 2019-12-26
Is the package able to extract top gainers via Yahoo API?


boyac:
Is the package able to extract top gainers via Yahoo API? Thanks!

CacheMoneyPlaya:
just use beautiful soup:

```
        url = "https://finance.yahoo.com/gainers"
        soup = BeautifulSoup(requests.get(url).text, 'html.parser')
        assets = soup.find_all('a', attrs={"class":"Fw(600)"})
```

This gets top 5 performers of the day

boyac:
@CacheMoneyPlaya  Thanks a million!!

legout:
You can also use pandas therefore.

```
import pandas as pd
gainers = pd.read_html('https://finance.yahoo.com/gainers')
```

drizham:
is there away of getting top gainers for a specific index? Like DAX or FTSE100 or Nikkei?

IoIxD:
This is an awful solution. The page will give you a 404 if you don't have a user agent set, and even if you do, the resulting page is ginormous because yahoo craps their site to the brim with useless shit.

    curl -A "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36" https://finance.yahoo.com/gainers -o lol
      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                     Dload  Upload   Total   Spent    Left  Speed
    100 1909k    0 1909k    0     0  1517k      0 --:--:--  0:00:01 --:--:-- 1517k
    ➜  ~ du lol -h
    1.9M    lol
    ➜  ~

ValueRaider:
`yahooquery` code has references to "gainers", maybe look into that (`yahooquery` uses API instead of scraping = faster). If you're interested in porting to yfinance, post in #1546

MikePendo:
```
screener = yf.Screener()
 screener.set_predefined_body(predefined_key='day_gainers')
 dict = screener.response
```

MikePendo:
The OP didnt mention what gainers he wants. I dont see an api of how to get 52 week gainers.
@ValueRaider any idea? (sorry I am very new to it)

ValueRaider:
Is 52 weeks gainers on website?

R5dan:
> Is 52 weeks gainers on website?

Yes: https://finance.yahoo.com/markets/stocks/52-week-gainers/

[...]

PR #2190, opened 2024-12-23
Redo Screener

R5dan:
It is annoying that the types are wrong on `Screener`, I find myself looking at the source code to determine the types of the `Screener` class so this is a fix


ericpien:
1. what do you mean the "types are wrong"?
2. What does it mean to "determine the types of the Screener class"?
3. How can we improve the documentation for Screener to avoid "looking at the source code"?

R5dan:
> 1\. what do you mean the "types are wrong"?

The return types on the Screener are wrong (for example `dict` not `dict_keys` on `predefined_bodies` which seems trivial but the methods are different on them) etc

> 2\. What does it mean to "determine the types of the Screener class"?

I go into the source code to work out what types different properties are as when I run code (which would be correct with the types) it fails (due to incorrect types provided)

I also have to go looking for what string literals are allowed in `set_predefined_body` when they are hard coded in and most type checkers can prompt you to which ones to use.

> 3\. How can we improve the documentation for Screener to avoid "looking at the source code"?

Mostly fixed it with this PR. Take a look if you want.

I also made the the types run faster and be more runtime safe: by putting the type in a string it only has to construct a string and not the full class and it also means it's easier to fix circular import errors or the class not actually existing for some reason as some of them could be moved under `t.TYPE_CHECKING`

R5dan:
@ValueRaider What do you think?

ValueRaider:
> How can we improve the documentation for Screener

Improve how user sets up a screener query with a simpler intermediate form:
```
from dataclasses import dataclass
from typing import List, Union, Any

@dataclass
class Query:
    operator: str
    operands: List[Union['Query', Any]]

    def to_dict(self) -> dict:
        return {
            "operator": self.operator,
            "operands": [op.to_dict() if isinstance(op, Query) else op 
                         for op in self.operands]}

def value_in(key: str, *values):
   return Query("or", [Query("eq", [key, v]) for v in values])

query = Query(
    "and",
    [
        value_in("exchange", "NMS", "NYQ"),
        Query("lt", ["epsgrowth.lasttwelvemonths", 15])
    ]
)
```

Should shorten the screener html page by a lot.

R5dan:
Should I also add #154 into this one?

ValueRaider:
ok

R5dan:
@ValueRaider any other predefined keys you know off? Still don't know where you found that 52 week gain

R5dan:
Also why is `Query` and `EquityQuery` different? `EquityQuery` doesn't even have region etc.

ValueRaider:
> @ValueRaider any other predefined keys you know off? Still don't know where you found that 52 week gain

I just traversed the gainers webpage with Firefox network inspector open.

R5dan:
@ValueRaider Still need to add proper logging and docs etc, but what do you think of this.
I am getting the error `{"finance":{"result":null,"error":{"code":"internal-error","description":"sort field is not a field: fundnetassets"}}}` returned by yahoo. I don't understand why as this part is the same as the original and that works fine

R5dan:
Due to how many classes there are, I decided to put it behind `yfinance.screener` @ValueRaider

ValueRaider:
Sure but:

> AttributeError: module 'yfinance' has no attribute 'screener'

Can you also set one of the predefined queries e.g. `aggressive_small_caps`. It'll help me understand how to use.

yfinance/screener/screener.py
```diff
@@ -154,72 +246,94 @@ def set_body(self, body: Dict) -> 'Screener':
...
+        if "query" in body.keys():
+            raise ValueError("Query must not be set here: set query")
+
         self._body_updated = True
         self._body = body
         return self
 
-    def patch_body(self, values: Dict) -> 'Screener':
+    
+    def patch_body(self, offset: 'int' = 0, size: 'int' = 100, sortField: 'str' = "ticker", sortType: 'str' = "desc", quoteType: 'str' = "equity", userId: 'str' = "", userIdType: 'str' = "guid"):
```
ericpien, line 269: doesn't this result in unintentional updates of certain fields?

i.e. user with body offset = 50, goes to patch the sortField but offset will reset to 0.

yfinance/screener/screener.py
```diff
@@ -1,145 +1,237 @@
...
-    The `Screener` class is used to execute the queries and return the filtered results.
+_REQUIRED_BODY_TYPE_ = TypedDict("_REQUIRED_BODY_TYPE_", {"offset": int, "size": int, "sortField": str, "sortType": str, "quoteType": str, "userId": str, "userIdType": str})
+_BODY_TYPE_ = TypedDict("_BODY_TYPE_", {"offset": int, "size": int, "sortField": str, "sortType": str, "quoteType": str, "query": dict, "userId": str, "userIdType": str}, total=False)
+_EMPTY_DICT_ = TypedDict("_EMPTY_DICT_", {}, total=False)
 
-    The Screener class provides methods to set and manipulate the body of a screener request,
-    fetch and parse the screener results, and access predefined screener bodies.
-    """
-    def __init__(self, session=None, proxy=None):
+class PredefinedScreener:
```
ericpien, line 17: why do we need a separate class for this? Isn't this just a Screener with pre-defined fields?

R5dan: No, it sends a different request to a different endpoint with a different response

yfinance/screener/screener.py
R5dan: Yes it would, except it is not implemented yet. Will change it when I do implement it

R5dan:
> Sure but:
> 
> > AttributeError: module 'yfinance' has no attribute 'screener'

Weird, I don't get that error:

```python
from yfinance.screener import Screener

predefined = Screener.predefined("day_gainers", count=25)
```

yfinance/screener/screener.py
ericpien: If you access the yahoo's screener via: https://finance.yahoo.com/research-hub/screener, the request is made to the same endpoint for the pre-defined queries.

imho, it's better to not create a separate class for this to allow user to construct a `Screener` that reflects the pre-defined body, and edit it as they need for flexibility. i.e. offset, or exchange requirements.

ericpien:
Below are just my two cents:

I think this affects more than just `types` as initially proposed and there isn't much example/test code to infer. Can you edit the tests for the changes to assist in code review? I'm actually not sure what problem this PR is trying to solve anymore.

> Also why is `Query` and `EquityQuery` different? `EquityQuery` doesn't even have region etc.

This was done so future developers can extend to different types of queries but leverage the same screener class. i.e. `quoteType = "MUTUAL FUNDS"`

> I also made the the types run faster and be more runtime safe: by putting the type in a string it only has to construct a string and not the full class and it also means it's easier to fix circular import errors or the class not actually existing for some reason as some of them could be moved under t.TYPE_CHECKING if needed

I personally have a preference for using the typing module instead for readability and maintainability.

R5dan:
It was initially just to edit the Types but then I was asked to fix the Query and EntityQuery and make screener better and easier. I understand what you are saying about having no direction, it is now to redo the screener to make it easier to use and make custom queries etc.

> I personally have a preference for using the typing module instead for readability and maintainability.

Personally I generally prefer as little from typing as possible (except certain things)

> there isn't much example/test code to infer

Yes, I am going to add more, just wanted to see whether people like this, or want specific and overall changes before implementing more specific, adding docs and tests, I just wanted to get something working. This is a draft pr so far.

ValueRaider:
@R5dan You deleted changes in `__init__.py` here https://github.com/ranaroussi/yfinance/compare/main...R5dan:yfinance-package:Screener-Types#diff-589131f3d462211152427076ff4599b8de70df363e8e4d64c8f68f3e0c7d514b

Are you double-checking the state of your local branch vs remote? Missing changes, divergence, etc

R5dan:
@ValueRaider My local branch is almost exactly the same as the one pushed here (only differences is docs or parameter name changes) so this pr is basically up to date (excluding what I'm currently working on). I have deleted the `EquityQuery` and `Screener` from `yfinance/__init__.py` and put them under `yfinance.screener` so they should be imported like `from yfinance.screener import Screener, Query, Validator # Or the premade validators`. What code are you running?

ValueRaider:
@R5dan I am running the commit I linked to. A fresh clean pull of your branch `Screener-Types`.

@ericpien Does the import code work for you? Or do you get same `no attribute` error as me?

R5dan:
> @R5dan I am running the commit I linked to. A fresh clean pull of your branch `Screener-Types`.

No what is your test code?

ValueRaider:
I'm literally just trying the most basic code:
```
>>> yfinance.screener
AttributeError: module 'yfinance' has no attribute 'screener'
>>> yfinance.Screener
AttributeError: module 'yfinance' has no attribute 'Screener'
```

You need to put `Screener` or `screener` into `__init__py`, like is done for everything else imported via `yfinance` like `yfinance.Ticker`.

R5dan:
Sorry, didn't understand the issue before.

You can't access it like that, must be `from yfinance import screener`
Implemented a fix for it by importing it in `__init__` as before it was classed as a sperate sub-package and so couldn't be imported from the main `yfinance` package.

ValueRaider:
Ok, I have your code running now. I've deleted those comments to tidy this thread and get back on track.

> I'm actually not sure what problem this PR is trying to solve anymore.

Might be my fault. I only just now understand that @ericpien implemented a `Query` class that appears to do [what I wanted](https://github.com/ranaroussi/yfinance/pull/2190#issuecomment-2565506718). I need to spend more time with the existing screener and query.

R5dan:
> Might be my fault. I only just now understand that @ericpien implemented a `Query` class that appears to do [what I wanted](https://github.com/ranaroussi/yfinance/pull/2190#issuecomment-2565506718).

Do you want me to stop this then?

ValueRaider:
> Do you want me to stop this then?

Pause for now. I'll look more over weekend.

R5dan:
@ValueRaider Made a decision? Are you just going to go with #2207

ValueRaider:
@R5dan relax, let's work through #2207 then after see where this PR fits in.

R5dan:
I was just checking. Wanted to see if you wanted me to continue or not, so Ill pause for now.

Issue #1972, opened 2024-06-29
yf.download() not accepting custom period anymore → is this truly desired behavior?


MichalRIcar:
### Describe bug

I've been using yfinance for a few years now with the same syntax. However, the last updates changed the API. 
Thus I wonder if download(period=X) is intentionally changed not to accept user defined period as it used to be and forcing to pre-defined "Valid periods"?

The orig. syntax 
→ yf.download('^SPX', period='30d', interval='1h', ignore_tz = True, progress=False) 
the orig. syntax produces error such as:
→ ...Period 'Xd' is invalid, must be one of ['1d', '5d', '1mo', '3mo', '6mo', '1y', '2y', '5y', '10y', 'ytd', 'max']")

The fix is to use start-end params. That is not a hard thing to do, but I think → **shouldn't the yf API keep its behavior and compute start internally, whereas the end param = int(_time.time()), when the user sets a custom period rather than throwing an error?**

### Simple code that reproduces your problem

```
import yfinance  as yf

# worked for several years
yf.download('^SPX', period='30d', interval='1h', ignore_tz = True, progress=False)
# leads to: ...Period '30d' is invalid, must be one of ['1d', '5d', '1mo', '3mo', '6mo', '1y', '2y', '5y', '10y', 'ytd', 'max']") 

# newly forced to use start-end instead
# yf.download('^SPX', start = X, interval='1h', ignore_tz = True, progress=False)
```

### Debug log

The bug is given by new syntax forcing user to use start-end instead of custom period.

### `yfinance` version

0.2.40

### Python version

3.12.4

### Operating system

Win11

ValueRaider:
yfinance hasn't changed handling period: https://github.com/ranaroussi/yfinance/blob/0.1.55/yfinance/base.py

yfinance isn't affiliated with Yahoo

MichalRIcar:
Thanks, I am aware of independent dev of yfinance and appreciate it very much. 
I've started observing the behavior (not accepting custom period anymore) a month ago. 
As the fix in this case is truly trivial - to compute internally start param for a custom (not supported) period - I was thinking is worth nothing to share the observation, rahter than just silently fix my pipeline.

andrei-cb:
30d period is still working with version 0.2.18, looks like an issue with newer versions of yfinance

ValueRaider:
It started with 0.2.32 when added cookie & crumb.

the-sean-c:
[...]

PR #2192, opened 2024-12-24
Custom period

R5dan:
Fix for #1972.

Coverts custom periods into a start time (with end being today)


ValueRaider:
Merge logic into `utils._interval_to_timedelta`

R5dan:
@ValueRaider From what I can see it would only be called then if the `interval` is 5d, 1wk, 1mo, 3mo: where the original code for `utils._interval_to_timedelta` is fine. It would also then only be run if `repair` is `True`. Could you explain why you want these changes?

ValueRaider:
It's good coding practice: avoid duplicate logic.

> It would also then only be run if repair is True

Huh? You are allowed to add another call to `_interval_to_timedelta`.

R5dan:
@ValueRaider done

PR #2194, opened 2024-12-26
Fix: Remove hardcoded keys in Analysis

### LINKED ISSUE #2188: eps_revisions always returns 'None' for 'downLast7days'
### Discussed in https://github.com/ranaroussi/yfinance/discussions/2187

<div type='discussions-op-text'>

<sup>Originally posted by **avasfx** December 22, 2024</sup>
Hello. I updated yfinance to version 0.2.51. 
then:
`import yfinance as yf
tsla = yf.Ticker("TSLA")
tsla.eps_revisions`
and i get something like that:
image

But on yahoo finance site there is data for revisions:
image
The same is for other stocks. Am i missing something or yahoo changed something once again?</div>

dhruvan2006:
Removed hard coded keys in the Analysis scraper functions to protect against Yahoo changing the keys in the future.

# Changes

End result is more or else the same except:
- Few column names have changed
- Index has names

#### Before:
image
#### After:
image

#### Before:
image
#### After:
image

#### Before:
image
#### After:
image

#### Before:
image
#### After:
image

#### Before:
image
#### After:
image

#### Before:
image
#### After:
image

Closes #2188

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -760,7 +760,7 @@ class TestTickerAnalysts(unittest.TestCase):
         self.assertIsInstance(data, pd.DataFrame, "data has wrong type")
         self.assertFalse(data.empty, "data is empty")
         self.assertTrue(len(data.columns) == 4, "data has wrong number of columns")
-        self.assertEqual(data.columns.values.tolist(), ['Firm', 'ToGrade', 'FromGrade', 'Action'], "data has wrong column names")
+        self.assertCountEqual(data.columns.values.tolist(), ['Firm', 'ToGrade', 'FromGrade', 'Action'], "data has wrong column names")
         self.assertIsInstance(data.index, pd.DatetimeIndex, "data has wrong index type")
 
         data_cached = self.ticker.upgrades_downgrades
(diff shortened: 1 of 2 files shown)


R5dan:
It would be quicker and simpler to simply store the keys as lower case strings

dhruvan2006:
> Surely it would be quicker and simpler to simply store the keys as lower case strings

I am aiming to keep backwards compatibility

ValueRaider:
And what if Yahoo changes another key? Better if most hardcodings go - just unpack the json and use what Yahoo provides as much as possible. Try this:
```
dd = []
for item in self._earnings_trend:
    ddi = {'period': item['period']}
    for k,v in item['epsRevisions'].items():
        if not isinstance(v, dict) or len(v) == 0:
            continue
        ddi[k] = v['raw']
    dd.append(ddi)
self._eps_revisions = pd.DataFrame(dd).set_index('period')
```

R5dan:
> And what if Yahoo changes another key?

Yes, especially as they already have added one: 90 days

ValueRaider:
Do you mind refactoring the other fetch functions?

dhruvan2006:
> Do you mind refactoring the other fetch functions?

Done. Edited PR description to show changes to end result

PR #2199, opened 2025-01-02
Added missing extra dependencies to requirements.txt

dhruvan2006:


ValueRaider:
No, those are optional dependencies. https://ranaroussi.github.io/yfinance/advanced/caching.html

dhruvan2006:
But this is for developers. The setup.py is unchanged. 

It is very disheartening as a newcomer to run tests and see them all fail.

ValueRaider:
You're right, thanks for correcting me.

PR #2202, opened 2025-01-02
Fix: Extra key in Search quotes

dhruvan2006:
Retrieving quotes using `Search` was returning an extra row.

```python
{
  'index': '78ddc07626ff4bbcae663e88514c23a0',
  'name': 'AAPlasma',
  'permalink': 'aaplasma',
  'isYahooFinance': False
}
```


ValueRaider:
Is this fixing a bug? Or is Yahoo suddenly returning more/different data?

dhruvan2006:
Yahoo returning something different

PR #2207, opened 2025-01-05
Improve Screener docs

ValueRaider:
The main change is converting the predefined bodies to use `EquityQuery`. Also simplify the queries, Yahoo has made them longer than necessary. This will make Screener doc page shorter and clearer.

@ericpien what is purpose of `Screener.patch_body()`? It's not documented well.


R5dan (changes requested):

ericpien:
> @ericpien what is purpose of Screener.patch_body()? It's not documented well.

It is meant to update just the provided portion of the body. i.e. user wants to repeatedly use the same `Screener` but just update the `_body.offset` by 200. Instead of reconstructing the whole `Query` or `Screener`, user can just update that value by passing the dict to the method then iterate in a loop.

yfinance/screener/screener_query.py
```diff
@@ -69,7 +77,9 @@ def __init__(self, operator: str, operand: Union[numbers.Real, str, List['Equity
         if len(operand) <= 0:
             raise ValueError('Invalid field for Screener')
             
-        if operator in {'OR','AND'}: 
+        if operator == 'IS-IN':
```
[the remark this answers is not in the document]
ericpien: I wonder about this from design perspective as well. I tried to retain as much "likeliness" as possible to Yahoo's operator types. Since there is no "is-in" operator in yahoo's own query, it could get unnecessarily difficult to debug / extend in the future. Is there a reason to take on that cost?

yfinance/const.py
```diff
@@ -431,7 +431,12 @@
         "region",
         "sector",
         "peer_group",
-        "exchanges"}, 
+        "exchange",
+        "categoryname",
+        "performanceratingoverall",
+        "initialinvestment", 
+        "annualreturnnavy1categoryrank", 
+        "riskratingoverall"},
```
ericpien, line 439: Are these valid fields to screen Equity? or is it reserved for Mutual Funds?

yfinance/screener/screener_query.py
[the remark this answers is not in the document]
R5dan: Yes, that was what I was trying to get across.
I think it would be better to have a function for `is_in` if at all:
```python
def is_in(key, *values):
    operands = []
    for oper in values:
        operands.append(EquityQuery("EQ", oper))
    
   return EquityQuery("OR", operands)
```

Similar to how the `to_dict` works, just earlier.

ValueRaider:
> > The main change is converting the predefined bodies to use EquityQuery
> 
> Is the advantage of this cleaner documentation?

It's the only reason, to shorten the docs. Next task is collapsing the long lists.

Also now `Screener` has been completely redone, it did not "feel right" before.

yfinance/const.py
ValueRaider: Good point. Screener needs a new `FundQuery` aka `FundOperation` class.

yfinance/screener/screener_query.py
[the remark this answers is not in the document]
ValueRaider: The primary goal is to simplify task of writing queries, and part of that is length.

> from yfinance import EquityQuery as EQy
> EQy(IS-IN, ['exchange', 'NMS', 'NYQ'])

vs

> {'operands': [{'operands': ['exchange', 'NMS'], 'operator': 'EQ'}, {'operands': ['exchange', 'NYQ'], 'operator': 'EQ'}], 'operator': 'OR'}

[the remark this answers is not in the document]
ValueRaider: What would the user type?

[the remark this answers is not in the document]
R5dan: Something like `EquityQuery.is_in` which directly returns all the `EquityQuery`'s

ValueRaider:
This is almost ready for review. I just need to think more about the actual execution of a query, my current refactoring still doesn't feel right.

yfinance/screener/screener.py
```diff
@@ -2,10 +2,33 @@
...
+    'solid_midcap_growth_funds': {"offset":0,"size":25,"sortType":"DESC","sortField":"fundnetassets","quoteType":"MUTUALFUND","query":{"operator":"and","operands":[{"operator":"or","operands":[{"operator":"EQ","operands":["categoryname","Mid-Cap Growth"]}]},{"operator":"or","operands":[{"operator":"EQ","operands":["performanceratingoverall",5]},{"operator":"EQ","operands":["performanceratingoverall",4]}]},{"operator":"lt","operands":["initialinvestment",100001]},{"operator":"lt","operands":["annualreturnnavy1categoryrank",50]},{"operator":"or","operands":[{"operator":"eq","operands":["exchange","NAS"]}]}]},"userId":"","userIdType":"guid"},
+    'top_mutual_funds': {"offset":0,"size":25,"sortType":"DESC","sortField":"percentchange","quoteType":"MUTUALFUND","query":{"operator":"and","operands":[{"operator":"gt","operands":["intradayprice",15]},{"operator":"or","operands":[{"operator":"EQ","operands":["performanceratingoverall",5]},{"operator":"EQ","operands":["performanceratingoverall",4]}]},{"operator":"gt","operands":["initialinvestment",1000]},{"operator":"or","operands":[{"operator":"eq","operands":["exchange","NAS"]}]}]},"userId":"","userIdType":"guid"}
+}
```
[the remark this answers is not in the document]
ValueRaider: They're all changed now.

> Why are these now here instead of `const.py`?

Can't remember why I moved them, probably to help me refactor other stuff. They can go back to `const.py`.

R5dan:
Can I ask why predefined screeners are still made and not just sent to the predefined endpoint?

ValueRaider:
Fewer characters - generated docs are shorter.

ValueRaider:
> This is almost ready for review. I just need to think more about the actual execution of a query, my current refactoring still doesn't feel right.

Fixed

yfinance/screener/screener.py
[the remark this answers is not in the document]
ValueRaider: I remember now. Putting these in const.py requires adding imports at top, not sure about that.

yfinance/const.py
```diff
@@ -305,21 +305,43 @@
...
+            elif isinstance(value, dict) and isinstance(result[key], dict):
+                result[key] = {
+                    k: (result[key].get(k, set()) | v if isinstance(v, set) 
+                        else v) if k in result[key]
+                    else v
+                    for k, v in value.items()
+                }
+        else:
+            result[key] = value
+    return result
```
R5dan, line 326: Why is this not in `utils.py`?

yfinance/screener/query.py
```diff
@@ -0,0 +1,211 @@
...
+                s += o.__repr__()
+            if i < len(self.operands)-1:
+                s += ', '
+        s += ']'
+        if root:
+            s += '"'
+        return s
+
+    def __str__(self) -> str:
+        return self.__repr__()
```
R5dan, line 125: Wouldn't `__str__ = __repr__` as it is more concise and slightly more readable?

yfinance/screener/screener.py
```diff
@@ -1,225 +1,120 @@
...
+    # Fetch
+    _data = YfData(session=session)
+    params_dict = {"corsDomain": "finance.yahoo.com", "formatted": "false", "lang": "en-US", "region": "US"}
+    response = _data.post(_SCREENER_URL_, 
+                            body=post_query, 
+                            user_agent_headers=_data.user_agent_headers, 
+                            params=params_dict, 
+                            proxy=proxy)
+    response.raise_for_status()
+    return response.json()['finance']['result'][0]
```
R5dan, line 138: Personally think this should be moved to a function and that all strings should be sent to the predefined screener url, that way it doesn't need updating if one of the screeners is updated or a new one is added.

yfinance/utils.py
```diff
@@ -961,29 +961,40 @@ def generate_list_table_from_dict(data: dict, bullets: bool=True) -> str:
...
+            table += ' '*5 + f"- {value_str}\n"
     return table
 
 def generate_list_table_from_dict_of_dict(data: dict, bullets: bool=True) -> str:
     """
     Generate a list-table for the docstring showing permitted keys/values.
     """
     table = _generate_table_configurations()
-    for key, values in data.items():
+    # for k, values in data.items():
```
R5dan, line 991: Why comment it out and not just delete it as it is not used anyway?

yfinance/screener/query.py
```diff
@@ -0,0 +1,211 @@
...
+    def __init__(self, operator: str, operand: Union[numbers.Real, str, List['QueryBase']]):
+        operator = operator.upper()
+
+        if not isinstance(operand, list):
+            raise TypeError('Invalid operand type')
+        if len(operand) <= 0:
+            raise ValueError('Invalid field for EquityQuery')
+            
+        if operator == 'IS-IN':
+            self._validate_isin_operand(operand)
```
R5dan, line 21: What about `NOT-IN`?

R5dan (changes requested):

yfinance/screener/query.py
ValueRaider: Requires Yahoo support `NOT-EQ`, or just `NOT`, have you seen this?

R5dan: Thought I had but can't find it so probably not, as yahoo also doesn't use it, but even then we know all the possibilities, so can just create a `or` query for all the others.
`Query.not_in("sector", "technology")` returns `Query("OR", [Query("EQ", sector) for sector in SECTORS if sector != not_in]` or something

ValueRaider: What about field `performanceratingoverall`? @ericpien didn't write down all values for `FundQuery` fields, so all I know is it's an integer.

yfinance/screener/screener.py
ValueRaider: > all strings should be sent to the predefined screener url

How is this not already happening?

yfinance/screener/query.py
ValueRaider: You mean like this?
```
class QueryBase(ABC):
    def __init__
    ...
QueryBase.__str__ = QueryBase.__repr__
```

That's less readable imo, it's not at the same indent as other class functions, not the expected place.

yfinance/const.py
ValueRaider: circular import happens

yfinance/screener/query.py
R5dan: I see your point, 2 options to make it work: 1 is trying to find out the options of the fields, 2 is only having it work for certain fields.



> What about field `performanceratingoverall`? ericpien didn't enumerate values for `FundQuery` fields, so all I know is it's an integer

`Query("OR", [Query("BTWN", 0, value-1), Query("BTWN", value+1, 99999...)]`?

yfinance/screener/screener_query.py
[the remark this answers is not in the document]
ValueRaider: Coming back to this:
> Is there a reason to take on that cost?

What's the cost? It's basically a one-liner now.

yfinance/screener/screener.py
R5dan: You have created all of the predefined screeners. Personally I think that it should be something closer to:
```python
def screen(query:'EquityQuery|str', ...):
    if isinstance(query, str):
        # Handle predefined
        _data = YfData()
        params_dict = {
		"scrIds": query,
		...
        }
        resp = _data.get(url=_PREDEFINED_URL_, params=params_dict, proxy=self.proxy)
        resp.raise_for_status()
        return resp.json()["finance"]["result"][0]
```
or something like this. There is an endpoint explicitly for predefined screeners, why not use it?

yfinance/screener/screener_query.py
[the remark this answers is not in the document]
R5dan: > It's basically a one-liner now.

`EquityQuery.in_in` is also a single line



> What's the cost?

Might be misunderstanding but I thought @ericpien was agreeing with me on that statement as he was talking about the cost of debugging `EquityQuery("in-in")` in the future. I think it is worth it as it keeps it similar as similar to yahoo as possible, it makes it easier to change or add a new one in the future. Also, currently it is in with the "basic" operands yahoo provides, which I feel is confusing as it is not one yahoo provides.

yfinance/screener/query.py
R5dan: It can be done with
```python
class QueryBase(ABC):
	def __str__(self):...

	__repr__ = __str__
```
but it was just something I thought of at the time and doesnt matter

R5dan:
Could there be `EquityQuery.screen` which calls `yf.screen` on itself. I think that might be a bit cleaner in may places.

yfinance/screener/screener.py
ValueRaider: > You have created all of the predefined screeners.

Nope, that was @ericpien https://github.com/ranaroussi/yfinance/blob/5adddf3e88f39720ef24a04927738848dee76761/yfinance/const.py#L534-L550

One thing I noticed when switching to predefined endpoint, is because only sending the name and not the other fields, then results are different - different sort type, different order. Probably why @ericpien hardcoded the entire query with fields.

R5dan: > predefined endpoint, is because only sending the name and not the other fields, then results are different - different sort type, different order

Thats one of the reasons I want to send the request to the other endpoint, a user shouldn't test something on yahoo and use it here only to find that they return 2 different things, with different functionality

ValueRaider: @ericpien can you remember your thinking with the predefineds? https://finance.yahoo.com/research-hub/screener uses the predefined endpoint and doesn't seem to set the other fields.

ericpien: > @ericpien can you remember your thinking with the predefineds? https://finance.yahoo.com/research-hub/screener uses the predefined endpoint and doesn't seem to set the other fields.

I noticed the "pre-defined" basically acts as a regular query with "pre-defined fields" and since these fields are part of the response object, I thought it would be better to use the regular `/screener` endpoint + the known fields so users can start the query with the pre-defined field, and modify as they need. i.e. if they want the top day gainers but restrict it to certain geography, this can be done by adjusting the Day Gainer Stocks query instead of reconstructing the entire query. 

When I spot checked, the returned list was identical.

yfinance/screener/screener_query.py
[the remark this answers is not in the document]
ericpien: > Might be misunderstanding but I thought @ericpien was agreeing with me on that statement as he was talking about the cost of debugging EquityQuery("in-in") in the future.

Yes, I was referring to cost of debugging. For me, the question was more so a design philosophy one. I'm new to the yfinance project so not sure what the preference is in maintaining "likeness" of the source vs offering functionalities where query doesn't explicitly exist on the yahoo side.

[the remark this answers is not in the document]
R5dan: Also, it is then possible to change the queries returned by the `is-in`. If the user wants to change a particular part of it later they can, without having to reconstruct the whole thing.

yfinance/screener/screener.py
R5dan: > since these fields are part of the response object

Yes, I had seen that. If they are also returned with the standard endpoint might be a useful logging tool.

> modify as they need

Yes, that was the only thing making me want to go with predefining fields. Maybe if there is a classmethod where you enter a string and it fetches from the predefined endpoint so that any additional predefined screeners can be fetched instantly without needing a package update to get them. That way the predefined bodies can stay, but if they update the screeners it will act as the website does.

R5dan:
I think that in the query there should be a to modify elements for geography (where data can be returned from not what data is returned), almost like a head tag in HTML as for a particular query, so users can keep the actual query but easily change the market for example.

That way there could be `is in` like functionality to set the geography or to set it worldwide

yfinance/screener/screener.py
ValueRaider: Proposal: the predefined queries + fields are only in the online documentation for reference. Only the string name is actually sent to Yahoo.

ValueRaider: > Can I ask why?

The documentation is 10% auto-generated from code. If it's not in the code, it's not in the docs.

R5dan: Im not disagreeing with it being in the code base and docs, Im asking if its in the code why not have it implemented?

ValueRaider: Bruh.

> There is an endpoint explicitly for predefined screeners, why not use it?

R5dan: True, I thought that the need to be able to edit was important, otherwise I completely agree with your suggestion.

ValueRaider: I figure they can just copy-paste from the docs. A `Query` instance shouldn't be modified, it undermines all of the validation work you two have done, so exposing `PREDEFINED_SCREENER_QUERIES` not useful.

ValueRaider: Latest commit should resolve this (remember to review web docs not just code)

yfinance/screener/screener.py
```diff
@@ -1,225 +1,177 @@
...
+                                "query": FndQy('and', [FndQy('eq', ['categoryname', 'Large Growth']), FndQy('is-in', ['performanceratingoverall', 4, 5]), FndQy('lt', ['initialinvestment', 100001]), FndQy('lt', ['annualreturnnavy1categoryrank', 50]), FndQy('eq', ['exchange', 'NAS'])])},
+    'solid_midcap_growth_funds': {"sortType":"DESC", "sortField":"fundnetassets",
+                                "query": FndQy('and', [FndQy('eq', ['categoryname', 'Mid-Cap Growth']), FndQy('is-in', ['performanceratingoverall', 4, 5]), FndQy('lt', ['initialinvestment', 100001]), FndQy('lt', ['annualreturnnavy1categoryrank', 50]), FndQy('eq', ['exchange', 'NAS'])])},
+    'top_mutual_funds': {"sortType":"DESC", "sortField":"percentchange",
+                        "query": FndQy('and', [FndQy('gt', ['intradayprice', 15]), FndQy('is-in', ['performanceratingoverall', 4, 5]), FndQy('gt', ['initialinvestment', 1000]), FndQy('eq', ['exchange', 'NAS'])])}
+}
+
+@dynamic_docstring({"predefined_screeners": generate_list_table_from_dict_universal(PREDEFINED_SCREENER_QUERIES, bullets=True, title='Predefined queries (Dec-2024)')})
+def screen(query: Union[str, QueryBase],
```
R5dan, line 53: In my opinion this should probably be a type literal instead of string. For the majority of users, they will want to stick with a very specific set of options and (if any more are added) they will want to be very deliberate about it

yfinance/screener/screener.py
```diff
@@ -1,225 +1,177 @@
...
+    if size is not None and size > 250:
+        raise ValueError("Yahoo limits query size to 250, reduce size.")
+
+    params_dict = {"corsDomain": "finance.yahoo.com", "formatted": "false", "lang": "en-US", "region": "US"}
+
+    args = locals()
+    post_query = None
+    if isinstance(query, str):
+        if query not in PREDEFINED_SCREENER_QUERIES.keys():
+            raise ValueError(f'Invalid key {query} provided for predefined screener')
```
R5dan, line 127: Could there be a `force` param? This would default to `False` but allow any string instead of just those in `PREDEFINED_SCREENER_QUERIES`. I know this is contradictory to having query heavily typed but that is because 99.9% of users will only ever need to use the ones stated here but if a new predefined is made, they shouldn't have to wait for an update or be forced to update.

R5dan (changes requested):

yfinance/screener/screener.py
R5dan: > remember to review web docs not just code

I didn't realize that PR's had docs? Or am I misunderstanding your comment?

yfinance/screener/screener.py
```diff
@@ -1,225 +1,177 @@
...
+        'userId': "",
+        'userIdType': "guid"
+    }
+
+    if size is not None and size > 250:
+        raise ValueError("Yahoo limits query size to 250, reduce size.")
+
+    params_dict = {"corsDomain": "finance.yahoo.com", "formatted": "false", "lang": "en-US", "region": "US"}
+
+    args = locals()
```
R5dan, line 123: Why is `args` assigned to `locals`? Surely it is safer, better and more readable to just assign it to what you want it to be?

My safety concern come from the docstring of locals:
> NOTE: Whether or not updates to this dictionary will affect name lookups in the local scope and vice-versa is *implementation dependent* and not covered by any backwards compatibility guarantees.

ValueRaider: Does HEAD resolve?

yfinance/screener/screener.py
ValueRaider: I've pushed a proposed solution. That check is gone, and if a HTTPError occurs then cross-check against `PREDEFINED_SCREENER_QUERIES` to print a helpful message.

yfinance/screener/screener.py
ValueRaider: They can read the online documentation.

ValueRaider: How is this compatible with allowing users to submit a string not in `PREDEFINED_SCREENER_QUERIES`?

R5dan: Because 99.9% of users will only need the ones we define, so it should have a type of a string literal to help that, however because of the ease of implementation we shouldn't stop the 0.1% of use cases (which is why there shouldn't have to be an error).

Basically most users will only need to use certain predefined screeners so we type hint to tell them, these are valid screeners
However for those who know the api well, shouldn't be restricted based on ones we defined, especially if it is new, hence why not raising an error and only having a type error (which isnt even raised at runtime)

ValueRaider: My concern with the string literal is it's long, 350 characters. Can you trial in your IDE?

yfinance/screener/screener.py
ValueRaider: This https://ranaroussi.github.io/yfinance/index.html. It's generated from code.

yfinance/screener/screener.py
R5dan: I have tested it, I personally dont mind having it as a raw string literal, but if you dont want that you could put it in a variable like `PREDEFINED_SCREENER_LITERAL` and then in the type hint it just shows that but still has the same "functionality" - where the IDE will prompt what to put in and the type checker will error if its not in it.

yfinance/screener/screener.py
R5dan: What do you mean by this?

yfinance/screener/screener.py
ValueRaider: It doesn't, you have to generate it https://github.com/ranaroussi/yfinance/discussions/1084

yfinance/screener/screener.py
ValueRaider: `HEAD` is a git term - the latest commit.

yfinance/screener/screener_query.py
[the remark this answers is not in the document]
ValueRaider: yfinance has never been a minimal Yahoo query tool, it also formats the data, does a little cleaning too. Why not make interaction a little easier?

The cost of debugging EquityQuery("is-in") is zero.

> If the user wants to change a particular part of it later they can, without having to reconstruct the whole thing.

Let's wait for users to actually request this, after this PR gets released.

yfinance/screener/screener.py
R5dan: > `HEAD` is a git term - the latest commit.

I know, I was curious about what you were asking

R5dan:
@ValueRaider Did you edit the `README`?

ValueRaider:
Not here https://github.com/ranaroussi/yfinance/pull/2207/files

PR #2209, opened 2025-01-06
Add config

R5dan:
Adds a new `Config` class with defaults defined

Allows setting of:
- language
- proxy
- request url
- session
- region

Use is global so don't need to set them separately everywhere (for example you most likely would use proxy everywhere and now you just set it once)

Allows creation of custom Configs and to set them as default.

Currently is a draft as only `Ticker.history` and `download` have it implemented

### DIFF
--- a/yfinance/__init__.py
+++ b/yfinance/__init__.py
@@ -19,6 +19,7 @@
 # limitations under the License.
 #
 
+from .config import Current as Config
 from . import version
 from .search import Search
 from .ticker import Ticker
@@ -39,4 +40,4 @@ import warnings
 warnings.filterwarnings('default', category=DeprecationWarning, module='^yfinance')
 
 __all__ = ['download', 'Market', 'Search', 'Ticker', 'Tickers', 'enable_debug_mode', 'set_tz_cache_location', 'Sector',
-           'Industry', 'EquityQuery', 'Screener']
+           'Industry', 'EquityQuery', 'Screener', 'Config']
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -49,7 +49,7 @@ from .const import _BASE_URL_, _ROOT_URL_
 class TickerBase:
     def __init__(self, ticker, config:'_Config'=None):
         self.ticker = ticker.upper()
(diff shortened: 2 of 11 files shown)


R5dan:
@ValueRaider what do you think? This would help with issues like #2210

ValueRaider:
The quantity of code changes concerns me - code bloat. Can it be done with less? I had already been thinking of a related solution for network settings but more like this:

```
yf.set_network(session=X, proxy=Y)
# which passes directly through to YfData singleton
YfData._set_session(X)
YfData._set_proxy(Y)
```

20 lines tops. And consider relaxing on the docstrings - `set_session` is obvious, it sets the session.

(I've not missed lang and region but that is another discussion, let's resolve network stuff first)

R5dan:
> The quantity of code changes concerns me - code bloat

Most of that is `Merge branch 'dev' into config` and updating the inputs (so there isnt an option for `proxy` or `timout` just config



> > Merge branch 'dev' into config
> 
> This will block you squashing commits. Better to rebase.

Yes, I realize that, I was given a button to "update" the pr as I had forgotten to pull the latest dev version, but it didn't do what I thought it would. Will make a new PR if you are happy.

R5dan:
> ```
> yf.set_network(session=X, proxy=Y)
> # which passes directly through to YfData singleton
> YfData._set_session(X)
> YfData._set_proxy(Y)
> ```

Hadn't thought of this method, I would say this makes it harder to expand it though, and doesn't allow for multiple configs for different queries.

ValueRaider:
> doesn't allow for multiple configs for different queries.

True, but unless there is big demand for this, then they can just recall `yf.set_network` different queries. Need to balance against code complexity, this code has to be maintained.

R5dan:
I could have it both return a typed dict with all of the config, that way it can be stored and passed to `yf.set_network` but also doesn't add too much code.

What do you think @ValueRaider?

R5dan:
Also where would you like this to go? I would put it in utils but that would result in a circular import.

R5dan:
Superseded by #2216

Issue #2219, opened 2025-01-12
0.2.51 is not calling Adj Close


SamuelLivingstone:
### Describe bug

0.2.51 is not calling Adj Close, this does work in 0.2.50

### Simple code that reproduces your problem

tickers = ['AAPL']
start_date = '2023-01-01'
end_date = '2024-01-01'

stock = yf.download(tickers, start = start_date, end = end_date)
stock.head()

### Debug log

x

### `yfinance` version

0.2.51

BiggRanger:
This behaviour changed from 0.2.50 and 0.2.51, you need to add 'auto_adjust=False' to the parameters of yf.download.
`data = yfinance.download(ticker, start=dateStart, end=dateEnd, interval='1d', progress=False, auto_adjust=False)`
Once that is done you will get both 'close' and 'adj close' again.

ValueRaider:
Read the release notes (and documentation)

SamuelLivingstone:
> Read the release notes (and documentation)

Fair point, I did miss this, just seems a strange "feature" to introduce (when so many people do not use this as a default).

Thanks for flagging.

SamuelLivingstone:
> > when so many people do not use this as a default
> 
> Really? Who is using the non-adjusted prices? Won't ex-div price drops trigger your algos?

People absolutely should be using adjusted prices. My point however was that people may not realise that the price is now the adjusted price, so it shares the same name as the unadjusted price from an earlier version. 

To be clear and to use an example

- using "close" in 0.2.50 for AAPL for 1980-12-12 would give you **0.128348**
- using "close" in 0.2.52 for AAPL for 1980-12-12 would give you **0.098834** which is the same as "Adj Close" for 0.2.50

In this case, I am more thinking about students and academics who rely on this package in order to access financial data (when those working for large banks or hedge funds will have access via LSEG/Refinitiv, Bloomberg, etc. 

In short, you're reliant on someone using auto_adjust previously. In reality, most people (students) would use df = yf.download('ticker') and then change column names. 

Hope this helps - feel free to message me directly if not :)

BiggRanger:
While it is unfortunate that previous to version 0.2.51 the output was incorrectly labeled and confusing to some, I feel the update in 0.2.51 is the best way forward. I use both the 'Close' and 'Adj Close' in my software and databases and would hate to see 'Close' go away.

>Who is using the non-adjusted prices? Won't ex-div price drops trigger your algos?
[...]

Issue #2214, opened 2025-01-09
Discrepancy Between Stock Prices on Yahoo! Finance and Those Downloaded via yfinance


roku357:
### Describe bug

Is the discrepancy between the stock prices downloaded using yfinance, which have up to six decimal places, and those listed on Yahoo! Finance due to dividend adjustments?

The price before the ex-dividend date does not match the stock price on Yahoo! Finance.

How can I obtain stock price data from yfinance that matches the price listed on Yahoo! Finance, which has not undergone dividend adjustments?

### Simple code that reproduces your problem

t = yf.Ticker("7202.T")
data = t.history(start="2024-09-15", end="2024-10-15")

### Debug log

<html><body>
<div class="stream output-id-18"><div class="output_subarea output_text"><pre>DEBUG    Entering history()
DEBUG:yfinance:Entering history()
DEBUG     Entering history()
DEBUG:yfinance: Entering history()
DEBUG      7202.T: Yahoo GET parameters: {'period1': '2024-09-15 00:00:00+09:00', 'period2': '2024-10-15 00:00:00+09:00', 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG:yfinance:  7202.T: Yahoo GET parameters: {'period1': '2024-09-15 00:00:00+09:00', 'period2': '2024-10-15 00:00:00+09:00', 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG      7202.T: yfinance received OHLC data: 2024-09-17 00:00:00 -&gt; 2024-10-11 00:00:00
DEBUG:yfinance:  7202.T: yfinance received OHLC data: 2024-09-17 00:00:00 -&gt; 2024-10-11 00:00:00
DEBUG      7202.T: OHLC after cleaning: 2024-09-17 09:00:00+09:00 -&gt; 2024-10-11 09:00:00+09:00
DEBUG:yfinance:  7202.T: OHLC after cleaning: 2024-09-17 09:00:00+09:00 -&gt; 2024-10-11 09:00:00+09:00
DEBUG      7202.T: OHLC after combining events: 2024-09-17 00:00:00+09:00 -&gt; 2024-10-11 00:00:00+09:00
DEBUG:yfinance:  7202.T: OHLC after combining events: 2024-09-17 00:00:00+09:00 -&gt; 2024-10-11 00:00:00+09:00
[...]

Issue #2198, opened 2025-01-01
Seems 0.2.51 only returns 5 columns instead of 6 expected


SwissMontainsBear:
### Describe bug

Seems 0.2.51 only downloads 5 columns (open, high, low, close and volume). adjclose seems to be missing.
Did not happen in 0.2.50. Thank you

### Simple code that reproduces your problem

0.2.51 version of yf.download(ticker, '2020-01-01', '2024-12-31') returns 
```
Price	Close	High	Low	Open	Volume
Ticker	CSNDX.SW	CSNDX.SW	CSNDX.SW	CSNDX.SW	CSNDX.SW
Date					
2020-01-03	494.450012	494.799988	490.250000	492.149994	2628
2020-01-06	494.649994	494.649994	488.950012	493.200012	3270
2020-01-07	497.399994	497.950012	495.299988	496.649994	7876
2020-01-08	497.850006	498.299988	494.600006	494.600006	2783
2020-01-09	505.100006	505.399994	502.899994	503.500000	2838
...	...	...	...	...	...
2024-12-19	1217.400024	1224.599976	1212.000000	1217.599976	35453
2024-12-20	1223.199951	1224.199951	1186.599976	1203.000000	2007
2024-12-23	1222.800049	1226.400024	1216.000000	1225.000000	4028
2024-12-27	1225.599976	1242.599976	1219.400024	1240.599976	4358
2024-12-30	1213.800049	1226.800049	1205.000000	1223.599976	55258
```

### Debug log

DEBUG    Entering download()
DEBUG     Disabling multithreading because DEBUG logging enabled
DEBUG     Entering history()
DEBUG      Entering history()
DEBUG       CSNDX.SW: Yahoo GET parameters: {'period1': '2020-01-01 00:00:00+01:00', 'period2': '2024-12-31 00:00:00+01:00', 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG       Entering get()
DEBUG        Entering _make_request()
DEBUG         url=https://query2.finance.yahoo.com/v8/finance/chart/CSNDX.SW
DEBUG         params=frozendict.frozendict({'period1': 1577833200, 'period2': 1735599600, 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'})
DEBUG         Entering _get_cookie_and_crumb()
DEBUG          cookie_mode = 'basic'
DEBUG          Entering _get_cookie_and_crumb_basic()
[...]

PR #2216, opened 2025-01-10
Config

R5dan:
Supersedes #2209 with better implementation.

Changes:
- Deprecates all `proxy`, `session` and `timeout` in functions
- Adds `yf.set_config`
- Allows setting:
    - `proxy`
    - `session`
    - `timout`
    - `url`
    - `language`
    - `region`
- Adds deprecator for future use 
- Adds docs for use of `yf.set_config`

PR #2230, opened 2025-01-19
Warn user when use download() without specifying auto_adjust

### LINKED ISSUE #2225: Data Anomaly in Downloaded Price Data (Before 2024-12-20) Detected Since Last Week
May I ask, is there anyone else like me who has found that the price data before 2024-12-20 of the downloaded data has been incorrect since last week?
For example:

!pip install yfinance
import datetime as dt
import yfinance as yf 

stock_symbol='AOA'
data_years=0.5
end_date=dt.date.today()
start_date=end_date-dt.timedelta(data_years*365)

stock_data=yf.download(stock_symbol,start_date,end_date) 

pd.set_option('display.max_rows', None)

print(stock_data)

ValueRaider:
Warn users that `download` changed `auto_adjust` to `True`.

Resolve #2225 #2219 #2214 #2198

### DIFF
--- a/yfinance/multi.py
+++ b/yfinance/multi.py
@@ -36,7 +36,7 @@ from . import shared
 
 @utils.log_indent_decorator
 def download(tickers, start=None, end=None, actions=False, threads=True,
-             ignore_tz=None, group_by='column', auto_adjust=True, back_adjust=False,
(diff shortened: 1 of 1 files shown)


R5dan:
Once #2216 is merged, I think that this should be moved to that instead of just printing

PR #2237, opened 2025-01-23
Update search

R5dan:
- Adds functions to `Search` which updates only that field:
	- `search_quotes`
	- `search_news`
	- `search_lists`
	- `search_nav`
	- `search_research`
	
- Adds a cache to `Search`
- Moves `fetch` to `SearchCache`


ValueRaider:
What does `SearchCache` do that can't be done with `request_cache`?

R5dan:
> What does `SearchCache` do that can't be done with `request_cache`?

To be honest I'd forgotten about it, but looking at it now:

- `requests-cache` is a extra, where I think this should be a core functionality as this is more so to save time in the request as the actual response won't change for a few days - (months or years for tickers)

Also why is `nospam` and extra? I feel like it should be baked in functionality

ValueRaider:
It's extra because not required to fetch data from Yahoo.

R5dan:
@ValueRaider how is this?

PR #2238, opened 2025-01-23
Update setup.py

R5dan:
- Fixes `setup.py`
- Adds new install command to properly print message
- Prints any warnings for specific versions (would have been useful for the last version with `Screener` being redone) - trying to stop issues when in docs for big overhauls.

**This is a draft as there is very little documentation for the actual `install` command, so it needs fixing. Any help welcome.**

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,6 @@ with io.open(path.join(here, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
 WARNINGS:'dict[str,str]' = {
-    "0.2.51":"TEST"
 }
 
 class CustomInstall(install):
--- a/setup.py
+++ b/setup.py
@@ -7,9 +7,17 @@
 """yfinance - market data downloader"""
 
 from setuptools import setup, find_packages
+from setuptools.command.install import install
 # from codecs import open
 import io
 from os import path
+
+MESSAGE = """
+NOTE: yfinance is not affiliated, endorsed, or vetted by Yahoo, Inc.
+
+You should refer to Yahoo!'s terms of use for details on your rights
+to use the actual data downloaded.
+"""
 
 # --- get version ---
 version = "unknown"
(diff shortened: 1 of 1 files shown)


dhruvan2006:
> * Fixes `setup.py`

What was broken?

R5dan:
> What was broken?

The message/disclaimer never printed
I also changed it to install `pandas[html]` instead of `pandas`+ everything needed for `pandas[html]`

setup.py
```diff
@@ -55,15 +68,19 @@
...
+        'Programming Language :: Python :: 3.11',
+        'Programming Language :: Python :: 3.12',
+        'Programming Language :: Python :: 3.13',
+
     ],
     platforms=['any'],
     keywords='pandas, yahoo finance, pandas datareader',
     packages=find_packages(exclude=['contrib', 'docs', 'tests', 'examples']),
-    install_requires=['pandas>=1.3.0', 'numpy>=1.16.5',
+    install_requires=['pandas[html]>=1.3.0', 'numpy>=1.16.5',
```
ValueRaider, line 79: nice

setup.py
```diff
@@ -75,9 +92,3 @@
         ],
     },
 )
-
-print("""
-NOTE: yfinance is not affiliated, endorsed, or vetted by Yahoo, Inc.
-
-You should refer to Yahoo!'s terms of use for details on your rights
-to use the actual data downloaded.""")
```
ValueRaider, line 83: why delete? was it appearing?

ValueRaider:
> This is a draft as there is very little documentation for the actual install command, so it needs fixing.

What are you trying to do (with `Install`)?

R5dan:
> What are you trying to do (with `Install`)?

Currently, I believe that it simply uses the default install command, as it (1) doesn't print to the terminal and (2) I have tried getting it to write to a file and it doesn't do it.

The only documentation is for the `Command` class (which `Install` is a subclass of) but I want to use install so it actually installs properly

setup.py
R5dan: It doesn't appear on install, hence why I am trying to put it in the custom command

PR #2248, opened 2025-01-30
Screener

R5dan:
These are some of my ideas for the screener class.
I thought it would be easier to put them in a PR then to try to explain them
Will remove any as wanted.

Changes:
- Better type hints
- Moves `is-in` to classmethod
- Doesn't import all of `screener` - have to go `from yfinance.screener import ...`
- Added `Query` class to control entire query
- Added `QueryHead` class for easy access to `region` `exchange` `sector`
- Added `screen` to queries - allows screening from there


ValueRaider:
Can you respond to this https://github.com/ranaroussi/yfinance/pull/2175#issuecomment-2599797944

R5dan:
Sorry just seen, I will fix and make a new PR.

PR #2252, opened 2025-01-31
add option to keep the timestamps provided by Yahoo

ofenloch:
I added parameter keep_timestamps to method `PriceHistory.history` and to function  `parse_quotes` in file utils.py.

If keep_timestamp is True, the original timestamps are added as an extra column to the DataFrame returned by `parse_quotes`. keep_timestamps defaults to False, so the default behaviour is not changed.

I also added type specifiers in `parse_quotes` and eliminated a violation of rule python:S6734.

### DIFF
--- a/yfinance/scrapers/history.py
+++ b/yfinance/scrapers/history.py
@@ -31,7 +31,7 @@ class PriceHistory:
                 start=None, end=None, prepost=False, actions=True,
                 auto_adjust=True, back_adjust=False, repair=False, keepna=False,
                 proxy=None, rounding=False, timeout=10,
-                raise_errors=False) -> pd.DataFrame:
+                raise_errors=False, keep_timestamps=False) -> pd.DataFrame:
         """
         :Parameters:
             period : str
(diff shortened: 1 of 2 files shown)


ValueRaider:
What scenario is this for?

For intraday data you can get identical values with `df.index.astype('int64')//10**9`. That's different for daily/weekly but then why do you want times?

ofenloch:
I read about the timezone issues in previous releases. When I store the values in a database the tz confusion gets even worse. Thus, I wanted to have the option to keep only the timestamps. The timestamps are the source of truth.

But I couldn't figure out a way to replace the Pandas.Timestamps completely. To be on the safe side of the tracks I store both - UNIX time and Pandas.Timestamps - in the downloaded files.

ValueRaider:
> I read about the timezone issues in previous releases

I fixed this few years ago by setting timezone on the index, I've not seen any problems since.

ofenloch:
OK, you're right. I'm closing this PR.

Issue #2251, opened 2025-01-31
Incorrect output for YFxxx exceptions


detlefm:
### Describe bug

For YFxxx exceptions the output is:

```
1 Failed download:
['QUATSCH']: YFPricesMissingError('$%ticker%: possibly delisted; no price data found  (period=1d) (Yahoo error = "No data found, symbol may be delisted")')

```
I suspect the intended output is

```
1 Failed download:
YFPricesMissingError('$['QUATSCH']: possibly delisted; no price data found  (period=1d) (Yahoo error = "No data found, symbol may be delisted")')
```
The error is in multi.py, function download
instead of
```
for err in errors.keys():
    logger.error(f'{errors[err]}: ' + err)
```
it should be
```
for err in errors.keys():
    logger.error(err.replace('%ticker%',errors[err])
```

### Simple code that reproduces your problem

```
import yfinance as yf

df = yf.download(tickers=["MSF.DE", "QUATSCH"], period="1d", progress=False)
```

### Debug log

DEBUG    Entering download()
DEBUG     Disabling multithreading because DEBUG logging enabled
DEBUG     Entering history()
DEBUG      Entering history()
DEBUG       MSF.DE: Yahoo GET parameters: {'range': '1d', 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG       Entering get()
DEBUG        Entering _make_request()
DEBUG         url=https://query2.finance.yahoo.com/v8/finance/chart/MSF.DE
DEBUG         params={'range': '1d', 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG         Entering _get_cookie_and_crumb()
DEBUG          cookie_mode = 'basic'
DEBUG          Entering _get_cookie_and_crumb_basic()
DEBUG           loaded persistent cookie
DEBUG           reusing cookie
DEBUG           crumb = '0DlVzIFi41Z'
DEBUG          Exiting _get_cookie_and_crumb_basic()
DEBUG         Exiting _get_cookie_and_crumb()
DEBUG         response code=200
DEBUG        Exiting _make_request()
DEBUG       Exiting get()
DEBUG       MSF.DE: yfinance received OHLC data: 2025-01-31 16:36:23 only
DEBUG       MSF.DE: OHLC after cleaning: 2025-01-31 17:36:23+01:00 only
[...]

PR #2256, opened 2025-02-02
Fix: Interval check and error message formatting in `multi.py`

dhruvan2006:
### Changes
- Fixed interval checking logic, intervals like `60m` and `90m` were previously not recognized.
- Updated error message formatting #2251 
```python
[*********************100%***********************]  27 of 27 completed

2 Failed downloads:
['AISUDFYHAOISDUFHA', 'ASDFSF']: YFTzMissingError('possibly delisted; no timezone found')
```


ValueRaider:
Just to check, what happens with single ticker?

PR #2260, opened 2025-02-09
Improve prices div repair

ValueRaider:
Extend div-repair to intraday
Increase lookback for "div date wrong"
Slightly loosen default "false threshold" for cluster of mixed False/True

Related:
Stop checking for down-changes after negative adj-close Handle avg-volume=0%

### DIFF
--- a/yfinance/scrapers/history.py
+++ b/yfinance/scrapers/history.py
@@ -1304,7 +1304,7 @@ class PriceHistory:
 
         if df is None or df.empty:
             return df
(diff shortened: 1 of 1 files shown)


ValueRaider:
> FAILED tests/test_ticker.py::TestTickerEarnings::test_earnings_dates_with_limit - requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

I didn't change anything relevant so the test workflow is bad.

PR #2266, opened 2025-02-16
Added custom error message

oddblaster:
# First time contribution 
Just doing this to change the modified message for errors when downloading multiple yfTickers. Don't want to change anything really, just playing around with the code and learning to find how this work. First time open source contributor and wanting to get into finance. Don't have to accept this pull request lmao.


paulmcq (changes requested):
1. This PR inserts blank lines in places that do not need them.
2. yfinance/data.py line 499: remove trailing spaces
3. line 428 looks like a debug print that does not belong in a PR
4. yfinance/multi.py line 194: prefer you just fix the f-string, rather than using string.replace

PR #2273, opened 2025-02-17
Add industries

### LINKED ISSUE #2272: Screening by industries
Hi!
Is it possible to add a feature for screening by industries?
Currently, I do it like this:
```
   from yfinance.const import EQUITY_SCREENER_EQ_MAP, EQUITY_SCREENER_FIELDS
   from yfinance import EquityQuery
   sector = "Basic Materials"
   industry = "Chemicals"
   EQUITY_SCREENER_FIELDS.get("eq_fields").update(["industry"])
   EQUITY_SCREENER_EQ_MAP.update({"industry": f"{industry}"})
   query =  EquityQuery("and", [EquityQuery("is-in", ["sector", f"{sector}"]), EquityQuery("is-in", ["industry", f"{industry}"]), EquityQuery("is-in", ["exchange", "NYQ", "NMS"])])
   resp = yf.screen(query)
```
But it looks a bit awkward, doesn't it?

R5dan:
- Adds a function for correct `industries`
- Allows `screen` and `QueryBase` to both accept dicts as operands
- Adds slightly better type hinting

Fixes #2272


R5dan:
Added docs to this. Need to fix them a bit though as the layout is off.
The unused import is deliberate as it shows it is also an option. @ValueRaider is it possible to disable ruff for `docs/`?

ggeeoorrgg:
As I can see, for the screener, you use SECTOR_INDUSTRY_MAPPING for industry names in the same way as it is used for Sector. However, I’m afraid those names differ from the ones required by the screener.
I did some research and found that the names must be specified exactly as they appear on the site, with one exception: if an industry name contains an em dash, there should be no whitespace between the em dash and the surrounding text.
Meanwhile, the current documentation specifies different sector names for the screener (see here: (https://ranaroussi.github.io/yfinance/reference/api/yfinance.EquityQuery.htm) and for Sector (see here: https://ranaroussi.github.io/yfinance/reference/api/yfinance.Sector.html).
Moreover, while Yahoo Finance requires selecting a sector before choosing an industry on the website, this is not the case with the API. The API allows specifying only industry names in a query without selecting a sector.
Therefore, it should be sufficient to define the correct industry names e.g. in a list.

PR #2286, opened 2025-02-19
fix(quote): handle None response from additional info fetch

skoenig:
A simple safeguard against empty fetch results.


ValueRaider:
You've also changed when user receives data from `_fetch_additional_info`

ValueRaider:
I'll spell it out. When info is none, user doesn't get the additional info even if it exists.

skoenig:
Hi @ValueRaider, do I understood you correctly that the desired behavior should be that either the return value of `self._fetch(proxy, modules=modules)` or `self._fetch_additional_info(proxy)` should be the base for further information gathering?

ValueRaider:
Yes, but already done #2382

PR #2298, opened 2025-02-20
Add Potential Market Closure Error

R5dan:
I want to add an error for when there is no data as the market is closed

Currently I have:
- `Ticker` contains its current `Market`
- `Market` now contains it name in the class

However I was going to implement this with the `Maket.status` but that doesn't have any way to get a custom time / date.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -39,6 +39,7 @@ from .scrapers.holders import Holders
 from .scrapers.quote import Quote, FastInfo
 from .scrapers.history import PriceHistory
 from .scrapers.funds import FundsData
+from .domain.market import Market
 
 from .const import _BASE_URL_, _ROOT_URL_, _QUERY1_URL_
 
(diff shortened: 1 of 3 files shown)


ValueRaider:
Just FYI `history_metadata` contains exchange hours, but maybe only for intraday intervals.

R5dan:
Yes, I know. That was another that I looked at, and will just implement this type of erroring if you want. I was kind of looking for market holidays as well. What do you think? @ValueRaider

ValueRaider:
I'm not against adding https://pypi.org/project/holidays

ValueRaider:
https://pypi.org/project/holidays

R5dan:
I did not know about that, thanks

R5dan:
@ValueRaider This seemed to be ideal, however they only support 7 markets, making it unfeasible to use.

PR #2320, opened 2025-02-23
Direct TTM Financial Data Retrieval Feature

JanMkl:
This PR completes the earlier #1643 and moves the PR in to the dev branch.

- Fetches the TTM values for cash flow and income statement from Yahoo.
- Added tests test_ttm_income_statement and test_ttm_cash_flow.

Changes to version.py removed from this commit.

This feature was also requested in #2287

Issue #2287, opened 2025-02-19
Direct TTM Financial Data Retrieval Feature


victoryg739:
I’m currently calculating Trailing Twelve Months (TTM) data by summing the last four quarters. However, when a quarter’s data is missing (e.g., with SMCI), this method becomes inaccurate.

Could we have a feature to fetch TTM financial data directly?

amigobrewbrew:
I have been using this in the meantime: https://github.com/init-js/yfinance

ValueRaider:
They submitted a PR but abandoned it, so if someone wants to complete it ...
https://github.com/ranaroussi/yfinance/pull/1643

JanMkl:
I created a PR that implements direct retrieval from Yahoo #2320 

This PR is based on the latest version in dev branch.

JanMkl:
In PR #2321 correctly against the dev branch

JanMkl:
@victoryg739 #2321 was merged. I think this issue can be now closed.

PR #2321, opened 2025-02-23
Direct TTM Financial Data Retrieval Feature

JanMkl:
This PR completes the earlier #1643 and moves the PR in to the dev branch.

- Fetches the TTM values for cash flow and income statement from Yahoo.
- Added tests test_ttm_income_statement and test_ttm_cash_flow.
- Transferring commit to dev branch

This feature was also requested in #2287

Obsoletes PR #2320

### DIFF
--- a/doc/source/reference/yfinance.financials.rst
+++ b/doc/source/reference/yfinance.financials.rst
@@ -10,12 +10,16 @@ Financials
 
    get_income_stmt
    income_stmt
+   quarterly_income_stmt
+   ttm_income_stmt
 
    get_balance_sheet
    balance_sheet
 
    get_cashflow
    cashflow
+   quarterly_cashflow
+   ttm_cashflow
 
    get_earnings
    earnings
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -37,10 +37,12 @@ ticker_attributes = (
     ("recommendations", Union[pd.DataFrame, dict]),
     ("recommendations_summary", Union[pd.DataFrame, dict]),
     ("upgrades_downgrades", Union[pd.DataFrame, dict]),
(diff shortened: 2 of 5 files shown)


ValueRaider:
Discard the mostly-nan columns, more that one column doesn't make sense with TTM.

Then open question about the user interface. Separate function like you've done? New argument on existing fetchers?

JanMkl:
Just double checking what you mean by discarding NaN. The ttm_cash_flow returns only the TTM column as Yahoo only returns one column. You would like to see same for the ttm_income_stmt where yahoo currently returns the TTM with all the values and then preceding 4 quarters with only some values and mostly NaN?

Meaning that we should implement a special if block in the get_financials_time_series to check if it was called with trailing timescale then it should return only one column and skip the rest (different behaviour than yearly and quarterly time scales)?

JanMkl:
@ValueRaider Added now in get_financials_time_series that it drops the NaN columns and only returns exactly one column. Is this what you meant by removing the most-nan columns?

amigobrewbrew:
@JanMkl @ValueRaider I would do separate functions like in PR: ttm_cashflow and ttm_income_stmt. I did a quick test seems to be good like this!

JanMkl:
@ValueRaider squashed into single commit.

ValueRaider:
I almost forgot, can you update the web docs: https://github.com/ranaroussi/yfinance/blob/main/doc/source/reference/yfinance.financials.rst

Instructions in guide

JanMkl:
@ValueRaider Added quarterly_income_stmt, ttm_income_stmt, quaterly_cashflow, ttm_cashflow to the docs. I added also the quarterly ones as those seemed to be missing from the docs although implemented.

Let me know if you still see need for further modifications.

PR #2345, opened 2025-03-06
Fix: History caching

### LINKED ISSUE #2344: Some Ticker APIs get empty results if called after get_history_metadata()
```python
import yfinance as yf

msft = yf.Ticker("MSFT")
# msft.get_history_metadata()
print(msft.get_splits())
```
The code above can ouput the correct result of splits, but after I added `get_history_metadata()` before `get_splits()` like the following code, the result became an empty pd.Series.

```python
import yfinance as yf

msft = yf.Ticker("MSFT")
msft.get_history_metadata()
print(msft.get_splits())

```

Does anyone meet the same problem, or am I missed something?

My yfinance version is 0.2.54.

dhruvan2006:
#### Changes
- `history()` now only caches data if the period is `max`
- Renamed `_history` to `_max_history` to ensure clarity
- Added a test to ensure proper behavior for chained history calls

Closes #2344

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -332,6 +332,12 @@ class TestTickerHistory(unittest.TestCase):
         self.assertIsInstance(data, pd.DataFrame, "data has wrong type")
         self.assertFalse(data.empty, "data is empty")
 
+    def test_chained_history_calls(self):
+        _ = self.ticker.history(period="2d")
+        data = self.ticker.dividends
+        self.assertIsInstance(data, pd.Series, "data has wrong type")
+        self.assertFalse(data.empty, "data is empty")
+
 
 class TestTickerEarnings(unittest.TestCase):
     session = None
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -428,17 +428,17 @@ class TickerBase:
     def get_cashflow(self, proxy=None, as_dict=False, pretty=False, freq="yearly"):
         return self.get_cash_flow(proxy, as_dict, pretty, freq)
 
-    def get_dividends(self, proxy=None) -> pd.Series:
-        return self._lazy_load_price_history().get_dividends(proxy)
+    def get_dividends(self, proxy=None, period="max") -> pd.Series:
+        return self._lazy_load_price_history().get_dividends(period=period, proxy=proxy)
 
-    def get_capital_gains(self, proxy=None) -> pd.Series:
-        return self._lazy_load_price_history().get_capital_gains(proxy)
+    def get_capital_gains(self, proxy=None, period="max") -> pd.Series:
+        return self._lazy_load_price_history().get_capital_gains(period=period, proxy=proxy)
 
-    def get_splits(self, proxy=None) -> pd.Series:
-        return self._lazy_load_price_history().get_splits(proxy)
+    def get_splits(self, proxy=None, period="max") -> pd.Series:
+        return self._lazy_load_price_history().get_splits(period=period, proxy=proxy)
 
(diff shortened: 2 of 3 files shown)


ValueRaider:
Not keen on default max.

Suggestions:
- add new argument `period` to `get_splits` and `get_dividends`, default 2y
- replace `self._max_history = None` with `self._history_cache = {}`, dict key is `(interval, period)`
- move cache R/W to a new middle function, so flow is: get_splits() -> self._history_cache() -> self.history()

Hope that makes sense

R5dan:
@ValueRaider what about if you have different days? I wouldn't want the cache for data today to be returned for the data yesterday or a week ago etc

ValueRaider:
> @ValueRaider what about if you have different days? I wouldn't want the cache for data today to be returned for the data yesterday or a week ago etc

yfinance already does via `cache_get` that and no one complains. This is memory cache not persistent.

dhruvan2006:
> Not keen on default max.

I was unsure on whether I should change default behavior? People using `ticker.splits` `ticker.dividends` might have code break cause they relied on the default being max

ValueRaider:
@dhruvan2006 good point, I missed the default is already max. max it is

JanMkl:
@dhruvan2006 if we want to refactor the caching behaviour, I believe it is not enough to create a history and history_max as also the metadata is different for different time periods. Exchange schedules (tradingPeriods) are only returned in metadata that is intraday and that is why get_history_metadata depends on intraday data.

Perhaps a more future proof change would be to refactor the structure so that _history is actually an array containing elements named data and metadata. Array keys could be the valid time periods.

But as said this is a larger refactor. For a simple solution just addressing #2344 I created a PR #2346 but if this rework is seen completed both for the data and metadata then my PR can be discarded.

ValueRaider:
@JanMkl covered in https://github.com/ranaroussi/yfinance/pull/2345#issuecomment-2704718120

ValueRaider:
@dhruvan2006 bump https://github.com/ranaroussi/yfinance/pull/2345#issuecomment-2704718120

dhruvan2006:
- @ValueRaider Implemented your data flow using `_get_history_cache()`. 

- `history()` does no caching now.

ValueRaider:
> Unsure whether prepost=True matters for fetching metadata

It does, it adds more columns to `dat.history_metadata['tradingPeriods']`:
```
                                          pre_start                   pre_end                     start                       end                post_start                  post_end
Date                                                                                                                                                                                 
2025-03-11 00:00:00-04:00 2025-03-11 04:00:00-04:00 2025-03-11 09:30:00-04:00 2025-03-11 09:30:00-04:00 2025-03-11 16:00:00-04:00 2025-03-11 16:00:00-04:00 2025-03-11 20:00:00-04:00
...
```

ValueRaider:
bug with cache_key

dhruvan2006:
> bug with cache_key in _get_history_cache

what is it?

ValueRaider:
@JanMkl yes, keep scope focused on these get_ functions.

There can be a debate whether this is needed at all - yfinance started with `self._history` but now it has
https://github.com/ranaroussi/yfinance/blob/fa3094daedf4a69e8af542e4bb171cbd3bd61862/yfinance/data.py#L413-L415

JanMkl:
lgtm

ValueRaider:
Just, try to test whether the cache is being hit with `yf.enable_debug_mode()`

dhruvan2006:
> It's obvious, surprised you missed it.

Fixed

PR #2346, opened 2025-03-07
Fixes issue 2344 empty result for get splits

JanMkl:
Simple fix for #2344. Checks the current history that it contains the timescale we expect for these functions if not refetch the history.
- Added check in get_history_metadata that if cached metadata doesn't contain trading periods then fetch intra day history.
- Added check that if cached history is not max period then refetch the history for get_dividends, get_capital_gains, get_splits and get_actions.

This pull request can be discarded if the larger rework #2345 is completed.

### DIFF
--- a/tests/test_prices.py
+++ b/tests/test_prices.py
@@ -453,6 +453,59 @@ class TestPriceHistory(unittest.TestCase):
 
         dat.history(start=start, end=end, interval=interval)
 
+    def test_get_history_metadata(self):
+        tkrs = ["AAPL", "INTC", "FORTUM.HE"]
+        periods = ["pre_start", "pre_end", "start", "end", "post_start", "post_end"]
+
+        for tkr in tkrs:
+            metadata = yf.Ticker(tkr).get_history_metadata()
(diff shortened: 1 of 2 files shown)


JanMkl:
Added some tests for basic validation of the data received from get_history_metadata, get_dividends, get_capital_gains, get_splits and get_actions.

If this PR is decided to be merged I can squash these to a single commit but until then I'll let it be two separate commits as the test could be cherry picked to rework #2345 if/when that is completed.

dhruvan2006:
@JanMkl I have implemented the caching mechanism in #2345. You can have a gander

JanMkl:
Closing this as #2345 is merged.

@ValueRaider do you see any value adding the tests on this PR to project? I can open a separate PR for those if you see them as useful.

dhruvan2006:
There's already a few tests here, maybe you can expand them?

https://github.com/ranaroussi/yfinance/blob/fa3094daedf4a69e8af542e4bb171cbd3bd61862/tests/test_ticker.py#L318-L331

Issue #2343, opened 2025-03-04
Error get stock data for EXTO


tamireiny:
### Describe bug

Hello, I have tried to get the stock data of ticker EXTO for example: 

ticker_data = yfinance.Ticker(ticker)
and when I run : 
ticker_info_data = ticker_data .info

I get the following error:

Traceback (most recent call last):
  File "G:\stock_miner_v1.0\venv_python_3p11\Lib\site-packages\IPython\core\interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-1-41b3e7b9f323>", line 1, in <module>
    ticker_info.info
  File "G:\stock_miner_v1.0\venv_python_3p11\Lib\site-packages\yfinance\ticker.py", line 159, in info
    return self.get_info()
           ^^^^^^^^^^^^^^^
  File "G:\stock_miner_v1.0\venv_python_3p11\Lib\site-packages\yfinance\base.py", line 232, in get_info
    data = self._quote.info
           ^^^^^^^^^^^^^^^^
  File "G:\stock_miner_v1.0\venv_python_3p11\Lib\site-packages\yfinance\scrapers\quote.py", line 508, in info
    self._fetch_info(self.proxy)
  File "G:\stock_miner_v1.0\venv_python_3p11\Lib\site-packages\yfinance\scrapers\quote.py", line 618, in _fetch_info
    result[quote]["result"][0]["symbol"] = self._symbol
    ~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range

### Simple code that reproduces your problem

ticker = "EXTO"
ticker_data = yfinance.Ticker(ticker)
ticker_info_data = ticker_data .info

### Debug log from yf.enable_debug_mode()

Traceback (most recent call last):
  File "G:\stock_miner_v1.0\venv_python_3p11\Lib\site-packages\IPython\core\interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-1-41b3e7b9f323>", line 1, in <module>
    ticker_info.info
  File "G:\stock_miner_v1.0\venv_python_3p11\Lib\site-packages\yfinance\ticker.py", line 159, in info
    return self.get_info()
           ^^^^^^^^^^^^^^^
  File "G:\stock_miner_v1.0\venv_python_3p11\Lib\site-packages\yfinance\base.py", line 232, in get_info
[...]

PR #2354, opened 2025-03-11
Fixes issue 2343

JanMkl:
- Check that result is not zero length before appending it to query_info

This resolves issue #2343

### DIFF
--- a/yfinance/scrapers/quote.py
+++ b/yfinance/scrapers/quote.py
@@ -614,11 +614,11 @@ class Quote:
 
         query1_info = {}
         for quote in ["quoteSummary", "quoteResponse"]:
-            if quote in result:
+            if quote in result and len(result[quote]["result"]) > 0:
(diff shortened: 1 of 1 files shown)


yfinance/scrapers/quote.py
```diff
@@ -615,14 +615,15 @@ def _fetch_info(self, proxy):
...
             if quote in result:
-                result[quote]["result"][0]["symbol"] = self._symbol
-                query_info = next(
-                    (info for info in result.get(quote, {}).get("result", [])
-                     if info["symbol"] == self._symbol),
-                    None,
-                )
-                if query_info:
-                    query1_info.update(query_info)
+                if len(result[quote]["result"]) > 0:
```
ValueRaider, line 618: Combine with an `and`

JanMkl: Done.

ValueRaider:
Final request: rename PR title to something more specific (but still concise). Helps collating release notes.

PR #2379, opened 2025-03-23
Fixed bug caused by applying .update on potential NoneType.

GimleLarpes:
Fixed a bug where result.update was called before checking whether result is NoneType.


ValueRaider:
duplicate #2286

Issue #2363, opened 2025-03-15
IndexError for on yf.Ticker('NEPT').info.


Fotoblysk:
### Describe bug

The error is likely here as result is empty:
```yfinance/scrapers/quote.py
        query1_info = {}
        for quote in ["quoteSummary", "quoteResponse"]:
            if quote in result:
                result[quote]["result"][0]["symbol"] = self._symbol # MY_COMENT - HERE RESULT IS [] FOR quote="quoteResponse"
                query_info = next(
                    (info for info in result.get(quote, {}).get("result", [])
                     if info["symbol"] == self._symbol),
                    None,
                )
                if query_info:
                    query1_info.update(query_info)
```

### Simple code that reproduces your problem

import yfinance as yf
print(yf.__version__)
print(yf.Ticker('NEPT').info)

### Debug log from yf.enable_debug_mode()

Not needed.

### Bad data proof

Not bad data just bad exception handling

### `yfinance` version

0.2.54

### Python version

3.12.3

### Operating system

Ubuntu

dhruvan2006:
Should be fixed by #2354. Can you check?

PR #2382, opened 2025-03-24
Fixes issue 2343 and 2363 Empty result and QuoteResponse

JanMkl:
Reintroduced changes from #2286 and added test case that covers both #2343 and #2363

#2363 has the same issues as this #2343 had but it seems that Yahoo doesn't anymore return any info for 'EXTO' while for 'NEPT' it returns result without 'quoteResponse' key which was the behaviour when PR #2354 was introduced.

In short the changes in this commit:
- Check that if result from _fetch was None then don't try to fetch additional info.
- Added test_empty_info test case to protect if the behavior in Yahoo changes.


ValueRaider:
> Check that if result from _fetch was None then don't try to fetch additional info.

Why is everyone doing that? Why not return the additional info if it exists?

paulmcq (commented):
line 1024 typo, remove final 'n': trailingPegRatio**n**

JanMkl:
@ValueRaider you mean that let it fall through all the way and fill the result with whatever Yahoo returns?

@paulmcq thanks for catching the typo.

There is on the Readme one change that is done one main branch but not in Dev and when I changed the PR branch dev as I noticed this. It picked up. @ValueRaider you can delete that from this PR if you don't want the README change to be merged to dev.

ValueRaider:
To solve the README issue you have to also rebase #1084

JanMkl:
Rebased and dropped the README commits.

PR #2388, opened 2025-03-26
Fixes issues 2301 and 2383 AttributeError: module 'requests.cookies' has no attribute 'update'

JanMkl:
- In utils.py line 196 if the session is None then _requests is mistakenly passed to as session to data.py which causes attribute error: module 'requests.cookies' has no attribute 'update'
- The above condition happens only when searching tickers with ISIN numbers
- Added tests with ISIN numbers
- Added tickerBase to raise an ValueError if an empty tickername is passed or isin search results in to empty ticker. This is to have consistent behaviour with utils.get_all_by_isin()

This PR requires that PR #2382 is first merged otherwise new tests introduced for ISIN will fail to same error as #2343 and #2363

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -61,9 +61,16 @@ class TickerBase:
         self._earnings = None
         self._financials = None
 
+        # raise an error if user tries to give empty ticker
+        if self.ticker == "":
+            raise ValueError("Empty ticker name")
+
         # accept isin as ticker
         if utils.is_isin(self.ticker):
+            isin = self.ticker
             self.ticker = utils.get_ticker_by_isin(self.ticker, None, session)
+            if self.ticker == "":
+                raise ValueError(f"Invalid ISIN number: {isin}")
 
         self._data: YfData = YfData(session=session)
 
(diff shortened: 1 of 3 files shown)


JanMkl:
@ValueRaider can you check this?

ValueRaider:
#2382 now merged. You can rebase this branch to freshly-pulled dev to "pull in" fix.

yfinance/base.py
```diff
@@ -61,9 +61,15 @@ def __init__(self, ticker, session=None, proxy=None):
...
 
+        # raise an error if user tries to give empty ticker
+        if self.ticker == "":
+            raise ValueError("Empty ticker name")
+
         # accept isin as ticker
         if utils.is_isin(self.ticker):
             self.ticker = utils.get_ticker_by_isin(self.ticker, None, session)
+            if self.ticker == "":
+                raise ValueError("Invalid ISIN number")
```
ValueRaider, line 72: Add value to error message e.g.
> raise ValueError(f"Invalid ISIN value '{ticker_copy}'")

ValueRaider (changes requested):

yfinance/base.py
JanMkl: Changed in both base.py and test_ticker.py

ValueRaider:
conflicts

PR #2409, opened 2025-04-22
Fix: Result of None Type doesn't have update() attribute

rohan-raut:
## Issue: 
Sometime _fetch() method returns None and before checking the NoneType an attribute of the _fetch() result is called. This breaks the code when get_info() method is called of Tickers which contains multiple stocks.

## Fix:
- Moved the result NoneType check before accessing its update() attribute in /scrapers/quote.py file


ValueRaider:
Already fixed in dev branch #1084

Issue #2419, opened 2025-04-28
Screener Is Broken


deeleeramone:
### Describe bug

"size" and "offset" no longer work using the Screener class. This isn't a Yahoo API thing because they work directly from the API. The library seems to be making a GET request, but the actual Yahoo API expects a POST.

This works:

```
    _data = YfData(session=session)
    response = _data.post(
        "https://query2.finance.yahoo.com/v1/finance/screener",
        body=body,
        user_agent_headers=_data.user_agent_headers,
        params=params_dict,
```

### Simple code that reproduces your problem

```python

In [6]: import yfinance as yf

In [7]: response = yf.screen("most_actives", size=250)

In [8]: len(response["quotes"])
Out[8]: 25

In [9]: response["total"]
Out[9]: 153

In [10]: response = yf.screen("most_actives", size=100)

In [11]: len(response["quotes"])
Out[11]: 25
```

### Debug log from yf.enable_debug_mode()

DEBUG    Using User-Agent: Mozilla/5.0 (X11; Linux i686; rv:135.0) Gecko/20100101 Firefox/135.0
DEBUG    Entering get()
DEBUG     Entering _make_request()
DEBUG      url=https://query2.finance.yahoo.com/v1/finance/screener/predefined/saved
DEBUG      params={'corsDomain': 'finance.yahoo.com', 'formatted': 'false', 'lang': 'en-US', 'region': 'US', 'scrIds': 'most_actives', 'size': 250}
DEBUG      Entering _get_cookie_and_crumb()
DEBUG       cookie_mode = 'basic'
DEBUG       Entering _get_cookie_and_crumb_basic()
DEBUG        loaded persistent cookie
DEBUG        reusing cookie
DEBUG        crumb = 'UiHlByh5Gyz'
DEBUG       Exiting _get_cookie_and_crumb_basic()
DEBUG      Exiting _get_cookie_and_crumb()
DEBUG      response code=200
DEBUG     Exiting _make_request()
DEBUG    Exiting get()

### `yfinance` version

0.2.56

ValueRaider:
>  This isn't a Yahoo API thing because they work directly from the API

It is a Yahoo API thing.

https://github.com/ranaroussi/yfinance/blob/13b26a99d04bdb79ac13a18bba7940165560255e/yfinance/screener/screener.py#L115-L125

deeleeramone:
> > This isn't a Yahoo API thing because they work directly from the API
> 
[...]

PR #2425, opened 2025-04-30
Fix Screener

R5dan:
This is to `main` branch as it is a major bug fix.

Yahoo changed from `size` to `count` param and are now using the `Query1` url. This pr reflects those changes

Closes:
#2419 
and maybe some more.

@ValueRaider

### DIFF
--- a/yfinance/screener/screener.py
+++ b/yfinance/screener/screener.py
@@ -2,7 +2,7 @@ from .query import EquityQuery as EqyQy
 from .query import FundQuery as FndQy
 from .query import QueryBase, EquityQuery, FundQuery
 
-from yfinance.const import _BASE_URL_, _SENTINEL_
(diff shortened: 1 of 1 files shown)


ValueRaider:
Can't see how but you've broken size/count for the other type of screens:

```
from yfinance import EquityQuery as EQ
import yfinance as yf
q = EQ('and', [
    EQ('is-in', ['exchange']+['ASX', 'NZE']), EQ('gt', ['cashfromoperations.lasttwelvemonths', 1.7*50e6]),
])
r = yf.screen(q, size=250, sortField='avgdailyvol3m', sortAsc=False)
quotes = r['quotes']
print(len(quotes))
```

Before, prints 250. After, prints 25.

R5dan:
It seems like they only changed the parameter for predefined queries, weird. Fixed now.

PR #2430, opened 2025-05-01
Incorporate curl_cffi to avoid rate limiting and cookie errors

bretsky:
~**Don't merge this!**~

~This is just a temporary workaround to allow yfinance to work if using an impersonated session from curl_cffi. Probably is better to actually incorporate curl_cffi into yfinance itself rather than require the session to be passed in to all requests, but this works for now.~

Made the change myself to use curl_cffi by default. Tested by downloading some data and appears to work locally.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,4 +10,4 @@ peewee>=3.16.2
 requests_cache>=1.0
 requests_ratelimiter>=0.3.1
 scipy>=1.6.3
-curl_cffi>=0.10.0
+curl_cffi>=0.7
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ peewee>=3.16.2
 requests_cache>=1.0
 requests_ratelimiter>=0.3.1
 scipy>=1.6.3
+curl_cffi>=0.10.0
--- a/meta.yaml
+++ b/meta.yaml
@@ -26,6 +26,7 @@ requirements:
     - frozendict >=2.3.4
     - beautifulsoup4 >=4.11.1
     - html5lib >=1.1
+    - curl_cffi >=0.7
     - peewee >=3.16.2
     - pip
     - python
@@ -41,6 +42,7 @@ requirements:
     - frozendict >=2.3.4
     - beautifulsoup4 >=4.11.1
     - html5lib >=1.1
+    - curl_cffi >=0.7
     - peewee >=3.16.2
     - python
 
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ setup(
                       'requests>=2.31', 'multitasking>=0.0.7',
                       'platformdirs>=2.0.0', 'pytz>=2022.5',
                       'frozendict>=2.3.4', 'peewee>=3.16.2',
-                      'beautifulsoup4>=4.11.1'],
+                      'beautifulsoup4>=4.11.1', 'curl_cffi>=0.7'],
     extras_require={
         'nospam': ['requests_cache>=1.0', 'requests_ratelimiter>=0.3.1'],
         'repair': ['scipy>=1.6.3'],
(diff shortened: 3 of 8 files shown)


requirements.txt
```diff
@@ -9,4 +9,5 @@ beautifulsoup4>=4.11.1
 peewee>=3.16.2
 requests_cache>=1.0
 requests_ratelimiter>=0.3.1
-scipy>=1.6.3
\ No newline at end of file
+scipy>=1.6.3
+curl_cffi==0.10.0
```
ValueRaider, line 13: why only this version?

ValueRaider (changes requested):
and why all the changes to cookie?

ValueRaider:
Makes sense. Can the cookie be exported and imported? Store persistently while it's valid, like previous code did.

bretsky:
Yes I believe so, I can look into it when I get the chance. Should be possible to just follow this pattern: https://curl-cffi.readthedocs.io/en/stable/cookies.html

wy6688:
I got the error:  ImportError: cannot import name 'MemoryQueueBucket' from 'pyrate_limiter' for pyrate_limiter>=3.0
I currently use pyrate_limiter==3.7

`class ChromeSession(Session):
    def __init__(self, **kwargs):
        kwargs['impersonate'] = "chrome"
        super().__init__(**kwargs)

# --- Combined Cached Limiter Session ---
class CachedLimiterSession(CacheMixin, ChromeSession):
    def __init__(
        self,
        cache_name=str(YFINANCE_CACHE),
        backend=None,
        expire_after=None,
        allowable_codes=(200,),
        *args,
        **kwargs
    ):
        if backend is None:
            backend = SQLiteCache(db_path=cache_name)
        elif isinstance(backend, SQLiteCache) and getattr(backend, 'db_path', None) is None:
            backend.db_path = cache_name

        super().__init__(
            cache_name=cache_name,
            backend=backend,
            expire_after=expire_after,
            allowable_codes=allowable_codes,
            *args,
            **kwargs
        )

        # http_pool_size = 100
        # https_pool_size = 100
        # self.mount("http://", HTTPAdapter(pool_connections=http_pool_size, pool_maxsize=http_pool_size))
        # self.mount("https://", HTTPAdapter(pool_connections=https_pool_size, pool_maxsize=https_pool_size))
        self.headers = MYAGENT_HEADERS

    def request(self, *args, **kwargs):
        response = super().request(*args, **kwargs)
        log_msg(f"Cache hit: {getattr(response, 'from_cache', False)} for {args}")
        return response
# Session with caching but no built-in rate limiting
limited_session = CachedLimiterSession(
    bucket_class = MemoryQueueBucket,
    cache_name=str(YFINANCE_CACHE),
    expire_after=timedelta(days=3650),
    # backend=SQLiteCache("yfinance.cache", expire_after=3600),
    ignored_parameters=["sessionId", "crumb"],
    allowable_codes=(200, 404)
)
limited_session.headers = MYAGENT_HEADERS
limited_session.timeout = 120
        try:
            if limiter.try_acquire('yfinance'):
                df = yf.download(tickers, session=limited_session, threads=False, timeout=120, **kwargs)
                # if df is None or df.empty or df.isna().all().all():
                #     log_msg(f"Empty DataFrame returned for {tickers} on attempt {attempt}, retrying...")
                #     time.sleep(2 ** attempt * 10)  # Exponential backoff: 10s, 20s, 40s
            else:
                raise BucketFullException("Rate limit exceeded")
        except BucketFullException as e:
            log_msg(f"Rate limit exceeded, retrying attempt {attempt}: {e}")
        except Exception as e:
            if "Too Many Requests" in str(e):
                log_msg(f"Rate limit error on attempt {attempt}: {e}")
                time.sleep(2 ** attempt * 10)
`

ValueRaider:
Don't post in closed pull requests.

PR #2439, opened 2025-05-03
Cleanup

R5dan:


ValueRaider:
how have you decided these are safe to delete?

R5dan:
- `.travis.yml` is for the travis cli which is not used
- `MANIFEST.in` is used for saying what to distribute and there is no `data` folder / file and `LICENSE.txt` is distributed automatically
- `setup.cfg` is used fir python 2 support to generate the `wheels` however `yfinance` is now using python 2 incompatible features (e.g. f-strings) and other required packages (e.g. `curl-cffi` and `pandas`)

R5dan:
@ValueRaider so can this be merged?

PR #2443, opened 2025-05-05
Market Holidays

R5dan:
This is the implementation of #2298

@ValueRaider @dhruvan2006 when I was implementing the market holidays for `Websocket` there is a `RecursionError`. Any ideas of how to get rid of it?

@ValueRaider I think I have fixed your commits for future pr's


R5dan:
@ValueRaider @dhruvan2006 Any ideas? I think part of the problem is the requirement of metadata and `Ticker.live`

ValueRaider:
Never merge main/dev into your branch, instead rebase.

Ensure your base matches target branch.

R5dan:
Ok, but do you like the idea? Shall I make the PR again?

ValueRaider:
Oh the idea. I'm undecided, so make the PR and let community vote on it.

I'm not adding a dependency to a brand new package, especially when solutions already exist like exchange_calendars.

R5dan:
> I'm not adding a dependency to a brand new package, especially when solutions already exist like exchange_calendars.

The problem with most of the others is that they are static, `BetterHolidays` uses web scraping for the holidays that can change year to year as well as some adhoc holidays. `exchange_calenders` seems to use a weird naming conversion as well. I built `BetterHolidays` because I don't like the other options, if you could tell me why you don't want to add it so I can fix or tell me which alternative to use.

ValueRaider:
> BetterHolidays uses web scraping

Where?

> exchange_calenders seems to use a weird naming conversion

So add a single dict mapping yf -> xcals. Are you willing to reproduce every calendar here? https://help.yahoo.com/kb/SLN2310.html?_guc_consent_skip=1747426492 (if you are, you have more impact adding them to xcals)

> I built BetterHolidays because I don't like the other options

That's the problem.

R5dan:
Sorry I have been busy:

> Where?

Oh sorry, unreleased pr adding more markets

> Are you willing to reproduce every calendar here? https://help.yahoo.com/kb/SLN2310.html?_guc_consent_skip=1747426492

Yes. I think it will take less time to do it in `BetterHolidays` as I will use web scraping though.

Issue #2441, opened 2025-05-05
[0.2.58/59] Yahoo may return bad crumb which is not detected properly


vsukhoml:
### Describe bug

From time to time I'm getting:
`requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://query1.finance.yahoo.com/v7/finance/quote?symbols=INGR&formatted=false&crumb=Too+Many+Requests%0D%0A`

Note `crumb=Too+Many+Requests%0D%0A` - this specific value shall be treated as no crumb is available.

### Simple code that reproduces your problem

Any yfinance query. 

### Debug log from yf.enable_debug_mode()
```
DEBUG    get_raw_json(): https://query2.finance.yahoo.com/v10/finance/quoteSummary/GOOG
DEBUG:yfinance:get_raw_json(): https://query2.finance.yahoo.com/v10/finance/quoteSummary/GOOG
DEBUG    Entering get()
DEBUG:yfinance:Entering get()
DEBUG     Entering _make_request()
DEBUG:yfinance: Entering _make_request()
DEBUG      url=https://query2.finance.yahoo.com/v10/finance/quoteSummary/GOOG
DEBUG:yfinance:  url=https://query2.finance.yahoo.com/v10/finance/quoteSummary/GOOG
DEBUG      params={'modules': 'financialData,quoteType,defaultKeyStatistics,assetProfile,summaryDetail', 'corsDomain': 'finance.yahoo.com', 'formatted': 'false', 'symbol': 'GOOG'}
DEBUG:yfinance:  params={'modules': 'financialData,quoteType,defaultKeyStatistics,assetProfile,summaryDetail', 'corsDomain': 'finance.yahoo.com', 'formatted': 'false', 'symbol': 'GOOG'}
DEBUG      Entering _get_cookie_and_crumb()
DEBUG:yfinance:  Entering _get_cookie_and_crumb()
DEBUG       cookie_mode = 'basic'
DEBUG:yfinance:   cookie_mode = 'basic'
DEBUG       Entering _get_cookie_and_crumb_basic()
DEBUG:yfinance:   Entering _get_cookie_and_crumb_basic()
DEBUG        crumb = 'Too Many Requests
             '
DEBUG:yfinance:    crumb = 'Too Many Requests
    '
DEBUG       Exiting _get_cookie_and_crumb_basic()
DEBUG:yfinance:   Exiting _get_cookie_and_crumb_basic()
[...]

PR #2448, opened 2025-05-06
Handle crumb='Too Many Requests'

vsukhoml:
Fix for #2441


ValueRaider:
I've just tested but still get rate error. Possible Yahoo changed again since.

ValueRaider:
So the real problem is yfinance still setting user-agent - it shouldn't be.

vsukhoml:
I think Yahoo started to reject all requests coming from standard `requests` package, so you need to use `curl_cffi.requests.Session`  with `impersonate=chrome` (what yfinance used by default is session is not provided). I wasn't able at first to connect it with request_cache and request_limiter, so replaced cache with https://github.com/ValueRaider/yfinance-cache and manually added throttling to avoid rate limits. It had been working several hours ago for me. 

In another thread I found proper way:
```
from curl_cffi import Session as CurlSession

class ChromeSession(CurlSession):
    def __init__(self, **kwargs):
        kwargs['impersonate'] = "chrome"
        super().__init__(**kwargs)


logger = logging.getLogger(__name__)

# Create a session class that combines caching and rate limiting
class CachedLimiterSession(CacheMixin, LimiterMixin, ChromeSession):
    pass

# Create a session with rate limiting and caching
session = CachedLimiterSession(
    limiter=Limiter(RequestRate(5, Duration.SECOND)),
    bucket_class=MemoryQueueBucket,
    backend=SQLiteCache("yfinance.cache", expire_after=3600),
    ignored_parameters=["sessionId", "crumb"]
)
```

ValueRaider:
> > So the real problem is yfinance still setting user-agent - it shouldn't be.
> 
> I looked what `finance.yahoo.com` sets in headers - it does set `User-Agent`. Ah, may be I also changed `user-agent` to `User-Agent` or vice versa.

Sorry I didn't mean you caused that, it was already present.

ValueRaider:
Your `CachedLimiterSession` example does not work at all. Nothing is cached, and the rate-limit doesn't work.

vsukhoml:
> Your CachedLimiterSession example does not work at all. Nothing is cached, and the rate-limit doesn't work. Which is what I said in other thread.

Thanks! I didn't try it, but guess you are correct - I couldn't make it work properly and it didn't look like `impersonate="chrome" ` worked as I immediately got 429 errors.

>Sorry I didn't mean you caused that, it was already present.
I meant I changed it locally during experiments which may explain why it works for me. I found that I had to specifically use curl_cffi's Session with impersonate.

vsukhoml:
I rebased to resolve conflict. It looks like the major reason I got that bad crumb was related to using custom `Session` from requests package which was rejected. When I switched to `curl_cffi`'s Session issue disappeared, however since we now learned this behavior of Yahoo server, I think it makes sense to handle it gracefully rather than trying send and possibly cache bad crumb.

ValueRaider:
I think better fix is, detect "Too Many Requests" where ever it happens and abort. Respect what the message is saying.

vsukhoml:
What is your proposed change to the behavior?  Change it to raise an exception? I don't want bad crumb to be "sticky" when cached. In my cases I got "Too many Requests\n\r" and no status=429. But Yahoo may change behavior. 429 errors are already handled.

ValueRaider:
Raise it. YF hides most errors, but killer errors like rate-limit should be raised https://github.com/ranaroussi/yfinance/blob/4c23d339bc88ce4b70ba0dd4add7e38743242833/yfinance/scrapers/history.py#L190-L191

PR #2455, opened 2025-05-07
Add automated formatting /  fixing of linting ( using Ruff / actions)

cole-st-john:
### Problem:
I am interested to contribute significantly to this project, as I am using the repository.   The fact that the project hasnt currently leveraged automated formatting (consistent line endings, quotations, visualization of lists, etc) , complicates creating clean commits for those of us currently working that way. 

### Proposed Solution:
- I propose adding at least a ruff.toml so that all users using ruff automated formatting will match the project standard.  
- Adjusting the current ruff action to include formatting of all pushes can be nice to reduce work on contributors / ensuring standardization.   
- Includes the action update of [#2402 ](https://github.com/ranaroussi/yfinance/pull/2402) to Astral-Shs up-to-date Ruff action.

I wanted to feel out if this type change might be desired in this project.  Any feedback or thoughts are appreciated.  

Cole


ValueRaider:
I might be minority but I think linting enforcement can go too far. I deliberately break a lot of ruff rules in `history.py` e.g. `if div_pct*100 < 0.1:` vs `if div_pct * 100 < 0.1:`

Also every change has to be verified not malicious, so that will delay merge.

cole-st-john:
> I might be minority but I think linting enforcement can go too far. I deliberately break a lot of ruff rules in `history.py` e.g. `if div_pct*100 < 0.1:` vs `if div_pct * 100 < 0.1:`
> 
> Also every change has to be verified not malicious, so that will delay merge. Do all strings have to be double-quoted?

We can just as easily make it where all quotes are single.  Consistency is nice.

ValueRaider:
You need to disable your linting for this project, or at least apply it much more gradually. This isn't like a company where you can trust your colleague and just click "Merge" - every character change has to reviewed.

Issue #2480, opened 2025-05-10
[0.2.59] YFRateLimitError: Too Many Requests. Rate limited.


lixin4sky:
### Describe bug

YFRateLimitError: Too Many Requests. Rate limited. Try after a while.

This bug is still here, 0.2.59/0.2.58/0.2.57.

### Simple code that reproduces your problem

import yfinance as yf

stock = yf.Ticker('SPY')
stock.info

### Debug log from yf.enable_debug_mode()

DEBUG    get_raw_json(): https://query2.finance.yahoo.com/v10/finance/quoteSummary/SPY
DEBUG    Entering get()
DEBUG     Entering _make_request()
DEBUG      url=https://query2.finance.yahoo.com/v10/finance/quoteSummary/SPY
DEBUG      params={'modules': 'financialData,quoteType,defaultKeyStatistics,assetProfile,summaryDetail', 'corsDomain': 'finance.yahoo.com', 'formatted': 'false', 'symbol': 'SPY'}
DEBUG      Entering _get_cookie_and_crumb()
DEBUG       cookie_mode = 'basic'
DEBUG       Entering _get_cookie_and_crumb_basic()
DEBUG        crumb = '<!DOCTYPE html>
             <html lang="zh">
    
             </body></html>
             '
DEBUG       Exiting _get_cookie_and_crumb_basic()
DEBUG      Exiting _get_cookie_and_crumb()
DEBUG      response code=429
DEBUG      toggling cookie strategy basic -> csrf
DEBUG      Entering _get_cookie_and_crumb()
DEBUG       cookie_mode = 'csrf'
DEBUG       Entering _get_crumb_csrf()
DEBUG        Failed to find "csrfToken" in response
DEBUG       Exiting _get_crumb_csrf()
DEBUG       toggling cookie strategy csrf -> basic
DEBUG       Entering _get_cookie_and_crumb_basic()
DEBUG        crumb = '<!DOCTYPE html>
             <html lang="zh">

             </body></html>
             '
DEBUG       Exiting _get_cookie_and_crumb_basic()
DEBUG      Exiting _get_cookie_and_crumb()
DEBUG      response code=429

### Bad data proof

<img width="929" alt="Image" src="https://github.com/user-attachments/assets/abe08537-46bf-4cdd-b813-a07a0b9adae0" />

### `yfinance` version

0.2.59

### Python version

3.10.16/3.13.3

### Operating system

OSX

xgiannak:
[...]

PR #2491, opened 2025-05-11
Fix for rate limit during getting crumb.

vsukhoml:
Address #2441, #2480


yfinance/data.py
```diff
@@ -223,6 +223,10 @@ def _get_crumb_basic(self, timeout=30):
         else:
             crumb_response = self._session.get(**get_args)
         self._crumb = crumb_response.text
+        if crumb_response.status_code == 429 or "Too Many Requests" in self._crumb:
```
ValueRaider, line 226: Does this change on its own solve the problem? Could revert the other changes.

vsukhoml: yes, this is the core change, and need to be in both places (basic and csrf). However, if bad crumb got cached somehow, it would make sense to prevent it from being used.

As for "fix spaces" - it seems there are trailing spaces in the source file so I need to change editor settings specifically to preserve those. Not sure how valuable are they for the preservation.

ValueRaider: How would it get cached somehow? It's deliberately never stored persistently.
Obviously keep the similar guard in `_get_crumb_csrf`, then no way to store a "429 crumb".

vsukhoml: if it is cached before yfinance update (e.g. requests-cache used or yfinance cache), and then used after update. But looks like it may not be needed anymore due to other changes and removal of requests-cache support.

ValueRaider: The `if self._session_is_caching` just before should have prevented that. Were you seeing it being reused?

vsukhoml: Yes, with .58 and yfinance cache(`set_tz_cache_location`) and requests-cache active I got into state when there was no attempt to update a crumb because crumb was always "Too Many Requests". I never had it with .59 version and now I'm not even sure if it was because I used Session from requests vs. curl_cffi. I removed this check as I don't think it will happen again due to proper error handling at getting crumb in the first place.

ValueRaider:
Finally just git squash to tidy the commit history https://ranaroussi.github.io/yfinance/development/code.html

ValueRaider:
> Merge branch 'ranaroussi:main' into crumb

Revert that, there's a better way https://ranaroussi.github.io/yfinance/development/code.html

Issue #2451, opened 2025-05-06
period="max" fails with multiple intervals


cole-st-john:
#### Should: 
Intuitively, when downloading price history, i am expecting using period="max" should return all available data for any interval.  

#### Is: 
Mode1 - Bug (what I want to focus on here): Currently, yfinance fails on a number of intervals - returning an empty dataframe.
Mode2 - Feature: I would expect for 1m, requests would be chunked so that the entirety of the available (30 day period) data would be requested in chunks and unioned into a single result.

### Simple code that reproduces your problem

```python
import yfinance as yf

INTRADAY_INTERVALS = ["1m", "2m", "5m", "15m", "30m", "60m", "1h", "90m"]
DAY_PLUS_INTERVALS = ["1d", "5d", "1wk", "1mo", "3mo"]
INTERVALS = [*INTRADAY_INTERVALS, *DAY_PLUS_INTERVALS]
TICKERS = ["EURUSD=X"]

for interval in INTERVALS:
    df = yf.download(tickers=TICKERS, period="max", interval=interval, progress=False)
    print(f"failed for {interval}") if df.empty else print(f"passed for {interval}")
```

### Debug log from yf.enable_debug_mode()

```sh
DEBUG    Entering download()
YF.download() has changed argument auto_adjust default to True
DEBUG     Disabling multithreading because DEBUG logging enabled
DEBUG     Using User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36
DEBUG     Entering history()
DEBUG      Entering history()
DEBUG       EURUSD=X: Yahoo GET parameters: {'period1': '2025-04-29 13:11:21+01:00', 'period2': '2025-05-06 13:11:21+01:00', 'interval': '1m', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG       Entering get()
DEBUG        Entering _make_request()
DEBUG         url=https://query2.finance.yahoo.com/v8/finance/chart/EURUSD=X
DEBUG         params={'period1': 1745928681, 'period2': 1746533481, 'interval': '1m', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG         Entering _get_cookie_and_crumb()
DEBUG          cookie_mode = 'basic'
DEBUG          Entering _get_cookie_and_crumb_basic()
[...]

PR #2508, opened 2025-05-19
adjusting for processing time in max period

cole-st-john:
This PR is to address issue #2451 - around yf.download with period="max".

- There was a bug that interval="2m" was not accounted for.
- Intraday intervals often fail due to processing time - which leads to requesting a period of data that exceeds the maximum period.  In reducing a buffer from the calculated end and start dates, the error is avoided.
- Also, the 1m interval chunk was increased  to 8 days as this is the actual max requestable period/chunk.

Cheers,
Cole


ValueRaider:
That's a lot of changes...

PR #2513, opened 2025-05-24
Improve exception handling

### LINKED ISSUE #2213: Can we just raise HTTPErrors instead of logging them?
### Describe bug

The `_fetch` method on `Quote` in `scrapers.quote` checks a few things and raises error. Just `requests.exceptions.HTTPError` is not raised. Instead it is caught and an error is logged (see: https://github.com/ranaroussi/yfinance/blob/5adddf3e88f39720ef24a04927738848dee76761/yfinance/scrapers/quote.py#L589)

```python
        try:
            result = self._data.get_raw_json(_QUOTE_SUMMARY_URL_ + f"/{self._symbol}", user_agent_headers=self._data.user_agent_headers, params=params_dict, proxy=proxy)
        except requests.exceptions.HTTPError as e:
            utils.get_yf_logger().error(str(e))
            return None
        return result
```

I would like to catch this error and handle it when using `yfinance`. Why not do something like:

```python
        try:
            result = self._data.get_raw_json(_QUOTE_SUMMARY_URL_ + f"/{self._symbol}", user_agent_headers=self._data.user_agent_headers, params=params_dict, proxy=proxy)
        except requests.exceptions.HTTPError as e:
            raise YFException(str(e))
        return result
```

(Posted it as "Bug" because it seems like an exception should be handled or raised)

```python
try:
    Ticker("BIG").info()
except Exception as err:
    error = err
print(error)
```

### Debug log

```
404 Client Error: Not Found for url: https://query2.finance.yahoo.com/v10/finance/quoteSummary/BIG?modules=financialData%2CquoteType%2CdefaultKeyStatistics%2CassetProfile%2CsummaryDetail&corsDomain=finance.yahoo.com&formatted=false&symbol=BIG&crumb=92tlCzJ1hkE
>>> error
NameError: name 'error' is not defined. Did you mean: 'OSError'?
```

### Bad data proof

_No response_

### `yfinance` version

0.2.51

### Python version

_No response_

### Operating system

_No response_

ValueRaider:
Begin task of stop hiding exceptions if `yf.set_config(hide_exceptions=False)`. Because hiding exceptions makes bugs difficult to manage.

But also improve handling exceptions, by replacing with more meaningful. E.g. replace 404 with `YFTickerMissingError ... possible delisted`

Fix #2213


yfinance/base.py
```diff
@@ -600,13 +614,17 @@ def get_shares_full(self, start=None, end=None, proxy=_SENTINEL_):
...
+                raise
             return None
         try:
             fail = json_data["finance"]["error"]["code"] == "Bad Request"
         except KeyError:
             fail = False
         if fail:
             logger.error(f"{self.ticker}: Yahoo web request for share count failed")
+            if not YfConfig().hide_exceptions:
+                raise requests.exceptions.HTTPError("Yahoo web request for share count returned 'Bad Request'")
```
R5dan, line 627: Might be worth making a finance error for this. Some users might use `except yfexception` and this could seem like an internal error.

yfinance/config.py
```diff
@@ -0,0 +1,33 @@
...
+    def __call__(cls, *args, **kwargs):
+        with cls._lock:
+            if cls not in cls._instances:
+                instance = super().__call__(*args, **kwargs)
+                cls._instances[cls] = instance
+            else:
+                # Update the existing instance
+                if 'hide_exceptions' in kwargs or (args and len(args) > 0):
+                    hide_exceptions = kwargs.get('hide_exceptions') if 'hide_exceptions' in kwargs else args[0]
+                    cls._instances[cls]._set_hide_exceptions(hide_exceptions)
```
R5dan, line 19: Why not just add this on the base class?
Couldn't this be done in the `__init__`?

yfinance/config.py
```diff
@@ -0,0 +1,33 @@
...
+class YfConfig(metaclass=SingletonMeta):
+    def __init__(self, hide_exceptions=True):
+        self._hide_exceptions = hide_exceptions
+
+    def _set_hide_exceptions(self, hide_exceptions):
+        self._hide_exceptions = hide_exceptions
+
+    @property
+    def hide_exceptions(self):
+        return self._hide_exceptions
```
R5dan, line 32: Might be worth changing this / adding `raise_exceptions` as most uses are negating this

yfinance/scrapers/history.py
```diff
@@ -80,13 +81,17 @@ def history(self, period="1mo", interval="1d",
             utils.print_once("YF deprecation warning: set proxy via new config function: yf.set_config(proxy=proxy)")
```
R5dan, line 81: Better to use actual deprecation error. Can still only be done once though.

yfinance/scrapers/history.py
```diff
@@ -95,7 +100,7 @@ def history(self, period="1mo", interval="1d",
                     err_msg = str(_exception)
                     shared._DFS[self.ticker] = utils.empty_df()
                     shared._ERRORS[self.ticker] = err_msg.split(': ', 1)[1]
-                    if raise_errors:
+                    if raise_errors or (not YfConfig().hide_exceptions):
```
R5dan, line 103: Set `raise_errors = raise_errors or not YfConfig().hide_exceptions`

yfinance/scrapers/history.py
```diff
@@ -430,12 +437,12 @@ def history(self, period="1mo", interval="1d",
                 err_msg = "auto_adjust failed with %s" % e
             else:
                 err_msg = "back_adjust failed with %s" % e
```
R5dan, line 439: Why not use f strings and still include ticket name

yfinance/scrapers/history.py
```diff
@@ -1885,7 +1894,7 @@ def cluster_dividends(df, column='div', threshold=7):
                     elif k == 'div_true_date':
                         div_status_df[k] = pd.Series(dtype='datetime64[ns, UTC]')
                     else:
-                        raise Exception(k,v,type(v))
+                        raise ValueError(k,v,type(v))
```
R5dan, line 1897: Why not make the error more readable (add an actual message)

yfinance/scrapers/holders.py
```diff
@@ -96,8 +99,10 @@ def _fetch_and_parse(self):
             self._parse_insider_transactions(data.get("insiderTransactions", {}))
             self._parse_insider_holders(data.get("insiderHolders", {}))
             self._parse_net_share_purchase_activity(data.get("netSharePurchaseActivity", {}))
-        except (KeyError, IndexError):
-            raise YFDataException("Failed to parse holders json data.")
+        except (KeyError, IndexError) as e:
+            if not YfConfig().hide_exceptions:
+                raise
+            raise YFDataException(f"# RESPONSE:\n{result}\n\nFailed to parse 'holders' json response from Yahoo Finance (above): {str(e)}")
```
R5dan, line 105: You probably want to raise a yf error either way, and is it supposed to raise `YfDataException` either way? Seems counter intuative

yfinance/scrapers/quote.py
```diff
@@ -522,8 +521,8 @@ def sustainability(self) -> pd.DataFrame:
             else:
                 try:
                     data = result["quoteSummary"]["result"][0]
-                except (KeyError, IndexError):
-                    raise YFDataException(f"Failed to parse json response from Yahoo Finance: {result}")
+                except (KeyError, IndexError) as e:
+                    raise YFDataException(f"# RESPONSE:\n{result}\n\nFailed to parse json response from Yahoo Finance (above): {str(e)}")
```
R5dan, line 525: Should this always raise? What about if `hide_exceptions` is true?

yfinance/scrapers/quote.py
```diff
@@ -536,8 +535,8 @@ def recommendations(self) -> pd.DataFrame:
             else:
                 try:
                     data = result["quoteSummary"]["result"][0]["recommendationTrend"]["trend"]
-                except (KeyError, IndexError):
-                    raise YFDataException(f"Failed to parse json response from Yahoo Finance: {result}")
+                except (KeyError, IndexError) as e:
+                    raise YFDataException(f"# RESPONSE:\n{result}\n\nFailed to parse json response from Yahoo Finance (above): {str(e)}")
```
R5dan, line 539: Same here?

R5dan (changes requested):

yfinance/config.py
ValueRaider: Do you know what a singleton is?

yfinance/scrapers/history.py
ValueRaider: What do you mean "actual error"? An exception? That would break users codes for no good reason.

yfinance/config.py
R5dan: Yes, where the class is limited to one instance. I am saying that I think the setting of the `hide_exceptions` should be done in the `__init__` or in a class method. I don't think it should happen here as it adds unnecessary complexity and makes it harder to set than it should be.

yfinance/config.py
R5dan: It could be a private property then, or a signal. I just think it would reduce complexity in the code base, but understand your point about the user.

yfinance/config.py
ValueRaider: This is how to set `hide_exceptions` after YfData created. It's necessary.

Also `SingletonMeta` is the base class.

ValueRaider:
squash broke history.

PR #2516, opened 2025-05-25
Feature: ISIN cache

ValueRaider:
Cache ISIN -> ticker symbol, avoid repeating lookup.

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -69,9 +69,14 @@ class TickerBase:
         # accept isin as ticker
         if utils.is_isin(self.ticker):
             isin = self.ticker
-            self.ticker = utils.get_ticker_by_isin(self.ticker, _SENTINEL_, session)
+            c = cache.get_isin_cache()
+            self.ticker = c.lookup(isin)
+            if not self.ticker:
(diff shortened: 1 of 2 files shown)


ValueRaider:
@szotsaki I don't use ISINs so value your input. Does pruning old ISINs make sense? E.g. a company $ticker acquires a new ISIN, so I prune old-ISIN -> $ticker.

szotsaki:
Yes, I think so. There are circumstances when ISIN of the security changes. I don't know whether the ticker also changes with it, though.

From Wikipedia:
> ISINs consist of two alphabetic characters, which are the [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) code for the issuing country, nine alpha-numeric characters (the [National Securities Identifying Number](https://en.wikipedia.org/wiki/National_Securities_Identifying_Number), or NSIN, which identifies the security, padded as necessary with leading zeros), and one numerical [check digit](https://en.wikipedia.org/wiki/Check_digit). 
> **When the NSIN changes due to corporate actions or other reasons, the ISIN will also change.**

PR #2543, opened 2025-06-20
Raise exception instead of only logging and returning None

Adoliin:
Currently, we're logging the error with `logger.error` and returning `None`, which makes it harder to handle errors downstream. I believe it would be more effective to raise the exception instead. This way, I can properly catch and handle it in the calling function, leading to more robust error handling overall.


ValueRaider:
redundant in version 1.0

PR #2544, opened 2025-06-20
Fix SSL certificate path issue with non-ASCII characters on Windows

bagcivan:
- Add _get_safe_session() function to handle SSL cert paths with non-ASCII chars
- Detect non-ASCII characters in certificate path and create temp cert when needed
- Add comprehensive tests for SSL certificate path handling
- Maintain backward compatibility - only activates workaround when necessary
- Fix for international users (Turkish, Chinese, Arabic, etc.) on Windows

❌ curl: (77) error setting certificate verify locations

### DIFF
--- a/tests/test_ssl_cert_path.py
+++ b/tests/test_ssl_cert_path.py
@@ -1,0 +1,116 @@
+"""
+Test SSL certificate path handling for non-ASCII characters.
+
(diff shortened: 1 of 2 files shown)


ValueRaider:
Looks like you're fixing a big in curl_cffi, then you should be sending a fix to that project not yfinance.

Issue #702, opened 2021-05-05
Add function to map MIC to yahoo suffix


mxdev88:
In most cases, Yahoo Finance builds its symbol from the ticker symbol + a [yahoo market suffix](https://help.yahoo.com/kb/finance-for-web/SLN2310.html;_ylt=AwrJKiCZFo9g3Y8AsDWPAwx.;_ylu=Y29sbwMEcG9zAzEEdnRpZAMEc2VjA3Ny?locale=en_US)

It would be quite useful if the ```yfinance``` had a built in way to map [MIC code](https://www.iso20022.org/market-identifier-codes) to Yahoo suffixes.

For example,
```py
suffix = yf.market_suffix(mic='XPAR')
```

This would allow to 'easily' build many of the yahoo symbols from systems that use MIC as market identifiers.

e.g. 'XPAR' => '.PA', 'XNYM' => '.NYM',  'XCBT' => '.CBT', and so on...

@ranaroussi, if that makes sense I can contribute a PR.

ValueRaider:
As only you have requested this, best if you implement outside `yfinance` to minimise code bloat.

PythonPeterr:
@mxdev88  did u find a fix for this?

mxdev88:
> [@mxdev88](https://github.com/mxdev88) did u find a fix for this?

I'm mapping myself. Glad to see i'm not the only one anymore 🙂

PythonPeterr:
Can u share it with me pls :)?

ValueRaider:
@mxdev88 Can you publish code on Github so we can evaluate adding to yfinance?

mxdev88:
> [@mxdev88](https://github.com/mxdev88) Can you publish code on Github so we can evaluate adding to yfinance?

@ValueRaider @PythonPeterr created a [PR](https://github.com/ranaroussi/yfinance/pull/2579). It's really only a mapping between official MIC code and Yahoo suffix.

PR #2579, opened 2025-08-03
feat: add market_suffix

mxdev88:
This PR implements https://github.com/ranaroussi/yfinance/issues/702

```py
import yfinance as yf

for mic in ['XNAS', 'XPAR', 'XLON', 'XBRU', 'XOSL']:
    print(f"Yahoo {mic} market suffix: {yf.market_suffix(mic)}")
```


ValueRaider:
After you map a MIC to yahoo suffix, how do you use it?

Maybe just expose dictionary `MIC_TO_YAHOO`?

PythonPeterr:
Did u check if all these MICs are also the supported ones from YFinance? As in; what happens whenever a MIC is not in this hardcoded list? Does it default back to no yahoo finance suffix?

mxdev88:
> After you map a MIC to yahoo suffix, how do you use it?
> 
> Maybe just expose dictionary `MIC_TO_YAHOO`?

I prefer wrapping in a function `market_suffix` as `_MIC_TO_YAHOO_SUFFIX` should be immutable.

The example usage is to dynamically generate the yahoo ticker when a user has a list of tuple exchange symbol and MIC.

```py
import yfinance as yf

# Example usage
equities = [
    ("OR", "XPAR"),      # L'Oréal on Euronext Paris
    ("AAPL", "XNYS"),    # Apple on NYSE
    ("GOOGL", "XNAS"),   # Alphabet on NASDAQ
    ("BMW", "XETR"),     # BMW on XETRA
]

def yahoo_ticker(symbol: str, mic: str) -> str:
    return f"{symbol}{yf.market_suffix(mic)}"

for symbol, mic in equities:
    ticker = yahoo_ticker(symbol, mic)
    print(f"{symbol} on {mic} -> {ticker}")
```

which gives you:

```sh
OR on XPAR → OR.PA
AAPL on XNYS → AAPL
GOOGL on XNAS → GOOGL
BMW on XETR → BMW.DE
```

So it could also be useful to expose `yahoo_ticker` function as well.

mxdev88:
> Did u check if all these MICs are also the supported ones from YFinance? As in; what happens whenever a MIC is not in this hardcoded list? Does it default back to no yahoo finance suffix?

I've tested on major ones I use. I did not test thoroughly **ALL** MIC's though but you get the idea. And it'd be great if you could test and provide feedback as well.

If not available, the current code defaults back to no suffix - it would probably be best to raise ValueError : "MIC _code_ is not supported."

PythonPeterr:
@mxdev88 where did u get the suffixes from? As in where did u find a comprehensive list?

mxdev88:
> @mxdev88 where did u get the suffixes from? As in where did u find a comprehensive list?

@PythonPeterr i've posted the link in the original post + it's included in the header of the file if you take a look at the diff.

PythonPeterr:
> > @mxdev88 where did u get the suffixes from? As in where did u find a comprehensive list?
> 
> @PythonPeterr i've posted the link in the original post + it's included in the header of the file if you take a look at the diff.

u legend

ValueRaider:
Does this syntax work better?
> yf.Ticker(("OR", "XPAR"))

yfinance can detect a tuple and apply MIC map. Similar idea to auto-translating ISIN

mxdev88:
> Does this syntax work better?
> 
> > yf.Ticker(("OR", "XPAR"))
> 
> yfinance can detect a tuple and apply MIC map. Similar idea to auto-translating ISIN

Sure, that would work too. are you suggesting it should be replacing the function entirely or an additional way to construct the yahoo ticker ? In any case,  it should probably be documented in Ticker __init__ docstring. tbh, i was not aware Ticker was accepting an ISIN as argument.

ValueRaider:
@mxdev additional, it's just an if-else

mxdev88:
@ValueRaider i've updated the PR with your suggestion.

ValueRaider:
Can the `__init__.py` changes be reverted?

mxdev88:
> Can the `__init__.py` changes be reverted?

reverted. I take it you don't want `market_suffix` or `yahoo_ticker` to be public?

ValueRaider:
Revert anything not needed. There should be reason for any code change.

mxdev88:
> Revert anything not needed. There should be reason for any code change.

@ValueRaider what do you mean? Everything left in the PR is needed. Can you be more specific?

ValueRaider:
E.g. What's the reason for `market_suffix` being public?

mxdev88:
It's not public. It's in the mic module so user can only import it by speciffically importing that module. And it is tested unitarly.

ValueRaider:
My bad, I misinterpreted your previous message.

I'll leave this open a little bit to give chance for others to discuss.

ValueRaider:
I've simplified the PR, seemed over-engineered.

mxdev88:
Looks like i can no longer use the functions independandly anymore. Not convinnce how that makes it better but i guess you have the authority on the code base qo leave it up to you.

ValueRaider:
Your functions were just wrapping a dictionary, they provided nothing extra. You can use the dictionary.

salsugair:
minor feed back on mic.py
    # Saudi Arabia
    'XTAD': '.SAU',
mic code for Saudi Arabia is XSAU and yahoo finance append .SR at the end example 2222.SR
I have stocks in 5 exchanges 4 are correct only this one wasn't accurate, when I open yahoo finance and use search bar 2222 it shows Equity SAU but it uses 2222.SR as symbol

ValueRaider:
Good spot, XTAD is Taiwan. https://ranaroussi.github.io/yfinance/development/index.html

PR #2580, opened 2025-08-03
Fix test_ticker unit test failure due to NEPT delisting

skyblue6688:
This PR fixes the bug in: https://github.com/ranaroussi/yfinance/issues/2576

The unit test now pass:

```
python3 -m unittest tests.test_ticker
```

```
Ran 64 tests in 25.539s

OK
```


ValueRaider:
If NEPT delisted since test made, then remove NEPT from tests.

skyblue6688:
@ValueRaider Sounds good, I've removed NEPT from the test

PR #2585, opened 2025-08-20
Update index.rst - duplicate market reference

roberttidball:
The market reference in the documentation (https://ranaroussi.github.io/yfinance/reference/api/yfinance.Market.html#) is duplicated.


ValueRaider:
Good spot. It's also duplicated in left pane, can you see why?

roberttidball:
Yes, just updated to remove the duplicate in the left pane, the issue was in the table of contents tree, thanks!

PR #2589, opened 2025-08-31
Fix earnings_dates

hjlgood:
TLDR;
pip uninstall yfinance
pip install git+https://github.com/hjlgood/yfinance.git
use function `Ticker.get_earnings_dates_using_scrape()`

---

Regarding https://github.com/ranaroussi/yfinance/issues/2566,

The current API does not fetch the most recent earnings data.
https://github.com/ranaroussi/yfinance/issues/2566#issuecomment-3111510412

This is a huge deal breaker that hasn't been fixed for quite a while.

As a work-around, I have implemented a function "get_earnings_dates_using_selenium" to scrap from the following link :
https://finance.yahoo.com/calendar/earnings?symbol=ERIE
(symbol is subject to change)

"get_earnings_dates_using_selenium" works pretty much the same as "get_earnings_dates"
but it requires "selenium" and "webdriver_manager" package in requirements.txt.

I understand that using selenium is not favorable because it's slow and it also brings more dependencies, 
but I think it is better than not fetching recent earnings data.

please take a look!

Issue #2566, opened 2025-07-06
get_earnings_dates API cannot return the future earnings_dates?


jason-zhang-hc:
### Describe bug

When I use **get_earnings_dates** api around several months ago, it can return the **future earnings dates as** below:

<img width="558" height="246" alt="Image" src="https://github.com/user-attachments/assets/0370c6a6-2777-447f-8545-43efe7d2653d" />

However recently when I use  this api again, it can only return the **historical earnings dates** as shown here:

<img width="690" height="288" alt="Image" src="https://github.com/user-attachments/assets/b627ecc0-b729-47b0-9864-deaf463e6214" />

Does anyone know how to solve this issue? I need to get the future earnings dates to monitor the stocks I wanna trade. Thanks for your help!

### Simple code that reproduces your problem

import yfinance as yf
from curl_cffi import requests

ticker = 'AAPL'
requests.Session(impersonate="chrome")

stock = yf.Ticker(ticker, session=session)
earnings_dates = stock.get_earnings_dates(20)
print(earnings_dates)

### Debug log from yf.enable_debug_mode()

no

### `yfinance` version

0.2.64

Alex-Naxitus:
I raised this issue a couple of weeks ago :  #2552 
The problem comes from Yahoo Finance's database. However, you can still use ticker.get_calendar() as a workaround, as that still has future earnings dates.
Edit : But it will be less precise, because get_calendar gives only the date, not the time of the future earnings.

jason-zhang-hc:
> I raised this issue a couple of weeks ago : [#2552](https://github.com/ranaroussi/yfinance/issues/2552) The problem comes from Yahoo Finance's database. However, you can still use ticker.get_calendar() as a workaround, as that still has future earnings dates. Edit : But it will be less precise, because get_calendar gives only the date, not the time of the future earnings.

Thanks Alex. I will follow your workaround first :)

Ohio168:
Is there a timeline for resolving the issue? The reporting season is approaching.

Alex-Naxitus:
> Is there a timeline for resolving the issue? The reporting season is approaching.

As this problem comes from Yahoo Finance, I doubt they'll fix the issue. I contacted them months ago, complaining that their earnings dates were inaccurate, they claimed they looked into it and didn't find a problem.

You can look online to search for the next day's earnings, and modify your code to add your earnings dates manually like I'm doing. Or you can subscribe to paying API service , like Finhub, which is what I plan to do eventually.

[...]

PR #2590, opened 2025-09-05
Fix get earnings dates

hjlgood:
Regarding 
https://github.com/ranaroussi/yfinance/issues/2566

I added get_earnings_dates_using_scrape
https://github.com/ranaroussi/yfinance/pull/2589

This is a new PR with multiple commits squashed into one.
I'm making a new PR to base:dev instead of base:main.


ValueRaider:
The rebase is bad, it should only contain commits made by you. You have to start over from scratch.

PR #2599, opened 2025-09-16
fix: add missing logging message

Phil997:
Add the missing text from the response when logging the error of the request.


ValueRaider:
Rename title because real problem is curl_cffi's `print(e)` is hiding the error.

PR #2609, opened 2025-10-05
Added more User-Agents to prevent error 429

ariasj07:


ValueRaider:
It's not used anymore.

Issue #2593, opened 2025-09-10
Wrong price to book value is returned for the Indian companies


InfantJerin:
### Describe bug

priceToBook value for Infosys Ticker INFY.NS ( https://finance.yahoo.com/quote/INFY.NS/ ) is returned by the API is given below. it is 573 but that is wrong, the web app shows the right values but the API returns different value.

"bookValue": 2.67,
"priceToBook": 573.55804,
"currentPrice": 1531.4,
"financialCurrency": "USD",
"currency": "INR",

Below is the observation done by chatgpt - 

🧐 Observations

bookValue = 2.67

This should normally mean Book Value per Share (BVPS).

But clearly 2.67 is not in INR, because Infosys BVPS (from filings) is around ₹230.

It looks like yfinance is mixing USD values for fundamentals while showing INR prices.

currentPrice = 1531.4 (INR)

This is the live NSE share price in rupees.

priceToBook = 573.55804

If you compute:

1531.4
2.67
≈
573.5
2.67
1531.4
	​

≈573.5

→ Exactly matches the JSON field.

So yfinance is dividing an INR price (₹1531) by a USD book value (2.67), which makes no sense.

✅ Correct Calculation

If we take Infosys’s real BVPS ≈ ₹231 (INR) from filings:

P/B
=
1531.4
231
≈
6.63
P/B=
231
1531.4
	​

≈6.63

That’s the correct number (and matches Screener, Smart-Investing, etc.). 

Main issue seems to be the money values are not in the same currency format.  

### Simple code that reproduces your problem

import yfinance as yf

symbol = 'INFY.NS'
ticker = yf.Ticker(symbol)
info = ticker.info

print(f"Info for {symbol}: {info}")

### Debug log from yf.enable_debug_mode()

NA

### `yfinance` version

0.2.65

ValueRaider:
Chatgpt is making assertions about the code without actually analysing the code. Give it the source code.

InfantJerin:
I dont know what is happening in the source code but it is the bug for priceToBook & it is not 573 for Infosys stock.

InfantJerin:
I just quickly did some digging - seems like this API itself might be returning the wrong priceToBook value - https://query2.finance.yahoo.com/v10/finance/quoteSummary/INFY.NS - I'm getting unauthorized error if I directly hit the API. 

you can refer to funds.py file where priceToBook fetched from API.

PR #2632, opened 2025-11-30
Sanitize priceToBook for mixed currencies

luciaxoll:
Changes were made to address Issue [#2593](https://github.com/ranaroussi/yfinance/issues/2593): 

1. Added _sanitize_price_to_book() helper method in quote.py (lines 614-631). This method detects when currency and financialCurrency fields differ in the info dictionary. When mismatched currencies are detected AND priceToBook exists, it sets priceToBook to None. This prevents propagating invalid price-to-book ratios caused by Yahoo's backend computing ratios with mismatched currencies. Other fields (currentPrice, bookValue, currency, financialCurrency) remain unchanged 

2. Integrated sanitization call in _fetch_info() method (line 690). After the info dictionary is fully populated and formatted, _sanitize_price_to_book() is called. This ensures all raw data from Yahoo is processed for currency mismatches before being returned to users.

3. Added regression test in test_ticker.py (lines 1054-1129). 

Test Results: All key tests pass:

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -1052,6 +1052,74 @@ class TestTickerInfo(unittest.TestCase):
         self.assertCountEqual(['quoteType', 'symbol', 'underlyingSymbol', 'uuid', 'maxAge', 'trailingPegRatio'], data.keys())
         self.assertIn("trailingPegRatio", data.keys(), "Did not find expected key 'trailingPegRatio' in info dict")
 
+    def test_price_to_book_mixed_currencies(self):
+        # Test for Issue #2593: priceToBook should be None when currency and financialCurrency differ.
+
+        from unittest.mock import Mock, patch
+        from yfinance.scrapers.quote import Quote
+        
+        # Test Case 1: Mixed currencies (INR price, USD book value) - the main issue
+        quote = Quote(Mock(), "TEST.NS")
+        quote._info = {
+            "symbol": "TEST.NS",
+            "currentPrice": 1531.4,  # INR
+            "bookValue": 2.67,       # USD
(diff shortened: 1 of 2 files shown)


ValueRaider:
Why is priceToBook useless when FX different? User can't do FX conversion?

luciaxoll:
Thanks for the reply!

It’s not that users can’t do FX conversion – it’s that the ratio we expose isn’t a standard P/B anymore once the currencies differ. In the case of Issue #2593, priceToBook is literally “price in INR ÷ book value in USD,” which doesn’t match Yahoo’s own UI or any conventional definition of P/B. To “fix” it, a user would need to know exactly which currency bookValue is in, which FX rate/date to use, and whether Yahoo has already transformed it upstream – none of that is available in the API (described in the Issue comment). So the 573× value is just a misleading artifact. 

So my idea is that for now: if we detect mixed currencies, we return None but still expose currentPrice, bookValue, currency, and financialCurrency so users can apply their own FX logic if they want.

Hopefully this clarifies things a bit.

ValueRaider:
> a user would need to know exactly which currency bookValue is in

Surely it the financial currency? Book is financial.

Can the sanitize be optional via an argument, e.g like I do with prices: `history(..., repair=True)`

Bigger picture, the info fetch needs a rethink - in this merge there are other data issues (I've added some debug checks)
https://github.com/ranaroussi/yfinance/blob/fc0bde214a4745fb6081a20ae81e3aba56ae72af/yfinance/scrapers/quote.py#L631-L641

```
Value mismatch: 'defaultKeyStatistics'['forwardPE'] = '20.81062', but already seen it with value '21.584623'
Value mismatch: 'dividendYield' = '2.88', but already seen it with value '0.028800001'
```

Interested?

PR #2640, opened 2025-12-11
Add overnight quote properties to FastInfo

### LINKED ISSUE #2630: Can you add the function to obtain overnight quote?
after midnight 12am, yahoo finance page https://finance.yahoo.com/quote/AAPL
can show the overnight quote of stock after the last day's close price. 

Can we add an api all to return this information?  e.g. via fast_info

AKHIL-149:
Fixes #2630

This adds support for retrieving overnight trading data (midnight to 7AM) through FastInfo.

Added six new properties:
- overnight_price
- overnight_time  
- overnight_high
- overnight_low
- overnight_open
- overnight_volume

The implementation filters the existing prepost data by timestamp, so there are no additional API calls. For symbols without overnight trading, the properties return None.

Tested with BTC-USD (has overnight data) and AAPL (returns None). All tests pass.

Example:
```python
import yfinance as yf

ticker = yf.Ticker("BTC-USD")
price = ticker.fast_info.overnight_price

if price:
    print(f"Overnight price: ${price:.2f}")
No breaking changes, all existing functionality works the same.

Much simpler and more direct.

### DIFF
--- a/test_overnight.py
+++ b/test_overnight.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+"""
+Test script for overnight quote functionality
+"""
+import yfinance as yf
+import sys
+
+def test_overnight_properties():
+    """Test overnight properties with a 24/5 trading symbol"""
+    print("=" * 60)
+    print("Testing Overnight Quote Functionality")
+    print("=" * 60)
+
+    # Test with BTC-USD (24/5 trading)
+    print("\n1. Testing with BTC-USD (24/5 trading symbol)")
+    print("-" * 60)
+    ticker = yf.Ticker("BTC-USD")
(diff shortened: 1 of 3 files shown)


ValueRaider:
@usccolumbia is this what you meant?

AKHIL-149:
@ValueRaider Yes, I have tried to work on that issue

ValueRaider:
A one-liner does it:
> df[df.index.hour.to_series().between(0,6).values]

so not seeing point of this PR.

Also this is bad Pandas:
> overnight_prices[overnight_prices.index.date == last_date]

usccolumbia:
> @usccolumbia is this what you meant?

yes. that is it. thanks.

PR #2642, opened 2025-12-16
Fix: Correct 30m interval alignment and filtering for NSE/early-open markets

### LINKED ISSUE #1436: problem with 30 mins Time frame for ^NSEI (national stock exchange of india)

biplavbarua:
Fixes #1436 

### Problem
When requesting `interval="30m"` for tickers like `^NSEI`, the first 15 minutes (09:15-09:30 IST) were dropped because they fell into a 09:00 bucket which the filter deemed "pre-market".

### Fix
1.  **`scrapers/history.py`**: Passed correct `interval` to cleaner functions.
2.  **`utils.py`**: Relaxed filtering to keep candles that *overlap* the market open (end > open), protecting the 09:00-09:30 candle.

### Verification
- Verified `^NSEI` 30m data now includes the 09:15 open.
- Passed `tests/test_prices.py` (22/22).

### DIFF
--- a/yfinance/scrapers/history.py
+++ b/yfinance/scrapers/history.py
@@ -320,8 +320,8 @@ class PriceHistory:
             })
 
         # Note: ordering is important. If you change order, run the tests!
-        quotes = utils.set_df_tz(quotes, params["interval"], tz_exchange)
-        quotes = utils.fix_Yahoo_dst_issue(quotes, params["interval"])
+        quotes = utils.set_df_tz(quotes, interval, tz_exchange)
+        quotes = utils.fix_Yahoo_dst_issue(quotes, interval)
         intraday = params["interval"][-1] in ("m", 'h')
         if not prepost and intraday and "tradingPeriods" in self._history_metadata:
             tps = self._history_metadata["tradingPeriods"]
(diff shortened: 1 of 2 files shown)


ValueRaider:
Good stuff. I've tidied it up - happy?

ValueRaider:
Also, can you rebase and target dev? https://ranaroussi.github.io/yfinance/development/code.html

ValueRaider:
You broke it - check files changed

biplavbarua:
Apologies for the noise! I've cleaned up the branch.
I reset it to `dev` and strictly re-applied the fix for the 30m interval/early-open filtering.
Verified that 'Files changed' is back to 2.
Ready for re-review. Thanks!

Issue #2617, opened 2025-10-29
Problem with Screener and Query by Industry


julioj4pa:
### Describe bug

Hi. I'm trying to query some industries, but not success 

All empty list:
Sector: Consumer Defensive - Industry: Beverages - Brewers
Sector: Consumer Defensive - Industry: Beverages - Non-Alcoholic
Sector: Consumer Defensive - Industry: Beverages - Wineries & Distilleries
Sector: Energy - Industry: Oil Gas Drilling
Sector: Energy - Industry: Oil Gas E P
Sector: Energy - Industry: Oil Gas Equipment Services
Sector: Energy - Industry: Oil Gas Integrated
Sector: Energy - Industry: Oil Gas Midstream
Sector: Energy - Industry: Oil Gas Refining Marketing
Sector: Financial Services - Industry: Banks Diversified
Sector: Financial Services - Industry: Banks Regional
Sector: Financial Services - Industry: Financial Data Stock Exchanges
Sector: Financial Services - Industry: Insurance Diversified
Sector: Financial Services - Industry: Insurance Life
Sector: Financial Services - Industry: Insurance Property Casualty
Sector: Financial Services - Industry: Insurance Reinsurance
Sector: Financial Services - Industry: Insurance Specialty
Sector: Healthcare - Industry: Diagnostics Research
Sector: Healthcare - Industry: Drug Manufacturers General
Sector: Healthcare - Industry: Drug Manufacturers Specialty Generic
Sector: Healthcare - Industry: Medical Instruments Supplies
Sector: Industrials - Industry: Aerospace Defense
Sector: Industrials - Industry: Airports Air Services
Sector: Industrials - Industry: Building Products Equipment
Sector: Industrials - Industry: Business Equipment Supplies
Sector: Industrials - Industry: Electrical Equipment Parts
Sector: Industrials - Industry: Engineering Construction
Sector: Industrials - Industry: Farm Heavy Construction Machinery
Sector: Industrials - Industry: Integrated Freight Logistics
Sector: Industrials - Industry: Pollution Treatment Controls
Sector: Industrials - Industry: Rental Leasing Services
Sector: Industrials - Industry: Security Protection Services
Sector: Industrials - Industry: Staffing Employment Services
Sector: Industrials - Industry: Tools Accessories
Sector: Real Estate - Industry: Real Estate Development
Sector: Real Estate - Industry: Real Estate Diversified
Sector: Real Estate - Industry: Reit Diversified
[...]

PR #2648, opened 2025-12-19
fix: Screener's sector industry mapping

ericpien:
the pr fixes the `Screener`'s sector industry mapping by:
1. update the sector industry for yahoo's updated industry naming 
 - yahoo seems to have added `&', '—', removed white spaces, and capitalized some words
 - addresses https://github.com/ranaroussi/yfinance/issues/2617

2. adds logic to enable emdash ('—')
 - passing the dictionary leads results in `json.dumps()` defaulting to `ensure_ascii=True` in `curl_cffi`'s `set_curl_options()` which breaks the emdash in the subsequent encoding. this is fixed by serializing before passing it as parameter 
 - serializing before sending it removes the required header `"Content-Type: application/json"` by design in `curl_cffi`'s logic. so this is fixed by adding it in parameter when `data` is passed by `Screener` instead of `json`.
 - addresses https://github.com/ranaroussi/yfinance/issues/2272
 
## sample working codes:
``` python
import yfinance as yf
from yfinance.screener import EquityQuery

MIN_MCAP_USD = 1_000_000_000

q = EquityQuery(
    "and",
    [
        EquityQuery('is-in', ['industry', 'Oil & Gas E&P']),
        EquityQuery("gte", ["intradaymarketcap", MIN_MCAP_USD]),
    ],
)
resp = yf.screener.screen(q)
print(str(resp))

q = EquityQuery('eq', ["industry", 'Insurance—Specialty'])
resp = yf.screener.screen(q)
print(str(resp))
```

image


ValueRaider:
Thanks, but to simplify the diff can you maintain the ordering?

PR #2656, opened 2025-12-24
Fix missing pegRatio in .info (Fixes #2570)

### LINKED ISSUE #2570: pegRatio missing from yfinance.info since June 2025 (worked before)
### Describe bug

The `pegRatio` field from `yfinance.Ticker(ticker).info` suddenly returns `None` for many tickers.

Until May 30, 2025, this worked fine for stocks like `AAPL`, `MSFT`, `GOOGL`. Since June, it seems to have silently broken – the field is missing entirely in the `.info` dictionary.

I checked that the PEG ratio is still visible on Yahoo Finance directly, so the data exists.

import yfinance as yf

ticker = yf.Ticker("AAPL")
info = ticker.info

print("PEG Ratio:", info.get("pegRatio"))  # returns None

PEG Ratio key missing from .info output. Tried with `yf.enable_debug_mode()`, but no specific error is shown – PEG key just doesn't appear anymore in the dictionary.

### Bad data proof

The PEG ratio is still listed here:
https://finance.yahoo.com/quote/AAPL/key-statistics/

Example:
PEG Ratio (5 yr expected): 1.97 (as of July 2025)

But `.info.get("pegRatio")` returns `None`.

### `yfinance` version

0.2.65

### Python version

3.11.2

### Operating system

Raspberry Pi OS Bookworm (Linux 6.12.34, armv7l)

welcra:
Verified that pegRatio is missing from both standard modules and the timeseries API, which caused previous attempts to fail. Implemented a fallback that parses the Key Statistics HTML page using pd.read_html when the API returns None.

### DIFF
--- a/yfinance/scrapers/quote.py
+++ b/yfinance/scrapers/quote.py
@@ -3,6 +3,7 @@ import datetime
 import json
 import numpy as _np
 import pandas as pd
+from io import StringIO
 
 from yfinance import utils
 from yfinance.config import YfConfig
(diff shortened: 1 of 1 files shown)


ValueRaider:
Because `key-statistics` is a separate page, then the fetch and accessor should be separate from `info`. E.g.:

```
@property
def statistics(self):
  return self._statistics.get()
```

Also fix that git history - you need to rebase to dev.

Issue #2686, opened 2026-01-27
Random Volume zero values


krydazo:
### Describe bug

Random Volume 0 values. They're in about half of the output. I see no pattern otherwise. The Yahoo Finance site doesn't let me check intra-day intervals.

Your logging hint below doesn't specify where to get the logging output from; it links to https://github.com/ranaroussi/yfinance/tree/main#logging which shows the same exact thing as https://github.com/ranaroussi/yfinance/tree/main. Searching for "logging" in either page shows 0 results. The hint appears to be outdated.

### Simple code that reproduces your problem

yf.download(tickers='BTC-USD', period='1y', interval='4h')

### Debug log from yf.enable_debug_mode()

The link specified doesn't show how to get the logging output.

### Bad data proof

First 5 rows:
Price                             Close           High           Low          Open      Volume   Returns          MA20           MA50        RSI
Ticker                          BTC-USD        BTC-USD       BTC-USD       BTC-USD     BTC-USD                                                  
Datetime                                                                                                                                        
2025-02-04 16:00:00+00:00  99188.710938  100737.468750  98856.343750  99416.734375           0 -0.002548  99082.402344  101547.333750  50.547037
2025-02-04 20:00:00+00:00  97887.187500   99211.312500  96223.234375  99040.710938  4382146560 -0.013122  98873.271484  101478.407812  46.440381
2025-02-05 00:00:00+00:00  98079.820312   98933.570312  97251.804688  97807.843750   977043456  0.001968  98695.653906  101447.943437  49.683217
2025-02-05 04:00:00+00:00  97920.804688   98313.703125  97436.570312  98064.625000   340377600 -0.001621  98559.483984  101363.494219  50.652722
2025-02-05 08:00:00+00:00  98232.234375   98242.726562  97317.171875  97909.187500   686034944  0.003180  98511.331641  101275.655625  63.823232

Last 5 rows:
Price                             Close          High           Low          Open      Volume   Returns          MA20          MA50        RSI
Ticker                          BTC-USD       BTC-USD       BTC-USD       BTC-USD     BTC-USD                                                 
Datetime                                                                                                                                      
2026-01-26 20:00:00+00:00  88307.937500  88548.187500  87453.585938  87912.304688           0  0.004413  88454.824219  89809.289062  42.263346
2026-01-27 00:00:00+00:00  88640.945312  88759.000000  88102.992188  88190.843750           0  0.003771  88394.151953  89675.416719  45.247839
[...]

PR #2687, opened 2026-01-28
Docs: clarify debug logging config and deprecate enable_debug_mode

KarthickKrishnasaisha:
Align logging documentation with the config-based API (`yf.config.debug.logging`) and clarify
where debug logs are emitted.

Marks `enable_debug_mode()` as deprecated in docs while keeping it documented for backward
compatibility.

Addresses logging confusion reported in #2686.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -62,8 +62,14 @@ body:
   - type: textarea
     id: debug-log
     attributes:
-      label: "Debug log from yf.enable_debug_mode()"
-      description: "Run code with debug logging enabled - `yf.enable_debug_mode()` - and post the full output. Context: https://github.com/ranaroussi/yfinance/tree/main#logging"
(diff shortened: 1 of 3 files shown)


ValueRaider:
Simply replace mentions of `enable_debug_mode` in docs with `yf.config.debug.logging = True`. yfinance already prints a deprecation warning when user calls it, don't need docs duplicating.

Source of confusion is bug report template, that needs updating.

These changes can go straight to main branch.

KarthickKrishnasaisha:
Thanks for the clarification.

I’ve updated the docs to only reference `yf.config.debug.logging = True` and removed all mentions of `enable_debug_mode`, per your guidance.

I also fixed the issue form (`bug_report.yaml`), which was the source of the confusion, so new bug reports now request logs using the config-based approach.

The PR is now targeting `main`.

ValueRaider:
> The PR is now targeting main.

Also rebase to main, get that commit log clean.

ValueRaider:
You've got someone else's commit in your branch.

PR #2695, opened 2026-02-15
Fix/fill devidends

david-belbeze:
**Subject:** Fix TypeError in `Ticker.history` when dividends contain currency suffixes (e.g., 'USD')

### Description

This pull request fixes a `TypeError` occurring in the `Ticker.history` API when dividend data contains strings (like 'USD' suffixes) instead of numeric values.

### Reproduction Case

The following code fails when fetching history for `BBOX`:

```python
import yfinance as yf

ticker_obj = yf.Ticker('BBOX')
prices = ticker_obj.history(period="max", interval="1d", auto_adjust=True)

print(prices)

```

**Error output:**
`TypeError: Invalid value '0' for dtype 'str'. Value should be a string or missing value, got 'int' instead.`

### Analysis

During execution, I observed that the "Dividends" column is sometimes cast as a `str` type. This happens because some dividend values include a currency suffix (e.g., '3.14 USD'), which often occurs when the stock is traded in a different currency (like GBX).

### Fix

The proposed fix strips the currency suffix and converts the dividend strings back to numeric values (float64), ensuring the DataFrame remains consistent.

### Side Effects & Limitations

* **Currency Conversion:** This fix prevents the crash but does not perform currency conversion. The raw numeric value is returned regardless of the original currency suffix.
* **Metadata:** In the case of `BBOX`, the currency metadata is returned as `None`, making automatic conversion difficult at this stage.
* **Data Accuracy:** While the API no longer crashes, users should be aware that the dividend value might be in a different currency than the price.

### DIFF
--- a/yfinance/scrapers/history.py
+++ b/yfinance/scrapers/history.py
@@ -358,8 +358,10 @@ class PriceHistory:
 
         if splits is not None:
             splits = utils.set_df_tz(splits, interval, tz_exchange)
+            self._splits = splits
         if dividends is not None:
             dividends = utils.set_df_tz(dividends, interval, tz_exchange)
+            self._dividends = dividends
             if 'currency' in dividends.columns:
                 # Rare, only seen with Vietnam market
                 #   or companies that distribute dividends in a different currency
@@ -371,20 +373,11 @@ class PriceHistory:
                     if repair and price_currency != '':
                         # Attempt repair = currency conversion
                         dividends = self._dividends_convert_fx(dividends, price_currency, repair)
-                        # FX conversion can fail, in that case, save as metadata
-                        save = (dividends['currency'] != price_currency).any()
-                    else:
-                        # No repair, but currency is important information to save
-                        save = True
-                    
-                    if save and not dividends.empty:
-                        # Save the first dividend currency to allow the user to perform 
-                        #   any conversion or analysis
-                        self._history_metadata['dividendCurrency'] = dividends['currency'].iloc[0]
                 dividends = dividends.drop('currency', axis=1)
 
         if capital_gains is not None:
             capital_gains = utils.set_df_tz(capital_gains, interval, tz_exchange)
+            self._capital_gains = capital_gains
         if start is not None:
             if not quotes.empty:
                 start_d = quotes.index[0].floor('D')
@@ -534,19 +527,19 @@ class PriceHistory:
             self._reconstruct_start_interval = None
         return df
 
-    def _get_history_cache(self, period="max", interval="1d") -> pd.DataFrame:
-        cache_key = (interval, period)
-        if cache_key in self._history_cache:
-            return self._history_cache[cache_key]
-
-        df = self.history(period=period, interval=interval, prepost=True)
-        self._history_cache[cache_key] = df
-        return df
+    def _get_history_cache(self, period="max", interval="1d", repair=False) -> pd.DataFrame:
+        cache_key = (interval, period, repair)
+        if cache_key not in self._history_cache.keys():
+            df = self.history(period=period, interval=interval, repair=repair, prepost=True)
+            self._history_cache[cache_key] = {'prices': df, 'dividends': self._dividends, 
+                                                'splits': self._splits, 
+                                                'capital gains': self._capital_gains}
+        return self._history_cache[cache_key]
 
     def get_history_metadata(self) -> dict:
         if self._history_metadata is None or 'tradingPeriods' not in self._history_metadata:
             # Request intraday data, because then Yahoo returns exchange schedule (tradingPeriods).
-            self._get_history_cache(period="5d", interval="1h")
+            self._get_history_cache(period="5d", interval="1h")['prices']
 
         if self._history_metadata_formatted is False:
             self._history_metadata = utils.format_history_metadata(self._history_metadata)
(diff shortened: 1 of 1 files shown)


ValueRaider:
Should store the dividend currency in metadata, so user can convert.

david-belbeze:
While analyzing the initial processing of the dividend DataFrame, I found that the code was merging a 2D DataFrame (value and currency) into a single column.

I have improved this logic and removed my previous fix to prevent incorrect conversions. As a result, the entire dividend DataFrame is now added in the metadata, allowing it to handle multiple currencies over the years.

ValueRaider:
I'm finally testing this but seems `BBOX` is delisted now. You got other tickers?

david-belbeze:
It's strange, I got the error on the dev branch today for `BBOX`.

You could test `TEP` instead (it's the only other ticket that causes the TypeError on my studied assets).

ValueRaider:
I was only fetching last year. max works.

I forgot yfinance had this already:
> dividends['Dividends'].astype(str) + ' ' + dividends['currency']

Restore that instead of putting dividends in metadata (also your metadata write is being lost).

david-belbeze:
Perhaps my previous explanation wasn't clear enough. Here is the root cause analysis:

**The Issue:**

My investigation shows that a type conversion in the dividends column is causing the `TypeError`.

In the `BBOX` example, the bug originates here:
https://github.com/ranaroussi/yfinance/blob/81631009a20bf682dc3d6799e954fb49af770580/yfinance/scrapers/history.py#L359

Because the `Dividends` column is converted to `str` (to include the currency), the subsequent assignment fails:
https://github.com/ranaroussi/yfinance/blob/81631009a20bf682dc3d6799e954fb49af770580/yfinance/scrapers/history.py#L405

Pandas does not allow filling a `str` dtype Series with an integer (`0`).

**Reproduction:**

I have created a dedicated branch to demonstrate this:  https://github.com/david-belbeze/yfinance/tree/fix/fill-devidends-debug (last commit)

* The script [`bug_dividends.py`](https://github.com/david-belbeze/yfinance/blob/fix/fill-devidends-debug/bug_dividends.py) reproduces the `TypeError` with `BBOX` asset (I have added some print to help debugging).
* The script [`bug_pandas_reproduction.py`](https://github.com/david-belbeze/yfinance/blob/fix/fill-devidends-debug/bug_pandas_reproduction.py) is a minimal standalone script reproducing the Pandas logic failure.

```console
$ python bug_pandas_reproduction.py
Prepare data
            Close
date             
2026-02-20  10.00
2026-02-21  10.50
2026-02-22  10.25
2026-02-23   9.50
            Dividends currency
date                          
2026-02-21        1.0      USD

Concat dividends
           Dividends
date                
2026-02-21   1.0 USD

Merge data frames
            Close Dividends
date                       
2026-02-20  10.00       NaN
2026-02-21  10.50   1.0 USD
2026-02-22  10.25       NaN
2026-02-23   9.50       NaN
Close        float64
Dividends        str
dtype: object
Reproduce TypeError
  File "***/yfinance/bug_pandas_reproduction.py", line 38, in <module>
    df.loc[df["Dividends"].isna(), "Dividends"] = 0
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ...
TypeError: Invalid value '0' for dtype 'str'. Value should be a string or missing value, got 'int' instead.
```

**Proposed Solution:**

My PR fixes this by keeping the dividends as numeric values and storing the original raw data (including currency) in the metadata instead. I believe providing access to the raw data is cleaner than forced string concatenation.

Does this approach work for you, or would you prefer a different fix?

ValueRaider:
The reason I converted to str and appended currency, was to force user to notice and decide what to do.

Also your solution discards the data anyway - did you test it?

david-belbeze:
I've tested the solution: `Dividends` are indeed returned as `float64` without currency. However, it is possible to access the original dividend data via the `PriceHistory` metadata:

```python
# ...
ticker_obj = yf.Ticker('BBOX')
prices = ticker_obj.history(period="max", interval="1d", auto_adjust=True)
# Accessing original metadata
print(ticker_obj._price_history._history_metadata['originalDividends'])
```

If we want to include the currency directly in the `prices` DataFrame, we have two main approaches:

* Change Dividends to `str` (e.g., "1.50 USD") after dividends operations in `PriceHistory.history` method: This forces the user to parse the string to perform calculations.

* Add a `DividendsCurrency` column: This ensures `Dividends` remains a `float64` for easy math, but makes the currency metadata optional (and potentially overlooked).

Trade-offs:

* Option 1 is a breaking change for users expecting numeric values but keeps data bundled.
* Option 2 is cleaner for API robustness and preserves type safety.

Personally, I find Option 2 more robust, though Option 1 is acceptable if well-documented. What do you think is the best fit for the library's philosophy?

ValueRaider:
3rd option: add new private variable to `PriceHistory` for storing dividend currency. Then whenever `get_history_metadata` is called, add div FX to it. Use dict key `dividendCurrency`

david-belbeze:
`get_history_metadata` calls the `history` method internally with a fixed period and interval.

I replaced the `originalDividends` key with `dividendCurrency`, using the first currency value found in the dividend dataframe.

Note: This does not inform the user if multiple currencies were used over the years.

ValueRaider:
I have a better solution - put it in `Ticker.dividends`. Can you let me push to your fork? Easier to just commit.

david-belbeze:
You are now a member of my fork

ValueRaider:
Try this.

david-belbeze:
My script accurately retrieves prices and dividends for all tickers, regardless of whether a currency is specified.

PR #2702, opened 2026-02-22
Add currency properties for earnings trend data

Sathv1kA:
Expose earningsCurrency, revenueCurrency, epsTrendCurrency, and epsRevisionsCurrency from the earningsTrend API response. These can differ from financialCurrency (e.g. TM has earningsCurrency=USD but financialCurrency=JPY).

Fixes #<2699>

### DIFF
--- a/yfinance/base.py
+++ b/yfinance/base.py
@@ -315,7 +315,17 @@ class TickerBase:
         """
         data = self._analysis.revenue_estimate
         return data.to_dict() if as_dict else data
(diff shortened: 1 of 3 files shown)


ValueRaider:
What if these returned FX with the original fetch? E.g. `eps_trend` returns tuple: dataframe and currency.

Sathv1kA:
Thats a great Idea!

Issue #2707, opened 2026-03-01
Documentation for Tickers.{history,download}() does not match function signature


pjmcdermott:
### Describe bug

The documentation at [yfiance.Ticker documentation](https://ranaroussi.github.io/yfinance/reference/api/yfinance.Tickers.html#yfinance.Tickers) states that the default for the 'period' parameter for these methods is None.

The actual signature from at least yfinance==0.2.66 is that the default for the period parameter is '1mo'.

There appears to be some recent change around 0.2.66 whereby yfinance is now throwing an error "ValueError: Setting period, start and end is nonsense. Set maximum 2 of them." which draws attention to this discrepancy.

Moreover, as a matter of style I believe it would be better to through an error about "too many parameters" directly at the user entry point, Tickers.{download,history}.

I would also note that the default 'period' parameter for PirceHistory.history() is different from Tickers.{download,history}().

### Simple code that reproduces your problem

n/a

### Debug log with debug logging enabled

n/a

### `yfinance` version

0.2.66

### Python version

Python 3.12.3

### Operating system

Linux larkfield 6.17.0-14-generic #14~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jan 15 15:52:10 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

ValueRaider:
Would presenting the `Tickers.history` docstring solve like here, and clarifying it depends on start & end being None? https://ranaroussi.github.io/yfinance/reference/yfinance.price_history.html#yfinance.scrapers.history.PriceHistory.history

> default 'period' parameter for PirceHistory.history() is different from Tickers.{download,history}().

I don't understand - `yf.Tickers('msft aapl goog').history()` returns 1 month.

pjmcdermott:
@ValueRaider I'm not sure I really understand your comment.

My point was that the method documentation does not match the actual signature. It appears that a default of `period=None` for some methods has changed to `period="1mo"`, but the documentation has not been updated.

I was specifying both the start and end dates in my method calls and assuming the period was set to `None`, as documented. But with the change in default parameter my method calls were no longer valid. It appears that recently a new error message has been inserted alerting users to this fact which is now causing my method calls to fail.

My key points are:

1. Documentation does not match actual method signatures.
2. I believe it is preferable to raise issues about bad user inputs as high up the calling stack as possible. Currently there are a number of deeper method calls before the warning is issued.
[...]

PR #2712, opened 2026-03-05
Tickers: validate conflicting date args before download

monkeyjack123:
## Summary\n- add explicit validation in  /  for conflicting date arguments\n- raise a clearer  when , , and  are all provided\n- add unit tests for both the validation path and a valid forwarding path\n\n## Why\nIssue #2707 reports confusing behavior around date argument handling on the  entry points. This keeps the existing constraint but surfaces it earlier with clearer wording.\n\nCloses #2707

### DIFF
--- a/tests/test_tickers.py
+++ b/tests/test_tickers.py
@@ -1,0 +1,23 @@
+import unittest
+from unittest.mock import patch
+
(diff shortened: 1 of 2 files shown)


monkeyjack123:

PR #2714, opened 2026-03-05
Docs: point Market reference link to detailed module page

### LINKED ISSUE #2358: Market API documentation
From the [api docs](https://yfinance-python.org/reference/index.html), the [market](https://yfinance-python.org/reference/api/yfinance.Market.html#yfinance.Market) link leads to an unhelpful autogenerated stub. The useful class information is contained in [the market class description](https://yfinance-python.org/reference/yfinance.market.html), which I think would be a better ref for the link

monkeyjack123:
## Summary
- update the API reference overview to link **Market** to the module docs page (`yfinance.market`) instead of the autogenerated class stub
- keeps the description text unchanged

## Why
Issue #2358 notes that the current Market link lands on an unhelpful stub page, while the module page contains the useful class documentation.

Closes #2358


monkeyjack123:

PR #2715, opened 2026-03-05
fix(history): guard null chart payload before result access

### LINKED ISSUE #2670: NoneType protection fails
### Describe bug

Line 224
yfinance/scrapers/history.py:
`if data['chart']['result'] is None:`

This line fails if `data['chart']` itself is None.

Recommended fix, since I don't know the process to push:

`if data.get('chart', {}).get('result') is None:`

n/a

n/a

### Bad data proof

_No response_

### `yfinance` version

1.0

### Python version

_No response_

### Operating system

_No response_

monkeyjack123:
## Summary
- guard data['chart'] access when Yahoo returns a null/non-dict chart payload
- reuse chart_data for metadata/error/result checks to avoid NoneType crashes
- keep existing YFPricesMissingError flow when history payload is invalid

Fixes #2670.

### DIFF
--- a/yfinance/scrapers/history.py
+++ b/yfinance/scrapers/history.py
@@ -221,10 +221,11 @@ class PriceHistory:
                 raise
 
         # Store the meta data that gets retrieved simultaneously
-        if data['chart']['result'] is None:
+        chart_data = data.get("chart") if isinstance(data, dict) else None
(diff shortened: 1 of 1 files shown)


monkeyjack123:

PR #2724, opened 2026-03-18
Docs: Add docstrings to Tickers.history() and Tickers.download()

### LINKED ISSUE #2707: Documentation for Tickers.{history,download}() does not match function signature

lindsey-comet:
## Summary

Fixes #2707

Added comprehensive docstrings to `Tickers.history()` and `Tickers.download()` methods to document the correct default value for the `period` parameter.

## Problem

The auto-generated documentation stated that the default for `period` parameter was `None`, but the actual default behavior is `'1mo'` (enforced by `multi.download()`). This caused confusion when users tried to call these methods.

## Changes

- Added docstrings to both `history()` and `download()` methods in `tickers.py`
- Documented that the default `period` value is `'1mo'`
- Listed all valid period values
- Documented all other parameters following existing patterns in `multi.download()`

## Testing

- `python3 -m py_compile yfinance/tickers.py` passes
- Syntax validated

## Quality Checklist

- [x] Documentation fix (addresses an actual issue)
- [x] Clear commit message with issue reference
- [x] Follows existing documentation patterns in the codebase
- [x] Single focused change (documentation only)

### DIFF
--- a/yfinance/tickers.py
+++ b/yfinance/tickers.py
@@ -51,7 +51,41 @@ class Tickers:
                 actions=True, auto_adjust=True, repair=False,
                 threads=True, group_by='column', progress=True,
                 timeout=10, **kwargs):
+        """
+        Fetch historical market data for multiple tickers.
 
+        :Parameters:
+            period : str
+                Valid periods: 1d,5d,1mo,3mo,6mo,1y,2y,5y,10y,ytd,max
+                Default: '1mo'
+            interval : str
+                Valid intervals: 1m,2m,5m,15m,30m,60m,90m,1h,1d,5d,1wk,1mo,3mo
+            start : str
+                Download start date string (YYYY-MM-DD) or datetime, inclusive.
+            end : str
+                Download end date string (YYYY-MM-DD) or datetime, exclusive.
+            prepost : bool
+                Include Pre and Post market data in results? Default is False.
+            actions : bool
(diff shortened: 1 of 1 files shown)


ValueRaider:
Already fixed #2709

PR #2703, opened 2026-02-22
Fix TypeError in history() when API returns None/malformed

a-drenaline:

PR #2726, opened 2026-03-20
Fix KeyError in FastInfo for delisted/unavailable tickers

### LINKED ISSUE #1951: fast_info returning error on non available quotes
### Describe bug

I have seen that fast info when passed a ticker that is not currently available, an error occurs due to the reading of non-existing parameters, in this case, currency.

import yfinance as yf

stock = yf.Ticker("FDQ.BE")

stock.get_fast_info() # same with stock.fast_info

### Debug log

  File "/home/user/code/src/functions.py", line 11, in get_info_data
    fast_info = stock.get_fast_info()
  File "/home/user/code/venv/lib/python3.10/site-packages/yfinance/scrapers/quote.py", line 165, in __getitem__
    return getattr(self, k)
  File "/home/user/code/venv/lib/python3.10/site-packages/yfinance/scrapers/quote.py", line 255, in currency
    self._currency = md["currency"]
KeyError: 'currency'

### Bad data proof

I think it could be easily solved by doing .get() of the dictionary. Knowing that all the response parameters have a default value of None, the get() would make them keep the None value and return a dictionary with those parameters to None as the user expects instead of throwing an error.

### `yfinance` version

0.2.40

### Python version

3.10.12

### Operating system

Ubuntu 22.04

Bortlesboat:
Fixes #1951

Several FastInfo properties use unsafe direct dict access that raises KeyError when Yahoo returns incomplete metadata for delisted or unavailable tickers.

## Changes
- Replace unsafe dict["key"] accesses with dict.get("key") in FastInfo properties
- Properties now return None gracefully instead of raising KeyError
- Follows the same defensive pattern established in #2703

## Testing
- Tested with known delisted tickers that previously crashed

### DIFF
--- a/yfinance/scrapers/quote.py
+++ b/yfinance/scrapers/quote.py
@@ -130,15 +130,19 @@ class FastInfo:
             self._prices_1y = self._tkr.history(period="1y", auto_adjust=False, keepna=True)
             self._md = self._tkr.get_history_metadata()
             try:
-                ctp = self._md["currentTradingPeriod"]
-                self._today_open = pd.to_datetime(ctp["regular"]["start"], unit='s', utc=True).tz_convert(self.timezone)
-                self._today_close = pd.to_datetime(ctp["regular"]["end"], unit='s', utc=True).tz_convert(self.timezone)
-                self._today_midnight = self._today_close.ceil("D")
+                ctp = self._md.get("currentTradingPeriod")
+                if ctp is not None and "regular" in ctp and self.timezone is not None:
+                    self._today_open = pd.to_datetime(ctp["regular"]["start"], unit='s', utc=True).tz_convert(self.timezone)
+                    self._today_close = pd.to_datetime(ctp["regular"]["end"], unit='s', utc=True).tz_convert(self.timezone)
+                    self._today_midnight = self._today_close.ceil("D")
+                else:
+                    self._today_open = None
+                    self._today_close = None
+                    self._today_midnight = None
             except Exception:
                 self._today_open = None
                 self._today_close = None
                 self._today_midnight = None
-                raise
 
         if self._prices_1y.empty:
             return self._prices_1y
@@ -198,7 +202,7 @@ class FastInfo:
             return self._currency
 
         md = self._tkr.get_history_metadata()
-        self._currency = md["currency"]
+        self._currency = md.get("currency")
         return self._currency
 
     @property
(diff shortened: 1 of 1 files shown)


yfinance/scrapers/quote.py
```diff
@@ -130,15 +130,19 @@ def _get_1y_prices(self, fullDaysOnly=False):
...
-                self._today_midnight = self._today_close.ceil("D")
+                ctp = self._md.get("currentTradingPeriod")
+                if ctp is not None and "regular" in ctp and self.timezone is not None:
+                    self._today_open = pd.to_datetime(ctp["regular"]["start"], unit='s', utc=True).tz_convert(self.timezone)
+                    self._today_close = pd.to_datetime(ctp["regular"]["end"], unit='s', utc=True).tz_convert(self.timezone)
+                    self._today_midnight = self._today_close.ceil("D")
+                else:
+                    self._today_open = None
+                    self._today_close = None
+                    self._today_midnight = None
```
ValueRaider, line 141: Is this change redundant? Does simply removing the `raise` below fix this problem?

ValueRaider:
Instead can `FastInfo` detect delisted and raise `YFTickerMissingError`?

yfinance/scrapers/quote.py
Bortlesboat: Good point — removing the `raise` would fix the crash in `_get_1y_prices` specifically, since the except block already sets `None` fallbacks. The `.get()` changes there are redundant with that.

However the `raise` removal alone doesn't cover `currency`, `quote_type`, `exchange`, and `timezone` properties — those do bare `md["key"]` accesses outside any try/except, so they'd still KeyError on delisted tickers.

I like your suggestion of raising `YFTickerMissingError` instead — that's more explicit than silently returning `None`. I can rework the PR to:
1. Remove the `raise` in `_get_1y_prices` (your suggestion)
2. Wrap the other property accesses to raise `YFTickerMissingError` when metadata keys are missing

Want me to go that direction?

ValueRaider: 1. Intercept KeyError and try to replace with a meaningful exception e.g. https://github.com/ranaroussi/yfinance/blob/4013a4fd745f6f3f324565eae685e7765b2a2714/yfinance/scrapers/history.py#L253-L255

2. To be consistent with rest of codebase, give user option to hide exceptions e.g. https://github.com/ranaroussi/yfinance/blob/4013a4fd745f6f3f324565eae685e7765b2a2714/yfinance/scrapers/history.py#L273-L276

Issue #2699, opened 2026-02-16
Earnings Forecast currency does not always match financialCurrency


etbala:
**Description:**
When retrieving earnings forecasts, the currency usually matches `financialCurrency`, but there are exceptions.

**Example:**
Toyota Motor Corporation (TM) is listed on NYSE.
* `currency`: 'USD'
* `financialCurrency`: 'JPY'

However, the earnings estimates are listed in USD, diverging from the financialCurrency.

**Financials Page:**

<img width="1038" height="556" alt="Image" src="https://github.com/user-attachments/assets/0891110a-5443-4746-8176-03d12acfaace" /><br>

**Analysis Page:**

<img width="1030" height="299" alt="Image" src="https://github.com/user-attachments/assets/e9db68d3-349e-493c-902f-2631a209dc90" /><br>

**Code:**
```py
import yfinance as yf

ticker = yf.Ticker("TM")

print(f"Listing Currency: {ticker.info.get('currency')}")
print(f"Financial Currency: {ticker.info.get('financialCurrency')}")

print(ticker.earnings_estimate)
```
**Console Output:**
```
Listing Currency: USD
Financial Currency: JPY
           avg     low    high  numberOfAnalysts  yearAgoEps  growth
period
0q       2.686   2.686   2.686                 1         NaN     NaN
+1q      4.765   4.765   4.765                 1     4.37944  0.0880
0y      18.366  18.052  18.680                 2    24.77793 -0.2588
+1y     22.162  21.061  24.135                 3    18.36600  0.2067
```
_Note: Testing done on yfinance version `1.2.0`_

**Proposal:**
Please add a method or attribute to retrieve the specific currency used for the listed forecasts so these discrepancies can be handled programmatically.

rshah98626:
Looking into this 👀

PR #2739, opened 2026-04-04
Add Currency to Analysis Data

etbala:
Implements #2699.

Updates `earnings_estimate`, `revenue_estimate`, `eps_trend`, and `eps_revisions` to return a currency along with the same dataframe (as a tuple -> returns `[data, currency]` instead of just `data`).


ValueRaider:
I'd like to avoid changing the return format when possible. Can currency be a new column?

etbala:
Implemented.

Issue #2370, opened 2025-03-18
Adding Username and Password to query


ClarkSims:
Yahoo has changed the data which it offers for free. In order to download historical data for ETF's or currencies, one now needs to log in with your username and password. You need to have a gold account which cost about $500 per year.

Is there a way to pass the username and password to yfinance objects, so they can access the same data?

ClarkSims:
I couldn't find the branch with this code.  How do I clone this branch?
I grepped for Auth and check_login in the dev branch. So I don't see it in dev yet.
Am I doing something wrong? Should it be in dev?
Thanks,
Clark

ValueRaider:
why are you looking in dev? it hasn't been merged

PR #2254, opened 2025-02-01
Feature: Login

dhruvan2006:
New `Auth` class that allows for login functionality by settings cookies.

```python
import yfinance as yf

auth = yf.Auth()
auth.set_login_cookies(os.getenv("COOKIE_T"), os.getenv("COOKIE_Y"))
auth.check_login()
auth.user
```

Inspired by #2220

PR #2761, opened 2026-04-20
Add Auth class for Logging In

etbala:
Revival of #2254. Adds an `Auth` class that allows user to login using manually retrieved cookies.

Implements #2370.

```py
import yfinance as yf
import os

auth = yf.Auth()
auth.set_login_cookies(os.getenv("COOKIE_T"), os.getenv("COOKIE_Y"))
if auth.check_login():
    print("Logged in")
    print(auth.user)
else:
    print("Invalid cookie")
```

I considered adding a `check_login` call at the end of `set_login_cookies` but decided against it. `check_login` is expensive (~1 second), and we don't want to give a false positive if the cookies the user provided expired after being cached. Curious if you have any thoughts on this.

Also just noting that the T/Y cookies expire after 1 year.

### DIFF
--- a/doc/source/reference/examples/auth.py
+++ b/doc/source/reference/examples/auth.py
@@ -1,0 +1,18 @@
+import yfinance as yf
+import os
+
+auth = yf.Auth()
+
+# Set log in cookies from browser
+auth.set_login_cookies(os.getenv("COOKIE_T"), os.getenv("COOKIE_Y"))
+
+# Check if the cookies worked
+if auth.check_login():
+    print("Logged in")
+else:
+    print("Invalid cookie")
+
+# Every subsequent request sent to Yahoo Finance will now be under the logged-in user.
+
+# Access user information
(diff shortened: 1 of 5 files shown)


ValueRaider:
Is there a way to get crumb for these cookies? The real proof that these cookies work, is to stop calling `_get_cookie_basic`

etbala:
I added `self._cookie = True` in `set_login_cookies` so `_get_basic_cookie` reuses ours instead of overwriting it.

I was able to get a crumb using the following:
```py
import yfinance as yf
from yfinance.data import YfData
import os

cookie_t = os.getenv("COOKIE_T")
cookie_y = os.getenv("COOKIE_Y")

data = YfData()
auth = yf.Auth()
auth.set_login_cookies(cookie_t, cookie_y)
crumb = data._get_crumb_basic()
print(f"{crumb}")
```

PR #2795, opened 2026-04-29
Fix get_history_metadata() raising YFPricesMissingError for valid tickers (#2745)

### LINKED ISSUE #2745: get_history_metadata() raises YFPricesMissingError for valid tickers when intraday data is unavailable

Jaypatel1511:
## Summary

Fixes #2745 — `get_history_metadata()` raised `YFPricesMissingError` for valid, non-delisted tickers (e.g. `ROG.SW`) even after `history()` succeeded, because the optional intraday enrichment call failed and was not caught.

Two bugs fixed in `scrapers/history.py`:

### Bug 1 — good metadata overwritten by failed fetch (line ~240)

`history()` unconditionally wrote `self._history_metadata = meta` after every fetch. When the intraday enrichment request returned no data, `meta` was `{}`, silently destroying the `currency`, `exchangeName`, `timezone`, etc. already cached from a prior successful `history()` call.

**Fix:** only overwrite `_history_metadata` when the new meta is non-empty, or when it has never been set:

```python
# Before
self._history_metadata = meta

# After
if meta or self._history_metadata is None:
    self._history_metadata = meta
```

### Bug 2 — YFPricesMissingError propagates from optional enrichment call

`get_history_metadata()` called `_get_history_cache(period="5d", interval="1h")` to populate `tradingPeriods`. For some tickers/exchanges (e.g. Swiss exchange), intraday data is unavailable and this raised `YFPricesMissingError`, even though all core metadata was already cached.

**Fix:** catch `YFPricesMissingError` around the enrichment call and return whatever metadata is already available:

```python
# Before
self._get_history_cache(period="5d", interval="1h")['prices']

# After
try:
    self._get_history_cache(period="5d", interval="1h")['prices']
except YFPricesMissingError:
    pass  # tradingPeriods is optional enrichment; return whatever metadata we have
```

## Test plan

- [x] Added `tests/test_get_history_metadata.py` with 4 regression cases
- [x] Verified good metadata is preserved after a failed intraday fetch
- [x] Verified `get_history_metadata()` returns cached metadata instead of raising
- [x] All 23 unit tests pass

### DIFF
--- a/tests/test_get_history_metadata.py
+++ b/tests/test_get_history_metadata.py
@@ -1,0 +1,125 @@
+"""
+Regression tests for issue #2745:
+get_history_metadata() raised YFPricesMissingError for valid tickers (e.g. ROG.SW)
+when the intraday enrichment request returned no data, even though history() had
+already populated _history_metadata with currency, exchangeName, timezone, etc.
+
+Two bugs fixed:
+  1. history() always overwrote _history_metadata with {} on a failed fetch,
+     destroying previously cached good metadata.
+  2. get_history_metadata() let the optional intraday fetch propagate
+     YFPricesMissingError instead of returning the metadata already cached.
+"""
+import unittest
+from unittest.mock import MagicMock, call, patch
+
+from yfinance.scrapers.history import PriceHistory
+from yfinance.exceptions import YFPricesMissingError
+
+
+def _make_ph(json_payload):
+    """Return a PriceHistory whose HTTP layer returns json_payload."""
+    mock_resp = MagicMock()
+    mock_resp.text = ""
+    mock_resp.json.return_value = json_payload
+
+    mock_data = MagicMock()
+    mock_data.get.return_value = mock_resp
+    mock_data.cache_get.return_value = mock_resp
+
+    return PriceHistory(data=mock_data, ticker="FAKE", tz="America/New_York")
+
+
+def _good_chart(meta: dict) -> dict:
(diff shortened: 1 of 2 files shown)


etbala:
`get_history_metadata` is used in `yf.Ticker`, so the tests for it should go in `test_ticker.py`.

Also this if statement does do what you said, but the way it is written it reads like you are checking if either meta or history_metadata is None, something a little more explicit might be nice.
```py
if meta or self._history_metadata is None:
    self._history_metadata = meta
```

ValueRaider:
> but the way it is written it reads like ...

@etbala TBF codebase is already littered with similar statements, mostly in `history.py`

ValueRaider:
rebase to dev https://ranaroussi.github.io/yfinance/development/code.html

Jaypatel1511:
Rebased onto dev.

etbala:
> `get_history_metadata` is used in `yf.Ticker`, so the tests for it should go in `test_ticker.py`.

@Jaypatel1511 Just in case it wasn't clear, this is still an issue.

PR #2505, opened 2025-05-17
Restore requests support

ValueRaider:
Thanks to @deeleeramone https://github.com/ranaroussi/yfinance/issues/2486#issuecomment-2870628204, learnt that requests can work if it has a curl_adapter.

Confirmed working with rate limiter. Also works with requests_cache, but probably always misses cache because of crumb changing.

Issue #2486, opened 2025-05-11
Unable to use request_cache to cache the yfinance response


ronak66:
### Describe bug

I want to cache the yfinance response so I am using request_cache and overriding the session, But I am getting
Too Many Requests. Rate limited. Try after a while.

```python
requestCacheSession = requests_cache.CachedSession('yfinance.cache', expire_after=3600*2, impersonate="chrome")
stock = yf.Ticker(symbolNew, session=self.requestSession)
```

Can anyone help

yfinance version - 0.2.59
python version - 3.10

### Simple code that reproduces your problem

```python
requestCacheSession = requests_cache.CachedSession('yfinance.cache', expire_after=3600*2, impersonate="chrome")
stock = yf.Ticker(symbolNew, session=self.requestSession)
```

### Debug log from yf.enable_debug_mode()

Too Many Requests. Rate limited. Try after a while.

### `yfinance` version

0.2.59

ValueRaider:
Because then you are overriding using curl_cffi

requests_cache officially does not work with curl_cffi https://github.com/JWCook/requests-ratelimiter/issues/112

Will have to remove from docs

cybtow:
Hi, I solved it specifying manually the headers, here my pices of code:

`    class CachedLimiterSession(CacheMixin, LimiterMixin, Session):
        def __init__(self, *args, **kwargs):
            super().__init__(*args, **kwargs)
            self.headers.update({
                'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.124 Safari/537.36'
            })
`

`    def _get_internal_cache(self, max_requests: int = 2, in_seconds: int = 10):
        self.CachedLimiterSession(
            limiter=Limiter(RequestRate(max_requests, Duration.SECOND * in_seconds)),
            bucket_class=MemoryQueueBucket,
            backend=FileCache(os.path.join(self.cache_path, "_yfinance_file_cache")),
        )
`

`ticker = yf.Ticker(symbol, session=self.yfinance_session)`

ValueRaider:
@cybtow Did you really solve it? `_get_internal_cache` isn't returning anything.

cybtow:
Yes, I solved it, those are pieces of my code it's not completed, I tried to adapted here, let me copy and paste more, these pieces os code belongs an extensive class which I cannot paste here totally:

[...]

PR #2802, opened 2026-05-09
Make curl_cffi optional with fallback to requests (closes #2692)

### LINKED ISSUE #2692: [Feature] Allow falling back to normal curl
`curl_cffi` is pretty painful to build (last time I tried, I could fix building `curl-impersonate` executable, but the lib does not build correctly, and then python package executable fails to find symbols at runtime). Can a fallback to normal `curl` be added? `curl` exists as a given on most systems and does not require exotic SSL library and a package, which is absent in most downstream distributions.

dokson:
## Summary

Closes #2692.

`curl_cffi` is hard to build on some platforms (older macOS, exotic Linux distros, sandboxed environments). When it fails, yfinance becomes unusable. This PR keeps `curl_cffi` as the **preferred and default** backend but adds a graceful runtime fallback to plain `requests` so the import no longer brings the whole library down.

## Approach

- New `yfinance/_http.py` abstracts the HTTP backend:
  - If `curl_cffi` is importable → use it (current behavior preserved).
  - Else → fall back to `requests` with a realistic Chrome User-Agent and emit a `WARNING` log so the downgrade is explicit.
- All sites that built sessions or caught `curl_cffi.requests.exceptions.HTTPError` now go through `_http`.
- Small compat shim (`cookie_jar()`) papers over the only meaningful API difference between the two backends (`session.cookies.jar` vs. `session.cookies`).

## Why this differs from #2505

PR #2505 used `requests + curl_adapter` (which still goes through `curl_cffi` internally) and hit a `401 after ~10 requests` issue, presumably from `curl_adapter` quirks. This PR uses plain `requests` and does **not** suffer that failure mode — verified locally: 15/15 `quoteSummary` calls succeed with no rate-limiting.

## Security notes

- `requirements.txt` still pins `curl_cffi>=0.15` (CVE), unchanged.
- The plain-`requests` fallback uses the system OpenSSL stack and is not affected by the `curl_cffi` CVE; it has no broader attack surface than any other Python HTTP client.
- The fallback is logged so users running without TLS impersonation know about it.

## Behavior

```python
# curl_cffi installed (default): unchanged
>>> from yfinance._http import HAS_CURL_CFFI; HAS_CURL_CFFI
True

# curl_cffi unavailable
WARNING  curl_cffi not available; falling back to requests without browser TLS
         impersonation. Yahoo Finance may rate-limit or block this client.
         Install curl_cffi (>=0.15) for the supported configuration.
>>> yf.Ticker("AAPL").history(period="5d")        # works
>>> yf.Ticker("AAPL").get_info()                  # works
```

## Related

- #2505 — earlier attempt at the same goal via `curl_adapter`, blocked by the 401-after-10-requests issue mentioned above. If this PR lands, #2505 can be closed.
- #2486 — users asking to combine `requests_cache` / `requests_ratelimiter` with yfinance. Not directly closed here (those wrappers extend `requests.Session`, and this PR only switches to plain `requests` when `curl_cffi` is unavailable rather than on demand), but the fallback path now makes those integrations work end-to-end in environments where `curl_cffi` is absent. A follow-up exposing an explicit backend selector would fully address #2486.

## Out of scope

Moving `curl_cffi` from `install_requires` to `extras_require` is left to the maintainer — it is a default-install policy choice, not a code change. The fallback alone resolves the immediate "unusable on platform X" pain reported in #2692.

## Test plan

- [x] Default backend (`curl_cffi` present): `yf.Ticker("AAPL").history(period="5d")` and 5/5 `get_info()` calls succeed
- [x] Fallback (`curl_cffi` blocked at import): same workload + 15/15 `get_info()` calls succeed, no 401s
- [x] Cookie cache compat verified across both backends

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ Install `yfinance` from PYPI using `pip`:
 $ pip install yfinance
 ```
 
+To install without `curl_cffi` for requests fallback, see [Advanced ▸ Installation](https://ranaroussi.github.io/yfinance/advanced/install.html).
+
 ### [yfinance relies on the community to investigate bugs and contribute code. Here's how you can help.](https://github.com/ranaroussi/yfinance/blob/main/CONTRIBUTING.md)
 
 ---
--- a/doc/source/advanced/index.rst
+++ b/doc/source/advanced/index.rst
@@ -5,6 +5,7 @@ Advanced
 .. toctree::
    :maxdepth: 2
 
+   install
    logging
    config
    caching
--- a/doc/source/advanced/install.rst
+++ b/doc/source/advanced/install.rst
@@ -1,0 +1,9 @@
+Installation
+============
+
+To install without ``curl_cffi`` for requests fallback:
+
+.. code-block:: bash
+
+    curl -fsSL https://raw.githubusercontent.com/ranaroussi/yfinance/main/requirements.txt | grep -vi '^curl_cffi' | pip install -r /dev/stdin
+    pip install --no-deps yfinance
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -1,0 +1,75 @@
+"""Tests for the HTTP backend abstraction.
+
+Verifies the fallback to plain ``requests`` when ``curl_cffi`` is
+unavailable, and the ``YF_DISABLE_CURL_CFFI`` env-var opt-out.
+"""
+import importlib
+import os
+import sys
+import unittest
+
+
+def _reload_http(curl_cffi_available: bool, disable_env: bool):
+    """Reload ``yfinance._http`` simulating curl_cffi presence / opt-out."""
+    saved_modules = {k: sys.modules[k] for k in list(sys.modules) if k == "yfinance._http"}
+    saved_env = os.environ.get("YF_DISABLE_CURL_CFFI")
+    if disable_env:
+        os.environ["YF_DISABLE_CURL_CFFI"] = "1"
+    elif "YF_DISABLE_CURL_CFFI" in os.environ:
+        del os.environ["YF_DISABLE_CURL_CFFI"]
+
+    saved_curl = sys.modules.get("curl_cffi")
+    if not curl_cffi_available:
+        sys.modules["curl_cffi"] = None  # type: ignore[assignment]
+
+    try:
+        sys.modules.pop("yfinance._http", None)
+        return importlib.import_module("yfinance._http")
+    finally:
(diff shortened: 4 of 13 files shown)


ValueRaider:
What's the easiest way for me to run yfinance through requests even though curl_cffi works for me?

dokson:
@ValueRaider easiest path with the PR as-is: pass a stdlib `requests.Session` explicitly:

```python
import requests, yfinance as yf

s = requests.Session()
s.headers["User-Agent"] = (
    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
    "(KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36"
)
yf.Ticker("AAPL", session=s).history(period="5d")
```

`is_supported_session()` accepts either backend, so a `requests.Session` you provide is taken on as-is — no need to uninstall `curl_cffi`. The only catch is the User-Agent: when yfinance creates the session itself it sets a Chrome UA in `_http.new_session()`; if you bring your own session, you get whatever UA `requests` defaults to (`python-requests/X.Y`), which Yahoo will likely 403. The snippet above mirrors what `_http.new_session()` does internally.

ryanpetrello:
I'd like to chime in on this one - I'm a downstream package builder trying to build + package `yfinance`, but the `curl_cffi` requirement is too complex of a build requirement, especially given:

https://github.com/lexiforest/curl-impersonate?tab=readme-ov-file#how

> To make this work, curl was patched significantly to resemble a browser. Specifically, The modifications that were needed to make this work:
> * Compiling with BoringSSL, Google's TLS library, which is used by Chrome and Safari.
> * Modifying the way curl configures various TLS extensions and SSL options.
> * Adding support for new TLS extensions.
> * Changing the settings that curl uses for its HTTP/2 connections.
> * Running curl with some non-default flags, for example --ciphers, --curves and some -H headers.
> * The resulting curl looks, from a network perspective, identical to a real browser.

It this PR is viable, I'd love to see it be supported as an optional fallback 👍

ValueRaider:
Add a new section/page to the Advanced documentation that contains:
```
To install without curl_cffi for requests fallback:

curl -fsSL https://raw.githubusercontent.com/ranaroussi/yfinance/main/requirements.txt | grep -vi '^curl_cffi' | pip install -r /dev/stdin
pip install --no-deps yfinance
```

Then link to it from README, just under the install instruction.

PR #2804, opened 2026-05-09
Allow lang and region scoping for Ticker (closes #2582)

### LINKED ISSUE #2582: Possible to get company names in Chinese?
### Describe bug

I would like to know if it's possible to get the Chinese names and/or native names of companies other than English.

Let's take the ticker "1810.HK" as an example.  The following code will produce "XIAOMI-W" as the result:

`import yfinance as yf`
`print(yf.Ticker("1810.HK").info['longName'])`

However, if we go to https://hk.finance.yahoo.com/ and check out 1810.HK, we could see that the result is shown as
小米集團－Ｗ

I'm wondering if we could tell yfinance to connect to hk.finance.yahoo.com instead of finance.yahoo.com.

Thanks

`import yfinance as yf`
`print(yf.Ticker("1810.HK").info['longName'])`

N/A

### Bad data proof

_No response_

### `yfinance` version

0.2.65

### Python version

3.13.6

### Operating system

Win 11

dokson:
## Summary

Closes #2582.

`yf.Ticker("1810.HK").info["longName"]` always returned `"XIAOMI-W"` (U.S. English) because `lang`/`region` were not forwarded to Yahoo. Both `quoteSummary` (`v10`) and `v7/quote` actually accept BCP-47 `lang` and ISO 3166-1 alpha-2 `region` and return localized strings — verified live across Latin (with diacritics), CJK and Cyrillic scripts:

```python
>>> import yfinance as yf
>>> yf.config.locale.lang = "zh-Hant-HK"
>>> yf.config.locale.region = "HK"
>>> yf.Ticker("1810.HK").info["longName"]
'小米集團－Ｗ'

>>> yf.config.locale.lang = "ja-JP"
>>> yf.config.locale.region = "JP"
>>> yf.Ticker("7203.T").info["longName"]
'トヨタ自動車'

>>> yf.config.locale.lang = "ru-RU"
>>> yf.config.locale.region = "RU"
>>> yf.Ticker("GAZP.ME").info["longName"]
'Публичное акционерное общество Газпром'
```

## Design — locale on YfConfig (per @ValueRaider review)

Earlier revision exposed `lang` / `region` as `Ticker(symbol, lang=..., region=...)` constructor params. Reworked to live on `YfConfig.locale` instead:

- A typical caller wants the same locale for every Ticker in their session — config matches the actual usage shape.
- One config knob covers every v7/v10 endpoint, including the v1 calendar in `_fetch_history_metadata`, with no kwargs threaded through three layers.
- Consistent with `YfConfig.network.proxy` / `YfConfig.debug.logging` / `YfConfig.locale` is the same pattern.

## Changes

- `YfConfig.locale.lang` (default `"en-US"`) and `YfConfig.locale.region` (default `"US"`).
- `Quote._fetch` (v10/quoteSummary) and `Quote._fetch_additional_info` (v7/quote) read locale from `YfConfig` on every call.
- `TickerBase._fetch_history_metadata` (v1 calendar) reads the same locale, replacing a hardcoded `lang=en-US, region=US`.
- `v8/chart` ignores `lang` server-side, so we don't forward there — would only generate noise.
- Doc note in `yfinance.ticker_tickers.rst` and 6 tests covering defaults, four script families (Latin/diacritics, two CJK variants, Cyrillic), and the U.S.-listing-not-translated invariant.

## Notes

- Yahoo only translates fields for natively-listed companies. `Ticker("AAPL")` correctly still returns `"Apple Inc."` under `lang=ja-JP` — there is no Japanese translation of the U.S. listing. This is the documented behaviour of the `lang` parameter, and the tests assert it explicitly so future changes don't accidentally regress.
- `YfConfig.locale` is independent from the `MarketRegion` enum from #2801 (different endpoint, different vocabulary — `EUROPE`/`ASIA` aren't valid here, while `FR`/`DE`/`JP`/`HK` are). Not unifiable without forcing a wrong abstraction.

## Test plan

- [x] `python -m unittest tests.test_ticker_locale -v` — 6/6 pass
- [x] Default `Ticker("AAPL")` unchanged (English everywhere)
- [x] Five script families (Latin/diacritics, two CJK variants, Cyrillic) return native names
- [x] `ruff` / `pyright` clean

### DIFF
--- a/doc/source/advanced/config.rst
+++ b/doc/source/advanced/config.rst
@@ -54,3 +54,27 @@ Debug
   .. code-block:: python
 
      yf.config.debug.logging = True
+
+Locale
+------
+
+Localized fields (``longName``, ``shortName``, ...) follow
+``yf.config.locale``. The default is ``en-US`` / ``US``. Switch the Yahoo
+locale once at session start and every subsequent v7 / v10 endpoint call
+(``Ticker.info``, ``Ticker.fast_info``, ``Ticker.calendar``,
+``earnings_dates``, ...) inherits it:
+
+.. code-block:: python
+
+   import yfinance as yf
+   yf.config.locale.lang = "zh-Hant-HK"
+   yf.config.locale.region = "HK"
+   yf.Ticker("1810.HK").info["longName"]   # → '小米集團－Ｗ'
+
+   yf.config.locale.lang = "ja-JP"
+   yf.config.locale.region = "JP"
+   yf.Ticker("7203.T").info["longName"]    # → 'トヨタ自動車'
+
+Yahoo only returns translated values for tickers natively listed in that
+locale; ``Ticker("AAPL")`` keeps returning ``"Apple Inc."`` under
+``ja-JP``.
--- a/tests/test_ticker_locale.py
+++ b/tests/test_ticker_locale.py
@@ -1,0 +1,54 @@
+"""Verify YfConfig.locale switches the lang/region forwarded to Yahoo."""
+from tests.context import yfinance as yf
+
+import unittest
+
+
+class TestTickerLocale(unittest.TestCase):
+    def setUp(self):
+        self._backup_lang = yf.config.locale.lang
+        self._backup_region = yf.config.locale.region
+
+    def tearDown(self):
+        yf.config.locale.lang = self._backup_lang
+        yf.config.locale.region = self._backup_region
+
+    def test_default_locale_is_en_us(self):
+        self.assertEqual(yf.config.locale.lang, "en-US")
+        self.assertEqual(yf.config.locale.region, "US")
+
+    def test_default_returns_english(self):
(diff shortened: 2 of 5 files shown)


ValueRaider:
Instead, can we put lang and region into YfConfig? Then have them apply to all calls to v7 or v10?

dokson:
@ValueRaider done: moved `lang` / `region` onto `YfConfig.locale` and removed the per-Ticker constructor params. `Quote._fetch` (v10), `Quote._fetch_additional_info` (v7) and `_fetch_history_metadata` (v1 calendar) now all read from the same config knob. Force-pushed; tests + ruff + pyright green.

ValueRaider:
Move rst to `doc/source/advanced/config.rst`

dokson:
> Move rst to `doc/source/advanced/config.rst`

done!

PR #2805, opened 2026-05-09
Make yf.download() reentrant by removing shared module globals

dokson:
## Summary

`yf.download()` used module-level dicts in `yfinance.shared` (`_DFS`, `_ERRORS`, `_TRACEBACKS`, `_ISINS`) as scratch space. Concurrent calls from different threads trampled each other (`RuntimeError: dictionary changed size during iteration`). The previous fix wrapped the whole call in `shared._LOCK`, which fixed the crash but silently serialized parallel downloads — hiding the race instead of removing it, and defeating the parallelism users reach for `download()` to get.

This PR removes the shared mutable state instead of locking around it.

## Changes

- **`_DownloadCtx`** per call: holds `dfs / errors / tracebacks / isins / progress_bar` plus a fine-grained `lock` for thread workers. Concurrent calls each get their own instance.
- `_download_one`, `_download_one_threaded`, `_realign_dfs` take the ctx as first argument and write into it.
- The call-wide `shared._LOCK.acquire()` is dropped; the threaded-path polling loop reads `len(ctx.dfs)` under `ctx.lock`.
- **`PriceHistory._last_error`**: soft errors (delisted, missing tz, `auto_adjust`/`back_adjust` failure) are recorded on the instance instead of in `shared._ERRORS`. `_download_one` reads it after `Ticker.history()` returns and surfaces it in `ctx.errors`.
- `yfinance.shared` is left in place for import-compat with third-party code that may still import its dicts, but yfinance no longer reads or writes them.

## Drawbacks (intentional)

- `yfinance.shared._DFS / _ERRORS / _TRACEBACKS / _ISINS` are no longer populated. They are underscore-prefixed (private) so this is silent — any caller using them was already on internal API.
- The internal `_download_one(ctx, ticker, ...)` signature gains a leading `ctx` arg. The existing `tests/test_multi.py` monkeypatched the old signature; it has been updated. External code that monkeypatches these private functions will need the same trivial update.

## Test plan

- [x] `python -m unittest tests.test_multi tests.test_download_concurrency -v` — 4/4 pass
- [x] **TDD red→green**: `test_download_does_not_use_module_globals` pre-populates `shared._DFS` with a sentinel; on `dev` it gets wiped by `download()`'s reset, on this branch it survives. Verified red on `dev` (via `git stash` + run) and green here.
- [x] End-to-end behavior unchanged for both `threads=True` and `threads=False` paths against live tickers.
- [x] Mixed-success workload (`['AAPL', 'BADTICKER123']`) produces the same column layout and the same logged error as `dev`.

### DIFF
--- a/tests/test_download_concurrency.py
+++ b/tests/test_download_concurrency.py
@@ -1,0 +1,57 @@
+"""Concurrent ``yf.download()`` calls must not share scratch state.
+
+These tests assert that two concurrent ``download()`` calls each receive a
+DataFrame containing exactly the tickers they requested — no contamination,
+no exceptions — and that ``download()`` no longer relies on module-level
+dicts in ``yfinance.shared`` as scratch space.
+"""
+import unittest
+from concurrent.futures import ThreadPoolExecutor
+
+from tests.context import yfinance as yf
+
+
+class TestDownloadConcurrency(unittest.TestCase):
+
+    def _fetch(self, tickers):
+        return yf.download(
+            tickers, period="5d", interval="1d",
+            threads=False, progress=False, auto_adjust=False,
+        )
+
+    def test_concurrent_downloads_keep_results_separate(self):
+        chunk_a = ["AAPL", "MSFT", "GOOG"]
+        chunk_b = ["NVDA", "META", "AMZN"]
+
+        with ThreadPoolExecutor(max_workers=2) as ex:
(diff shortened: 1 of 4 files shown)


ValueRaider:
Why are you deleting comments?

dokson:
> Why are you deleting comments?

I didn't want to "spam" 😞

PR #2811, opened 2026-05-10
Fall back to chunked requests when single-URL fundamentals fetch times out

### LINKED ISSUE #2791: Balance sheet and cash flow methods return empty data frames
### Describe bug

When trying to get MSFT's balance sheet or cash flow using t.get_cashflow(), t.cashflow, t.get_balance_sheet, t.balance_sheet , they return empty data frames. Accessing for example the income statements using t.get_income_stmt() works fine. 

```python
import yfinance as yf
yf.config.debug.logging = True
t = yf.Ticker("MSFT")

print("get_cashflow:", t.get_cashflow())
print(".cashflow:", t.cashflow)
print("get_balance_sheet:", t.get_balance_sheet())
print(".balance_sheet:", t.balance_sheet)
```

DEBUG       url=https://query2.finance.yahoo.com/ws/fundamentals-timeseries/v1/finance/timeseries/MSFT?symbol=MSFT&type=annualForeignSales,annualDomesticSales,annualAdjustedGeographySegmentData,annualFreeCashFlow,ann...
DEBUG       params=None
DEBUG       Entering _get_cookie_and_crumb()
DEBUG        cookie_mode = 'basic'
DEBUG        Entering _get_cookie_and_crumb_basic()
DEBUG         Entering _get_cookie_basic()
DEBUG          Entering _load_cookie_curlCffi()
DEBUG          Exiting _load_cookie_curlCffi()
DEBUG          reusing persistent cookie
DEBUG         Exiting _get_cookie_basic()
DEBUG         Entering _get_crumb_basic()
DEBUG          Entering _get_cookie_basic()
DEBUG           reusing cookie
DEBUG          Exiting _get_cookie_basic()
DEBUG          crumb = 'gjmESxwovH0'
DEBUG         Exiting _get_crumb_basic()
DEBUG        Exiting _get_cookie_and_crumb_basic()
DEBUG       Exiting _get_cookie_and_crumb()
DEBUG    Exiting _fetch_time_series()
get_cashflow: Empty DataFrame
Columns: []
Index: []
.cashflow: Empty DataFrame
Columns: []
Index: []
DEBUG    Entering _fetch_time_series()
DEBUG     Entering get()
DEBUG      Entering _make_request()
DEBUG       url=https://query2.finance.yahoo.com/ws/fundamentals-timeseries/v1/finance/timeseries/MSFT?symbol=MSFT&type=annualTreasurySharesNumber,annualPreferredSharesNumber,annualOrdinarySharesNumber,annualShareIss...
DEBUG       params=None
DEBUG       Entering _get_cookie_and_crumb()
DEBUG        cookie_mode = 'basic'
DEBUG        Entering _get_cookie_and_crumb_basic()
DEBUG         Entering _get_cookie_basic()
DEBUG          reusing cookie
DEBUG         Exiting _get_cookie_basic()
DEBUG         Entering _get_crumb_basic()
DEBUG          reusing crumb
DEBUG         Exiting _get_crumb_basic()
DEBUG        Exiting _get_cookie_and_crumb_basic()
DEBUG       Exiting _get_cookie_and_crumb()
DEBUG    Exiting _fetch_time_series()
get_balance_sheet: Empty DataFrame
Columns: []
Index: []
.balance_sheet: Empty DataFrame
Columns: []
Index: []

### Bad data proof

_No response_

### `yfinance` version

1.3.0

### Python version

3.12.3

### Operating system

Windows 11

dokson:
## Summary

Closes #2791.

`Ticker.get_balance_sheet()` / `Ticker.get_cashflow()` silently returned empty DataFrames for users on WSL2 (and similarly restrictive networks: corporate proxies, some VPN exit nodes). The reporter showed a `curl_cffi` 30-second timeout with 0 bytes received — Yahoo was silently dropping the request, not yfinance discarding the response.

Root cause: `_get_financials_time_series` packs ~150 metric names into one `type=` query parameter. URL lengths:

| Statement | Total URL |
|---|---|
| Income | ~3093 chars |
| Cash flow | ~3819 chars |
| Balance sheet | ~4215 chars |

Above ~3.5 KB the request hits the WSL2 NAT MTU / proxy URL-length cap on the user's network and dies. Income (3093 chars) gets through; cash-flow and balance-sheet don't — exactly matching the bug report.

## Fix

Reactive fallback in `_get_financials_time_series`:

1. Try the existing single-URL fast path — no behaviour change for the 99% of users where it works.
2. On any failure (`Timeout`, `ConnectionError`, empty payload from Yahoo, …) fall back to chunked requests: split the `type=` list into 60-key chunks (~1.8 KB per URL) and merge the resulting `timeseries.result` arrays. Output shape is bit-identical.

Why reactive (not threshold-based proactive chunking):

- Don't pay the latency cost of multi-round-trip on networks where the single call works.
- The "what URL length is too long" answer is per-network — a static threshold either over-chunks (slowing the common case) or under-chunks (still fails on tighter networks). The reactive approach auto-discovers it.

## Test (TDD)

`test_balance_sheet_chunked_fallback_on_timeout` and `test_cash_flow_chunked_fallback_on_timeout` patch `YfData.cache_get` to:
- raise `curl_cffi.requests.exceptions.Timeout` on the first (long) URL — simulating the WSL2 silent drop;
- return canned per-chunk JSON on subsequent (shorter) URLs.

Both tests assert the call returns populated data and that the fallback path was actually exercised (≥ 3 cache_get calls — 1 timeout + ≥ 2 chunks). Verified red→green: removing the `try/except` block makes the tests fail with the same `Timeout` that bites real WSL2 users.

## Drawbacks

| | Cost |
|---|---|
| Latency on failing networks | First fetch eats one timeout (curl_cffi default 30s) before falling back. Subsequent fetches in the same session hit the cache. |
| API call volume | 2-3× requests **only** when the fast path failed. Networks where the single call works are unchanged. |
| Cache footprint | 2-3 cache rows instead of 1 per (ticker, timescale) on the slow path. Negligible. |
| Code | ~30 lines: a small helper + a `try/except` chunking loop. |

## Test plan

- [x] `python -m unittest tests.test_ticker.TestTickerMiscFinancials.test_balance_sheet_chunked_fallback_on_timeout tests.test_ticker.TestTickerMiscFinancials.test_cash_flow_chunked_fallback_on_timeout` — green
- [x] `python -m unittest tests.test_ticker.TestTickerMiscFinancials.test_balance_sheet` — existing live-data test still green (fast path unchanged)
- [x] `ruff check` — clean
- [x] `pyright . --level error` — 0 errors

### DIFF
--- a/yfinance/scrapers/fundamentals.py
+++ b/yfinance/scrapers/fundamentals.py
@@ -108,25 +108,47 @@ class Financials:
                 raise
             pass
 
+    # Tuned so the resulting URL stays under ~2KB even for the longest "annual"
+    # prefix, which keeps it below the practical limits of typical NAT / proxy
+    # paths (notably WSL2, which silently drops the long single-shot URL).
+    _CHUNK_KEYS = 60
+
     def _get_financials_time_series(self, timescale, keys: list) -> pd.DataFrame:
         timescale_translation = {"yearly": "annual", "quarterly": "quarterly", "trailing": "trailing"}
         timescale = timescale_translation[timescale]
 
-        # Step 2: construct url:
-        ts_url_base = f"https://query2.finance.yahoo.com/ws/fundamentals-timeseries/v1/finance/timeseries/{self._symbol}?symbol={self._symbol}"
-        url = ts_url_base + "&type=" + ",".join([timescale + k for k in keys])
         # Yahoo returns maximum 4 years or 5 quarters, regardless of start_dt:
         start_dt = datetime.datetime(2016, 12, 31)
         end = pd.Timestamp.now('UTC').ceil("D")
-        url += f"&period1={int(start_dt.timestamp())}&period2={int(end.timestamp())}"
+        period_qs = f"&period1={int(start_dt.timestamp())}&period2={int(end.timestamp())}"
 
-        # Step 3: fetch and reshape data
-        json_str = self._data.cache_get(url=url).text
-        json_data = json.loads(json_str)
-        data_raw = json_data["timeseries"]["result"]
-        # data_raw = [v for v in data_raw if len(v) > 1] # Discard keys with no data
+        ts_url_base = f"https://query2.finance.yahoo.com/ws/fundamentals-timeseries/v1/finance/timeseries/{self._symbol}?symbol={self._symbol}"
+        full_url = ts_url_base + "&type=" + ",".join([timescale + k for k in keys]) + period_qs
+
+        # Fast path: single long URL. Falls back to chunked requests if it
+        # fails (silent drop on WSL2 NAT / restrictive proxies). Sticky so a
+        # loop over tickers doesn't eat one timeout per ticker. If the chunked
+        # fallback also fails, URL length isn't the problem — revert the flag
+        # and re-raise so the next call retries the fast path.
+        if self._data.fundamentals_use_chunked:
+            data_raw = self._fetch_fundamentals_chunked(ts_url_base, timescale, keys, period_qs)
+        else:
+            try:
+                data_raw = self._fetch_fundamentals_payload(full_url)
+            except Exception as e:
+                utils.get_yf_logger().debug(
+                    f"{self._symbol}: single-URL fundamentals fetch failed ({type(e).__name__}); "
+                    f"falling back to chunked requests for this and subsequent fetches"
+                )
+                self._data.fundamentals_use_chunked = True
+                try:
+                    data_raw = self._fetch_fundamentals_chunked(ts_url_base, timescale, keys, period_qs)
+                except Exception:
+                    self._data.fundamentals_use_chunked = False
+                    raise
+
         for d in data_raw:
-            del d["meta"]
+            d.pop("meta", None)
 
         # Now reshape data into a table:
         # Step 1: get columns and index:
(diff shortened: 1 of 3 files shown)


ValueRaider:
I agree with strategy. But to minimise requests, when first detect timeout, store somewhere so then subsequent fetches of financials are chunked (imagine user looping over tickers).

If you're lazy, shove it into `YfData`. If you want to do properly, create a new singleton `YfState` class.

dokson:
> I agree with strategy. But to minimise requests, when first detect timeout, store somewhere so then subsequent fetches of financials are chunked (imagine user looping over tickers).
> 
> If you're lazy, shove it into `YfData`. If you want to do properly, create a new singleton `YfState` class.

YfState ready 😄

yfinance/scrapers/fundamentals.py
```diff
@@ -108,25 +109,41 @@ def _create_financials_table(self, name, timescale):
...
+        else:
+            try:
+                data_raw = self._fetch_fundamentals_payload(full_url)
+            except Exception as e:
+                utils.get_yf_logger().debug(
+                    f"{self._symbol}: single-URL fundamentals fetch failed ({type(e).__name__}); "
+                    f"falling back to chunked requests for this and subsequent fetches"
+                )
+                YfState.fundamentals_use_chunked = True
+                data_raw = self._fetch_fundamentals_chunked(ts_url_base, timescale, keys, period_qs)
```
ValueRaider: Can we wrap this in another exception? If fetch fails again, then problem isn't length (and revert `fundamentals_use_chunked` to False).

Might need to restructure the whole block.

ValueRaider (changes requested):

yfinance/scrapers/fundamentals.py
dokson: Restructured: the chunked fallback is now wrapped in its own try/except — if it also raises, the flag reverts to False and the exception propagates, so the next call retries the fast path instead of being stuck on chunked forever. Also folded the flag onto YfData (per your earlier comment's first option) and deleted the dedicated state module.

ValueRaider:
git squash

PR #2810, opened 2026-05-10
Simplify phantom-dividend repair branch + drive-by typo/lint fixes

dokson:
## Summary

Originally proposed extracting a ticker-allowlist into a separate module. Per maintainer feedback on the review (#discussion_r3244327805), the allowlist was leakage from a custom debug build and should not be in `main` at all — phantom-dividend detections are safe to approve unconditionally in that branch. This PR now just applies that simplification plus two drive-by fixes that were already in the diff.

## Changes

- `yfinance/scrapers/history.py`
  - `_fix_bad_div_adjust`: replace `if c == 'phantom' and self.ticker in ['KAP.IL', 'SAND']:` with `if c == 'phantom':`. Removes a hardcoded ticker dependency from the repair kernel.
  - Drive-by: split a semicolon-joined statement (around line 2735) that was tripping ruff E702 in CI.
- `yfinance/domain/industry.py`
  - Drive-by typo: `compnaies_column` / `compnaies_values` → `companies_column` / `companies_values` (8 occurrences in `_parse_top_performing_companies` and `_parse_top_growth_companies`). Local-variable rename, no behavioural impact.

## Behaviour

`_fix_bad_div_adjust` now approves *all* phantom-dividend detections, not just those for `KAP.IL` / `SAND`. Per maintainer this is the intended behaviour.

## Test plan

- [x] `python -m unittest tests.test_price_repair` — 13 tests pass, 2 skipped
- [x] `ruff check yfinance/` — clean

PR #2808, opened 2026-05-10
Add opt-in polars backend across the public yfinance API

dokson:
## Summary

Closes #1868.

Adds **opt-in polars support** across every yfinance API that returns a DataFrame. Pandas remains the default backend; users opt into polars with one line:

```python
import yfinance as yf
yf.config.dataframe.backend = 'polars'
yf.Ticker('MSFT').history(period='1mo')   # → polars.DataFrame
```

Polars is an optional install: `pip install yfinance[polars]`.

## Why this differs from #2782

PR #2782 is a v2.0 major-bump migration that makes polars the default and breaks every existing pandas caller (`download()` shape changes from MultiIndex to long-form, `history()` returns polars unless `as_pandas=True`). It also bumps the minimum Python to 3.10 and migrates packaging to `pyproject.toml`+`uv`.

This PR is the **opposite trade-off**: zero breaking changes, no version bump, no Python drop, no packaging churn. Pandas users (the vast majority) see exactly the current behaviour. Polars users get polars output everywhere via a single config flag.

| | This PR | #2782 |
|---|---|---|
| Default backend | pandas (no break) | polars (breaks every caller) |
| Polars dependency | extras_require['polars'] | hard dep |
| Min Python | unchanged | bumps to 3.10 |
| Packaging | unchanged | setup.py → pyproject.toml + uv |
| Repair engine | pandas bridge (same as #2782) | pandas bridge |
| Output coverage | every public DataFrame API | every public DataFrame API |

### Single source of truth

- `YfConfig.dataframe.backend` — `'pandas'` (default) or `'polars'`. Raises `ValueError` on unknown values.
- `utils.current_backend()` — used by every dispatch site.
- `utils.empty_backend_df()`, `df_to_backend()`, `series_to_backend()`, `df_from_records()` — central dispatch helpers.

### Native polars helpers

New module `yfinance/polars_helpers.py` with one-to-one polars counterparts of the pandas pipeline helpers:

- `parse_quotes`, `parse_actions` (chart payload → polars).
- `set_df_tz`, `fix_Yahoo_dst_issue`, `fix_Yahoo_returning_prepost_unrequested`.
- `auto_adjust`, `back_adjust`.
- `resample` — uses `group_by_dynamic` + `dt.truncate` (the recipe FBruzzesi pointed at after narwhals#2532). Covers 1wk/5d/1mo/3mo with the same ytd-aware origin handling as the pandas implementation.
[...]

PR #2782, opened 2026-04-24
feat: migrate to polars + uv (v2.0.0) — RFC / Discussion

buzzvolt:
## RFC: pandas → Polars migration + uv tooling (v2.0.0)

This is an **invitation to discuss**, not a request to merge immediately.

The full rationale, every breaking change, and side-by-side code comparisons
are documented in [`docs/migration-v2-polars.md`](docs/migration-v2-polars.md)
(rendered below on GitHub).

### Why open this PR?

The two most consistent pain points in yfinance's history have been:

1. **The MultiIndex column output of `download()`** — there is a dedicated
   doc page for it that simply points to a Stack Overflow answer. The
   community's own workaround (`df.stack(level=1).reset_index()`) produces
   exactly the long-form shape this PR returns natively.

2. **pandas as a hard dependency** — polars is faster, has an immutable
   DataFrame model that eliminates a whole class of mutation bugs present
   in the current codebase (including `df._consolidate()` being called in
   production), and requires no index concept (dates are explicit columns).

### What this PR does

| Area | Change |
|---|---|
| **Packaging** | `setup.py` → `pyproject.toml` + `uv` |
| **CI** | Re-enables the disabled `pytest.yml`; all workflows use `uv run` |
| **`download()`** | Returns long-form `pl.DataFrame` with `"Ticker"` column instead of MultiIndex |
| **`history()`** | Returns `pl.DataFrame`; `as_pandas=True` bridge for backward compat |
| **All scrapers** | Fully migrated to polars; no pandas import in normal code path |
| **Price repair** | Pragmatic pandas bridge retained internally; repair still works |
| **Tests** | 42/46 pass; 4 failures are pre-existing (permissions, live data counts) |
| **Docs** | 1446-line migration guide in `docs/migration-v2-polars.md` |

### What stays the same

- All function signatures (parameters, defaults, behaviour)
- `actions=True` default — Dividends/Stock Splits columns present as before
- `repair=True` — works when `pandas[pandas]` extra installed
- `as_pandas=True` on `history()` — returns exact v1 DataFrame shape

### Questions for the maintainer / community

- Is a long-form `download()` output something the community would embrace?
- Should the pandas compat bridge be a first-class supported path or temporary?
- Would you prefer the repair engine be fully native polars before merging?
- Any concerns about dropping Python 3.6–3.9 support (now `>=3.10`)?

Happy to split this into smaller incremental PRs (e.g. just the uv/packaging
change, or just the `download()` shape change) if that makes review easier.

PR #2817, opened 2026-05-15
Add opt-in polars output via pandas pipeline + boundary adapter

### LINKED ISSUE #1868: Add support for python-polars
Besides the use of `pandas`, it would be great to add support for the `polars` Python interface: https://github.com/pola-rs/polars.
Polars has very similar functionalities to Pandas, but uses a Rust backend that significantly speeds up operations compared to Pandas.

dokson:
Alternative to #2782 and #2808 — closes #1868.

## Summary

Opt-in polars output without doubling the codebase. One line of user code:

```python
yf.config.dataframe.backend = 'polars'
yf.Ticker('MSFT').history(period='1mo')   # → polars.DataFrame
```

Pandas remains the default. Polars is a single optional install: `pip install yfinance[polars]`.

## Why this differs from #2808

@ValueRaider's review on #2808 pointed out the core problem:

> The main problem with this, is now we have to maintain Polars for no performance improvement — yfinance tables aren't big enough for choice to matter.

This PR takes that feedback as the design constraint. Instead of porting every scraper into a native polars implementation (#2808: ~5000 lines, including a 2272-line `polars_repair.py`), the internal pipelines stay 100% pandas and polars support is a **thin output adapter** applied at every public-API boundary. There is one code path internally; polars never participates in transformations.

| | This PR | #2808 |
|---|---|---|
| LoC added | ~700 | ~5000 |
| Polars-native repair engine | none (pandas + convert) | 2272 lines |
| Polars-native scrapers | none | every scraper |
| Default backend | pandas (no break) | pandas (no break) |
| Polars dependency | extras_require | extras_require |
| Performance | unchanged | unchanged |
| Maintenance surface | +1 adapter file, ~700 lines | duplicate every transform |

## Why this is not a performance play

It isn't. yfinance frames are small enough that pandas vs polars makes no practical difference — exactly @ValueRaider's point. The value here is **convenience**: users whose downstream pipelines are polars get polars frames straight from yfinance instead of threading `pl.from_pandas(...)` after every call. That's it.

## No breaking changes

- Default backend is `pandas`. Existing users see no behavioural difference.
- `df_to_backend(df)` is identity (`assertIs`) when backend == `pandas`. Unit-tested.
- Public-facing return annotations are widened from `pd.DataFrame` to `DataFrameLike = Union[pd.DataFrame, "pl.DataFrame"]`. Pandas remains a valid subtype, so all existing pandas-typed callers continue to type-check. Polars is imported only under `TYPE_CHECKING`, so users without `polars` installed see no import error.
- Caches keep frames in pandas; conversion happens on each public read, so switching backend mid-session is honoured immediately by every subsequent call.

## Surface covered

- `Ticker.history()` (4 return paths) and `yf.download()`.
- All 40+ `Ticker.X` @property accessors (via `base.py get_X()` wraps).
- All `Ticker.get_*` methods returning DataFrame/Series: dividends, splits, capital_gains, actions, shares, shares_full, earnings_dates, earnings, financial statements, recommendations, upgrades_downgrades, holders ×6, valuation_measures, sustainability, analysis getters ×6.
- `Ticker.option_chain()` (calls + puts).
- `Ticker.funds_data.{fund_operations, top_holdings, equity_holdings, bond_holdings}`.
- `Lookup.*`, `Calendars.*`, `Sector.industries`, `Industry.top_*`, `Domain.top_companies`.

Internal scraper `_quote.X / _holders.X / _analysis.X / _fundamentals.X` attributes are intentionally not wrapped — they are private (`_`-prefixed) and only used by `base.py`, which wraps at the public boundary.

## Tests

`tests/test_dataframe_backend.py` — 37 tests, **no network required**:

- Config validation (default / accepted / unknown / no-corruption-on-failure).
- `df_to_backend` semantics (passthrough on pandas, polars conversion, named-index promotion, override, RangeIndex drop, empty).
- `series_to_backend` semantics.
- Lookup parity on both backends + empty result.
- All 17 wrapped base.py getters return the right backend type, plus `as_dict=True` always returns a plain dict.
- Backend switch invariant: pandas → polars → pandas in sequence returns the matching type each time.
- Calendars `_to_backend` on both backends.
- Sector.industries property switches with backend.

```
$ python -m unittest tests.test_dataframe_backend tests.test_utils
Ran 58 tests in 0.290s
OK (expected failures=1)
```

Ruff clean for every file touched.

## Out of scope

- `download()` long-form polars shape: kept as MultiIndex pandas converted via reset_index. Native polars long-form is a separate opinionated decision.
- Price-repair engine stays pandas-internal — it's a numerical kernel with `scipy.ndimage`, not idiomatic DataFrame ops. The pandas frame is converted at the same `history()` boundary.

## Test plan

- [x] `python -m unittest tests.test_dataframe_backend tests.test_utils` — 58 pass
- [x] `ruff check` clean on every changed file (the lone remaining E702 in `history.py:2734` is pre-existing and addressed by #2810)
- [x] `pyright` — no new errors on the changed files

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,7 @@ setup(
     extras_require={
         'nospam': ['requests_cache>=1.0', 'requests_ratelimiter>=0.3.1'],
         'repair': ['scipy>=1.6.3'],
+        'polars': ['polars>=1.0'],
     },
     # Include protobuf files for websocket support
     package_data={
--- a/tests/test_dataframe_backend.py
+++ b/tests/test_dataframe_backend.py
@@ -1,0 +1,378 @@
+"""Tests for the dataframe backend dispatch.
+
+Covers ``YfConfig.dataframe.backend`` validation and parity of every
+wrapped boundary in the codebase — using mocked-data unit tests so the
+suite stays network-independent.
+"""
+from __future__ import annotations
+
+import unittest
+from unittest.mock import MagicMock
+
+import pandas as pd
+
+from yfinance._backend import df_to_backend, series_to_backend
+from yfinance.config import YfConfig
+from yfinance.lookup import Lookup
+
+
+_LOOKUP_RESPONSE = {
+    "finance": {
+        "result": [
+            {
+                "documents": [
+                    {"symbol": "AAPL", "longName": "Apple Inc.", "quoteType": "EQUITY"},
+                    {"symbol": "MSFT", "longName": "Microsoft", "quoteType": "EQUITY"},
+                ]
+            }
+        ]
+    }
+}
+
+
+def _polars_df_type():
+    import polars as pl
+    return pl.DataFrame
+
+
+# ---------------------------------------------------------------------------
+# YfConfig.dataframe.backend validation
+# ---------------------------------------------------------------------------
+class TestDataframeBackendConfig(unittest.TestCase):
+
+    def tearDown(self):
+        YfConfig.dataframe.backend = "pandas"
+
+    def test_default_backend_is_pandas(self):
+        self.assertEqual(YfConfig.dataframe.backend, "pandas")
+
+    def test_polars_backend_accepted(self):
+        YfConfig.dataframe.backend = "polars"
+        self.assertEqual(YfConfig.dataframe.backend, "polars")
+
+    def test_pandas_backend_accepted(self):
+        YfConfig.dataframe.backend = "polars"
+        YfConfig.dataframe.backend = "pandas"
+        self.assertEqual(YfConfig.dataframe.backend, "pandas")
+
+    def test_unknown_backend_raises(self):
+        with self.assertRaises(ValueError):
+            YfConfig.dataframe.backend = "modin"
+
+    def test_unknown_backend_does_not_corrupt_state(self):
+        try:
+            YfConfig.dataframe.backend = "modin"
+        except ValueError:
+            pass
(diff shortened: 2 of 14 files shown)


ValueRaider:
@buzzvolt @edoaltamura Does this work for you? It still uses Pandas internally but optionally converts to Polars at very end. Much easier than porting all the internal logic to Polars.

edoaltamura:
That'd be a useful functionality, thanks!

dokson:
great @edoaltamura!!! 

I'm very happy that you appreciate this and that it can be useful to you

buzzvolt:
> @buzzvolt @edoaltamura Does this work for you? It still uses Pandas internally but optionally converts to Polars at very end. Much easier than porting all the internal logic to Polars.

One of the reasons why I chose to go with polars is that as I had put it in https://github.com/ranaroussi/yfinance/discussions/2783 and [docs/migration-v2-polars.md](https://github.com/bazaarbuzz/yfinance/blob/68972af2b1385f8204c060ebd475fc0cd018e275/docs/migration-v2-polars.md) obviously for speed especially for group-by and to avoid things of multi-index. 

Instead of viewing it as a strict "either/or", many data teams are adopting hybrid workflows. You can do heavy data ingestion, filtering, and feature engineering in Polars, and then convert it into a Pandas object at the boundaries of your Machine Learning pipelines.

Issue #2327, opened 2025-02-24
Yahoo data changed from EST to UTC?


man-chan:
I'm not sure is my bug or Yahoo or yFinance. Since I didn't log the command I used, so I guess the command should look similar to the following. Not sure the change in Yahoo data or yFinance.

This week, I got saved 2025-02-14 data as follow. I am using yFinance 0.2.54 & python 3.9.
```python
yf.download('SPY', '2025-02-08', '2025-02-22', interval='1h', progress=False)

2025-02-14,09:00:00,609.31,610.5,609.22,610.26,0
2025-02-14,10:00:00,610.33,610.51,609.41,609.74,0
2025-02-14,11:00:00,609.73,609.74,608.67,608.73,0
2025-02-14,12:00:00,608.73,609.3,608.6,609.1,0
2025-02-14,13:00:00,609.11,611.0,608.76,609.39,0
2025-02-14,14:00:00,609.45,610.0,609.02,609.845,0
2025-02-14,14:30:00,609.9400024414062,610.989990234375,609.760009765625,610.239990234375,5214577
2025-02-14,15:30:00,610.239990234375,610.8499755859375,609.4000244140625,609.989990234375,3910433
2025-02-14,16:30:00,610.010009765625,610.2949829101562,609.27001953125,609.3201293945312,2764489
2025-02-14,17:30:00,609.3499755859375,610.0900268554688,609.0700073242188,609.9298706054688,1443250
2025-02-14,18:30:00,609.9299926757812,610.7506713867188,609.8800048828125,610.1300048828125,2059667
2025-02-14,19:30:00,610.1500244140625,610.3024291992188,609.5900268554688,609.6699829101562,2430485
2025-02-14,20:30:00,609.6599731445312,610.2899780273438,609.5449829101562,609.7000122070312,5528441
2025-02-14,21:00:00,609.69,610.0901,609.42,609.928,0
2025-02-14,22:00:00,609.93,609.95,609.66,609.83,0
2025-02-14,23:00:00,609.78,609.88,609.47,609.625,0
2025-02-18,09:00:00,610.77,611.91,610.77,611.51,0
```

Last week, I got the saved data, I was using 0.1.69 (worked like this for years).
```python
yf.download('SPY', '2025-02-01', '2025-02-15', interval='1h', progress=False) 

2025-02-14,04:00:00,609.31,610.5,609.22,610.26,610.26,0
[...]

PR #2818, opened 2026-05-17
Add tz= kwarg to history() and download()

dokson:
## Summary
Lets callers choose the timezone of the returned DataFrame's index, instead of always getting the exchange tz on `Ticker.history()` or UTC on `download()`.

```python
yf.Ticker('SPY').history(period='5d', interval='1h', tz='UTC')
yf.download('IBM', period='1mo', interval='1d', tz='Europe/Rome')
```

Accepts an IANA name (`"UTC"`, `"Europe/Rome"`) or any `tzinfo` (e.g. `zoneinfo.ZoneInfo`). Bad values propagate from pandas — no silent fallback.

### Design notes
- Input parsing (naive `start`/`end` interpreted in exchange tz) is **decoupled** from output formatting (`tz=`). The exchange tz is still used to interpret user-provided dates; only the output index is converted.
- The existing internal `tz` local in `PriceHistory.history` (exchange tz) is preserved as-is; the new kwarg is captured into `_output_tz` at the top of the function to avoid shadowing.
- On `download()`, `tz=` overrides the `ignore_tz` auto-default (so daily intervals can also opt into a specific tz). Passing both `tz=...` and `ignore_tz=True` raises `ValueError` rather than silently picking one.
- A `WARNING` is logged when the user converts daily/weekly bars to a non-exchange tz, since midnight-at-exchange bars will shift their calendar date.

Addresses the follow-up request in #2327.

## Test plan
- [x] `tz=None` is a no-op (regression guard)
- [x] `tz="UTC"` on intraday converts the index, same moments in time
- [x] `tz=ZoneInfo("Europe/Rome")` works equally
- [x] Invalid tz string raises
- [x] Daily interval with `tz=` emits warning
- [x] `download(tz="UTC")` on daily interval overrides `ignore_tz` auto-default
- [x] `download(ignore_tz=True, tz="UTC")` raises `ValueError`
- [x] `ruff check` clean

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -298,6 +298,64 @@ class TestTickerHistory(unittest.TestCase):
         self.assertIsInstance(data, pd.DataFrame, "data has wrong type")
         self.assertFalse(data.empty, "data is empty")
 
+    def test_history_output_tz_none_is_noop(self):
+        # Regression guard: tz=None must not touch the index.
+        baseline = self.ticker.history(period="5d", interval="1h")
+        same = self.ticker.history(period="5d", interval="1h", tz=None)
+        self.assertEqual(str(baseline.index.tz), str(same.index.tz))
+
+    def test_history_output_tz_daily_warns(self):
+        # Daily/weekly bars shift calendar dates when converted -> must warn.
+        with self.assertLogs("yfinance", level="WARNING") as cm:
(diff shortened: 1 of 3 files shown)


ValueRaider:
Are you using sloppy AI? It isn't thinking, it's just blasting code.

Thought exercise to figure out the problem.

dokson:
sorry @ValueRaider...misinterpreted the issue

Issue #1308, opened 2023-01-16
Get info of all UPCOMING corporate events for given ticker?


Robczan:
Is it possible to obtain all announced upcoming corporate actions for given ticker (dividends, splits, mergers, etc)? They are listed on

**https://finance.yahoo.com/calendar/**

ValueRaider:
Probably possible to implement scraping, but not possible with `yfinance` right now. Same request as #1223

VictorGob:
Is it not possible because it hasn't been implemented or for some legal reason?

ValueRaider:
Not implemented

PR #2819, opened 2026-05-17
Expose upcoming Split Date/Ratio/Type and Is Earnings Date Estimate in calendar

dokson:
## Summary
Surfaces upcoming-event fields that Yahoo's quoteSummary already returns but yfinance was dropping:

- **`Split Date` / `Split Ratio` / `Split Type`** — from the previously-unwhitelisted `corporateActions` module. Yahoo populates it whenever a forward/reverse split is announced. The legacy `calendarEvents.splitDate` is kept as a fallback. Date is resolved in **UTC** so the calendar day doesn't shift on machines in negative-offset zones.
- **`Is Earnings Date Estimate`** — boolean from `calendarEvents.earnings`. Lets callers tell a confirmed earnings date from an estimate.

No new HTTP call: `_fetch_calendar` requests both modules in one `quoteSummary` round-trip. Picks off the splits part of #1308. Mergers stay out of scope — Yahoo doesn't surface them via `quoteSummary`.

`earningsCallDate` was investigated but is intentionally **not** surfaced: in observed payloads it can point to a past quarter's call rather than the next one (e.g. AAPL returns 2026-04-30 while `Earnings Date` is 2026-07-30), so exposing it without disambiguation would mislead callers.

## Verified on real tickers

| Ticker | Split Date | Ratio | Type | Is Earnings Date Estimate |
|---|---|---|---|---|
| CCWYF | 2026-05-19 | 1:20 | Reverse stock split | – |
| BYOGF | 2026-05-20 | 1:30 | Reverse stock split | True |
| AAPL | – | – | – | True |
| NVDA | – | – | – | False |
| MSFT | – | – | – | True |

Verified that already-completed splits (GMEX/GREH/TLIH/GPUSF, split 2026-05-01) do **not** appear — Yahoo filters server-side, no client-side guard needed.

## Test plan
- [x] Existing `test_calendar` still passes (live, GOOGL)
- [x] Mocked `test_calendar_split_from_corporate_actions` — injects a fake SPLIT payload and asserts Date/Ratio/Type populate (no live dependency)
- [x] Mocked `test_calendar_no_split_when_corporate_actions_empty` — asserts no Split* keys leak when actions list is empty
- [x] Sanity-checked across CCWYF, BYOGF (with upcoming splits) and AAPL/NVDA/MSFT (no split — guard kicks in, keys absent)
- [x] `ruff check` clean

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -872,8 +872,52 @@ class TestTickerMiscFinancials(unittest.TestCase):
         if self.ticker.ticker != "GOOGL":
             self.assertIn("Dividend Date", data.keys(), "data missing expected key")
         # ex-dividend date is not always available
+        # isEarningsDateEstimate is present in the raw payload whenever
+        # earnings info is known; surface it when Yahoo provides it.
+        if "Is Earnings Date Estimate" in data:
+            self.assertIsInstance(data["Is Earnings Date Estimate"], bool)
         data_cached = self.ticker.calendar
         self.assertIs(data, data_cached, "data not cached")
+
+    def test_calendar_split_from_corporate_actions(self):
+        """Mock the quoteSummary payload so the test is independent of any
+        live ticker happening to have an announced split today."""
+        from unittest.mock import patch
+        # 1779163200000 ms = 2026-05-19 04:00 UTC -> calendar date 2026-05-19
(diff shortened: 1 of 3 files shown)


ValueRaider:
No more AI.

PR #2821, opened 2026-05-17
Drop frozendict hard dependency in favour of an internal fallback

### LINKED ISSUE #1752: replace frozendict with immutabledict
### Describe bug

Hi, Gentoo Linux is going to remove the python package frozendict:
https://bugs.gentoo.org/918899
They write "Originally added for net-im/synapse, which switched to dev-python/immutabledict since.  Hacky C extension that supports up to Python 3.10". "Removal on 2023-12-30."

The project site of immutabledict says it is a fork of frozendict, written in pure python: https://github.com/corenting/immutabledict/
The package did not have additional dependencies as I installed it.

I tried the following change in yfinance/data.py and yfinance seemed to work (the script I wrote with it did everything I expected from it):
````python
#from frozendict import frozendict
from immutabledict import immutabledict
````

see bug description

### Debug log

There are no errors:
DEBUG    get_raw_json(): https://query2.finance.yahoo.com/v10/finance/quoteSummary/AMZ.DE
DEBUG    Entering get()
DEBUG     url=https://query2.finance.yahoo.com/v10/finance/quoteSummary/AMZ.DE
DEBUG     params={'modules': 'financialData,quoteType,defaultKeyStatistics,assetProfile,summaryDetail', 'ssl': 'true'}
DEBUG     Entering _get_cookie_and_crumb()
DEBUG      cookie_mode = 'basic'
DEBUG      Entering _get_cookie_and_crumb_basic()
DEBUG       loaded persistent cookie
DEBUG       reusing cookie
DEBUG       crumb = 'SnU9.2cbXIN'
DEBUG      Exiting _get_cookie_and_crumb_basic()
DEBUG     Exiting _get_cookie_and_crumb()
DEBUG    Exiting get()

### Bad data proof

_No response_

### `yfinance` version

yfinance-0.2.32

### Python version

Python 3.11.6

### Operating system

Linux

### PR DESCRIPTION
**Pull request by dokson**
## Summary
`frozendict` is currently a hard dependency of yfinance but is used in exactly one place — `lru_cache_freezeargs` in `data.py`, to make dict arguments hashable so `functools.lru_cache` can key on them.

This PR makes the dependency optional (PEP 814 fallback pattern): if `frozendict` is installed it is used as-is, otherwise a small pure-Python equivalent defined in `yfinance/utils.py` is used. `frozendict>=2.3.4` is removed from `requirements.txt` and `setup.py`.

```python
try:
    from frozendict import frozendict
except ImportError:
    class frozendict(dict):
        __slots__ = ()
        def __hash__(self):
            return hash(frozenset(self.items()))
        def __setitem__(self, *a, **kw):
            raise TypeError(f"'{type(self).__name__}' object doesn't support item assignment")
        def __delitem__(self, *a, **kw):
            raise TypeError(f"'{type(self).__name__}' object doesn't support item deletion")
        def _readonly(self, *a, **kw):
            raise AttributeError(f"'{type(self).__name__}' object is read-only")
        pop = popitem = clear = update = setdefault = _readonly
```

The fallback mirrors the upstream class name (`frozendict`, lowercase) and exception types (`TypeError` on `__setitem__`/`__delitem__`, `AttributeError` on `pop`/`popitem`/`clear`/`update`/`setdefault`) so callers see identical behavior either way.

## Why
- The `frozendict` package ships a C extension, which downstream distros have started dropping (Gentoo removed it in 2023).
- yfinance's actual usage is trivial and well within stdlib reach — there is no reason to *require* an external package for it.
- One fewer hard dependency to install, build, audit and version-pin; users who already have `frozendict` keep using it transparently.

## Verified
- 4 unit tests in `tests/test_data.py` cover the dict-like behavior, hashability (including order-insensitivity), mutation rejection, and the end-to-end `lru_cache_freezeargs` integration. Tests pass both with `frozendict` installed and against the pure-Python fallback.
- Live smoke test: `Ticker('AAPL').history(period='5d')` works and the second call is served from `cache_get`'s lru cache, equal to the first.
- `ruff check` clean.

Closes #1752.

## Test plan
- [x] `tests/test_data.py::TestFrozenDict` — 4/4 pass
- [x] Existing tests unaffected (only data.py imports change)
- [x] Live smoke test against Yahoo

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,6 @@ requests>=2.31
 multitasking>=0.0.7
 platformdirs>=2.0.0
 pytz>=2022.5
-frozendict>=2.3.4
 beautifulsoup4>=4.11.1
 lxml>=4.9.0
 peewee>=3.16.2
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ setup(
     install_requires=['pandas>=1.3.0', 'numpy>=1.16.5',
                       'requests>=2.31', 'multitasking>=0.0.7',
                       'platformdirs>=2.0.0', 'pytz>=2022.5',
-                      'frozendict>=2.3.4', 'peewee>=3.16.2',
+                      'peewee>=3.16.2',
                       'beautifulsoup4>=4.11.1', 'curl_cffi>=0.15',
                       'protobuf>=3.19.0', 'websockets>=13.0'],
     extras_require={
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -1,0 +1,53 @@
+"""Tests for yfinance.data internal helpers."""
+import unittest
+from functools import lru_cache
+
+from yfinance.data import lru_cache_freezeargs
+from yfinance.utils import frozendict
+
+
+class TestFrozenDict(unittest.TestCase):
+    def test_is_dict_subclass(self):
+        d = frozendict({"a": 1, "b": 2})
+        self.assertEqual(d["a"], 1)
+        self.assertEqual(d.get("b"), 2)
+        self.assertEqual(set(d.keys()), {"a", "b"})
+        self.assertIsInstance(d, dict)
+
+    def test_is_hashable(self):
+        d1 = frozendict({"a": 1, "b": 2})
+        d2 = frozendict({"b": 2, "a": 1})
+        self.assertEqual(hash(d1), hash(d2))
+        self.assertEqual(len({d1, d2}), 1)
+
+    def test_mutation_raises(self):
+        d = frozendict({"a": 1})
+        with self.assertRaises(TypeError):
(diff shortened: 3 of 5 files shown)


ValueRaider:
No more AI.

dokson:
@ValueRaider please at least reopen this one 😞

ValueRaider:
Why? I want you to explain, not your AI.

dokson:
We can remove the `frozendict` (almost useless) hard dependency with just that internal class.....simple as that

ValueRaider:
Change case to `frozendict`, and only define class if it doesn't already exist https://peps.python.org/pep-0814.

dokson:
> Change case to `frozendict`, and only define class if it doesn't already exist https://peps.python.org/pep-0814.

gg

PR #2822, opened 2026-05-17
Drop multitasking dependency in favour of concurrent.futures.ThreadPoolExecutor

dokson:
## Summary
`multitasking` is used in exactly three places in `multi.py`, all of which have a clean stdlib equivalent in `concurrent.futures`:

| `multitasking` | replacement |
|---|---|
| `cpu_count()` | `os.cpu_count()` |
| `set_max_threads(N)` | `ThreadPoolExecutor(max_workers=N)` (cap is now actually enforced) |
| `@task` decorator + busy-wait join | `with ThreadPoolExecutor(...) as ex: ex.submit(...)` (context exit blocks on completion) |

No new dependency — `concurrent.futures` is stdlib. `multitasking>=0.0.7` removed from `requirements.txt` and `setup.py`.

## Why this isn't a behavior regression — benchmark

`multitasking` spawns a fresh OS thread per `@task` call and gates them with a semaphore. The semaphore cap is supposed to mirror `max_workers`, but in practice we observed thread counts vastly above the cap (presumably from churn during submit).

Benchmarked on the same machine, identical large-cap US ticker lists, distinct sets per N to defeat `cache_get` reuse:

| N | OLD (multitasking) | NEW (ThreadPoolExecutor) | Speedup | Peak OS threads OLD → NEW |
|---|---|---|---|---|
| 10 | 0.53s | 0.46s | 1.15× | 12 → 12 |
| 100 | 3.02s | **1.74s** | **1.74×** | 102 → 30 (−71%) |
| 500 | 6.98s | **3.00s** | **2.33×** | 257 → 30 (−88%) |

Strict improvement: faster at every N, far fewer OS threads at scale, no churn.

## Subtle behavior change worth flagging
`multitasking.task` silently swallowed any exception leaking out of a worker. The replacement re-raises after the pool drains:

```python
for f in futures:
    exc = f.exception()
    if exc is not None:
        raise exc
```

In practice the existing `_download_one` already catches its own errors into `ctx.errors`, so this only fires on programmer bugs in the worker path — a feature, not a regression.

## Test plan
- [x] Existing `tests/test_multi.py` passes locally (no test-side changes needed; the public API of `yf.download` is unchanged)
- [x] Live benchmark above (N=10/100/500) — strict improvement, no regression
- [x] `ruff check` clean

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 pandas>=1.3.0
 numpy>=1.16.5
 requests>=2.31
-multitasking>=0.0.7
 platformdirs>=2.0.0
 pytz>=2022.5
 frozendict>=2.3.4
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,7 @@ setup(
     keywords='pandas, yahoo finance, pandas datareader',
     packages=find_packages(exclude=['contrib', 'docs', 'tests', 'examples']),
     install_requires=['pandas>=1.3.0', 'numpy>=1.16.5',
-                      'requests>=2.31', 'multitasking>=0.0.7',
+                      'requests>=2.31',
                       'platformdirs>=2.0.0', 'pytz>=2022.5',
                       'frozendict>=2.3.4', 'peewee>=3.16.2',
                       'beautifulsoup4>=4.11.1', 'curl_cffi>=0.15',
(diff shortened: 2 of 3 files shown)


ValueRaider:
No more AI.

PR #2828, opened 2026-05-20
sync dev -> main

ValueRaider:
Tests pass.

Features
- #2761 @etbala 
- #2803 @dokson
- #2802 @dokson

Fixes
- #2777
- #2778 @a-drenaline
- #2780 @gottostartsomewhere
- #2794 @Jaypatel1511
- #2797
- #2801 @dokson
- #2805 @dokson
- #2804 @dokson
- #2825

Maintenance
- #2779 @gottostartsomewhere
- #2792 @etbala
- #2810 @dokson
- #2821 @dokson

PR #2832, opened 2026-05-25
Preserve Date/Datetime index name in yf.download() output

dokson:
## Summary

After the `reindex_dfs()` refactor from #2825 (released in 1.4.0 via #2828), `yf.download()` returns a DataFrame whose `index.name` is `None` instead of `Date` (daily) / `Datetime` (intraday). User code that does `data.stack().reset_index()` then references the date column by name now sees a `level_0` column and breaks. Reported in https://github.com/ranaroussi/yfinance/pull/2828#issuecomment-4536026065.

## Root cause

`reindex_dfs()` rebuilt the combined index with:

```python
all_indices = set()
for df in dfs.values():
    all_indices.update(df.index)
idx = sorted(all_indices)
idx = _pd.to_datetime(idx)
```

`pd.to_datetime(list)` returns an unnamed `DatetimeIndex`. `df.reindex(idx)` then propagates that nameless index to every per-ticker df, and the final `pd.concat` has nothing to preserve.

Before the refactor, every per-ticker df shared the same `index.name` (set by `history.py`), and `pd.concat(..., axis=1, sort=True)` preserved it.

## Fix

Use `Index.union` to build the combined index — it's the pandas-idiomatic way to merge indices and preserves the name natively when all sources agree:

```python
idx = None
for df in dfs.values():
    if df is not None and not df.empty:
        idx = df.index if idx is None else idx.union(df.index)
if idx is None:
    idx = _pd.DatetimeIndex([])
```

Mixed-name edge case (one df `Date`, another `Datetime` — shouldn't happen in practice) falls back to `None`, which is the safe pandas default.

Drive-by: added `isinstance(data.columns, _pd.MultiIndex)` narrow around `swaplevel` to silence a pre-existing pyright warning. No runtime behaviour change — after `concat(..., keys=...)` columns are always a MultiIndex.

## Test plan

- [x] New `tests/test_multi.py::TestDownloadIndexName::test_preserves_date_index_name` asserts `data.index.name == 'Date'` and that `stack().reset_index()` produces a `Date` column.
- [x] Smoke: `yf.download(['AAPL','MSFT'], interval='1d')` → `index.name == 'Date'`; `interval='1h'` → `index.name == 'Datetime'`.
- [x] `ruff check` / `pyright` clean.

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -324,6 +324,8 @@ class TestTickerHistory(unittest.TestCase):
                                                interval=interval, progress=False)
                             self.assertIsInstance(data, pd.DataFrame, "data has wrong type")
                             self.assertFalse(data.empty, "data is empty")
+                            expected_name = 'Datetime' if interval[-1] in ('m', 'h') else 'Date'
+                            self.assertEqual(data.index.name, expected_name)
                             if ignore_tz:
                                 self.assertIsNone(data.index.tz)
                             else:
--- a/yfinance/multi.py
+++ b/yfinance/multi.py
@@ -214,7 +214,7 @@ def _download_impl(ctx, tickers, start=None, end=None, actions=False, threads=Tr
                           keys=ctx.dfs.keys(), names=['Ticker', 'Price'])
     data.rename(columns=ctx.isins, inplace=True)
 
-    if group_by == 'column':
+    if group_by == 'column' and isinstance(data.columns, _pd.MultiIndex):
         data.columns = data.columns.swaplevel(0, 1)
         data.sort_index(level=0, axis=1, inplace=True)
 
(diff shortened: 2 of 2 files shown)


ValueRaider:
AI slop

dokson:
> AI slop

AI slop for speed, but the fix is correct my friend... @ValueRaider 

Just wanted to help the community.

dokson:
@ValueRaider also removed a warning in 
` if group_by == 'column': `

dokson:
https://github.com/ranaroussi/yfinance/pull/2825/changes pls...this is the cause...

ValueRaider:
It's slop because of the massive test change. What's wrong with adding to `test_download`?

dokson:
fair...should have just added an assert inside the existing test_download loop

PR #2838, opened 2026-06-01
feat: add force_download parameter to bypass cache

mfolofy:
When force_download=True, Ticker.history() skips cached data and fetches fresh from Yahoo Finance. Useful for real-time data during market hours. Resolves long-standing user request for cache bypass.


ValueRaider:
Only 2 types of history fetches should hit that cache:
- period
- start & end specified

Is that what you see?

Maybe makes more sense to automatically bypass cache when period set or end in future. No new argument then.

PR #2844, opened 2026-06-06
Fix Auth.check_login() always returning False

avesed:
## Problem
`Auth.check_login()` always returns `False`, even with valid login cookies set
via `set_login_cookies()`.

It looks for `<script id="nimbus-benji-config">` and reads `i13n.user.guid`.
That element is built **client-side by JS at runtime** (via
`document.getElementById("nimbus-benji-config")`), so it is **absent from the
server-fetched HTML**. `soup.find(...)` is therefore always `None` and the
method returns `False` regardless of login state.

Verified on `dev` with a real logged-in session (cookies set via
`Auth.set_login_cookies`): `check_login()` returns `False` even though
authenticated requests succeed.

## Fix
Parse the **server-rendered** `window.YAHOO.context` object instead, and treat
the session as logged in when `authed == "1"` and `user.guid` is present.

Observed difference (same homepage, anonymised):
- logged in:  `authed == "1"`, `user` has `guid` / `alias` / `login`
- logged out: `authed == "0"`, `user` has no `guid`

After the fix `check_login()` returns `True` for a logged-in session and
populates `Auth.user` (which also carries the per-session `crumb` already
present in that object).

### Notes
- `window.YAHOO.context` is a JS assignment, not a JSON `<script>`, so it is
  located by regex and sliced out with string-aware brace matching. Each
  candidate match is tried, so a stray occurrence inside a JS string literal is
  skipped; any parse failure fails safe to `False`.
- No reliable login-only API endpoint exists (`/v1/test/getcrumb` returns a
  crumb for anonymous sessions too), so the server-rendered context is the most
  reliable signal.
  crumb for anonymous sessions too), so the server-rendered context is the most
  reliable signal.
- The helper is module-level (mirroring `_is_transient_error`) so it can be
  unit-tested directly.

## Backward compatibility
No public API change. `check_login() -> bool` and `Auth.user -> dict | None`
keep their signatures, semantics, and the `self._user` caching short-circuit.
The old path never succeeded, so no caller observed the previous `_user` shape.

## Tests
Adds `tests/test_auth.py` — offline pure unit tests (mocked homepage, no login
cookies needed), 14 cases: logged-in / logged-out / missing context / non-dict
user / missing authed / empty guid / invalid JSON / decoy-string match /
`user` property / caching. Run: `python -m unittest tests.test_auth`.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -1,0 +1,137 @@
+"""Unit tests for yfinance.data.Auth login detection.
+
+These are pure unit tests: the Yahoo homepage response is mocked, so they run
+offline and require no Yahoo Finance login cookies. They lock in the fix for
+``check_login()`` (which previously looked for the client-side-only
+``<script id="nimbus-benji-config">`` element and therefore always returned
+``False``) by parsing the server-rendered ``window.YAHOO.context`` object.
+"""
+import unittest
+from unittest import mock
+
+from yfinance.data import Auth, _extract_window_yahoo_context
+
+
+# --- Synthetic fixtures (no real personal data) -----------------------------
+# Mimic the real page: window.YAHOO.context is assigned inside a <script>.
+_LOGGED_IN_HTML = """<!doctype html><html><head><script>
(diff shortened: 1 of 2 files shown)


avesed:
Find a better way.

PR #2845, opened 2026-06-07
Determine login and subscription tier via the subscriptions API

avesed:
Auth.check_login() parsed the finance.yahoo.com homepage for a 'nimbus-benji-config' element, which Yahoo builds client-side and is absent from the fetched HTML, so it always returned False even when logged in. Query the OBI subscriptions endpoint instead (/ws/obi-integration/v1/subscriptions): a lightweight JSON request (Yahoo rate-limits scraping of its large consumer web pages far more aggressively than its JSON APIs) where HTTP 200 with a guid means logged in and 401 means not.

Also add Auth.subscription_tier() -> 'gold'|'silver'|'bronze'|'free'|None, with the tier inferred from granted feature flags (stable across Yahoo's non-contiguous tier numbering). Auth.user returns {'guid': ...}.

The check is made live on each call rather than cached: the endpoint is cheap and not rate-limited at realistic volumes, so a fresh call keeps the answer correct after a runtime login/upgrade/expiry and lets the caller swap login cookies mid-process.

Add an allow_strategy_switch flag to YfData.get/_make_request; the login probe passes False so an expected 401 does not trigger the cookie-strategy toggle (which clears the session cookies, wiping T/Y). set_login_cookies() now also clears the cached crumb so the next request mints a crumb matched to the new login.

### DIFF
--- a/doc/source/reference/examples/auth.py
+++ b/doc/source/reference/examples/auth.py
@@ -3,16 +3,20 @@ import os
 
 auth = yf.Auth()
 
-# Set log in cookies from browser
-auth.set_login_cookies(os.getenv("COOKIE_T"), os.getenv("COOKIE_Y"))
-
-# Check if the cookies worked
-if auth.check_login():
+# Set login cookies obtained from your browser. The call stores them, validates
+# them with a live login check, and returns whether the account is logged in.
+if auth.set_login_cookies(os.getenv("COOKIE_T"), os.getenv("COOKIE_Y")):
     print("Logged in")
 else:
-    print("Invalid cookie")
+    print("Invalid or expired cookies")
 
-# Every subsequent request sent to Yahoo Finance will now be under the logged-in user.
+# Every subsequent request sent to Yahoo Finance is now under the logged-in user.
 
-# Access user information
-auth.user
+# Re-check the live login state at any time (re-queried each call, never cached).
+auth.check_login()        # -> True / False
+
+# Yahoo Finance subscription tier of the logged-in account.
+auth.subscription_tier()  # -> 'gold' / 'silver' / 'bronze' / 'free' / None
+
+# Access user information.
+auth.user                 # -> {'guid': ...} or None
--- a/doc/source/reference/yfinance.auth.rst
+++ b/doc/source/reference/yfinance.auth.rst
@@ -11,7 +11,8 @@ Authentication
 
 Class
 ------------
-The `Auth` module, allows you to login to Yahoo! Finance.
+The `Auth` module lets you log in to Yahoo! Finance, check the login state and
+read the account's subscription tier.
 
 .. autosummary::
    :toctree: api/
@@ -20,7 +21,7 @@ The `Auth` module, allows you to login to Yahoo! Finance.
 
 Auth Sample Code
 ------------------
-The `Auth` module, allows you to login to Yahoo! Finance.
+Logging in with browser cookies, then checking the login state and subscription tier:
 
 .. literalinclude:: examples/auth.py
    :language: python
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -1,0 +1,123 @@
+"""Unit tests for yfinance.data.Auth (login state + subscription tier).
+
+Pure unit tests: the OBI subscriptions endpoint response is mocked, so they run
+offline with no Yahoo login cookies. Auth derives both login state and the
+subscription tier from a single lightweight JSON call (no web-page scraping).
+The call is made live each time (not cached), so the answer never goes stale.
+"""
+import unittest
+from unittest import mock
+
+from yfinance.data import Auth, _SUBSCRIPTIONS_URL
+
+_GUID = "ABCDEFGHIJKLMNOPQRSTUV1234"
+
+
+def _result(guid=_GUID, tier=None, action="ACTIVE"):
+    """Build a subscriptions ``result`` with an optional subscription tier id."""
+    view = [{"action": action, "tier": tier}] if tier is not None else []
+    return {"guid": guid, "subscriptionView": view}
+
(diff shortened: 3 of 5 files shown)


avesed:
Subscription tiers are verified with real anonymous/free/Silver/Gold accounts.

ValueRaider:
@etbala This make sense?

ValueRaider:
Does `allow_strategy_switch` make more sense as class member not function argument?

avesed:
> Does `allow_strategy_switch` make more sense as class member not function argument?

Honestly, both work. I kept it as an arg because it is leak/race-free for free.

etbala:
This API looks like a significant improvement over the way Auth currently validates login credentials.

However, I am a confused about why you're checking the tier using feature flags instead of the raw integer. Sure, in theory Yahoo could switch the tier number but couldn't they just as easily (and probably more likely) switch the features of each tier?

Also, as far as `allow_strategy_switch`, this feels unnecessary. Logging a user out without telling them seems like a bad idea, but even if this was opt-in, what is the actual value? Logged in users should be able to see everything a free user sees, so switching back to unauthenticated shouldn't fix the 4xx error, right?

I don't have a premium account currently so I can't test the outputs outside of free/unauthenticated, do you have some sample outputs for Bronze/Silver/Gold that you could post here?

Also - just a suggestion, but since the API is much faster than the previous scraping implementation, consider calling `check_logon` directly in `set_login_cookies`.

avesed:
> This API looks like a significant improvement over the way Auth currently validates login credentials.
> 
> However, I am a confused about why you're checking the tier using feature flags instead of the raw integer. Sure, in theory Yahoo could switch the tier number but couldn't they just as easily (and probably more likely) switch the features of each tier?
> 
> Also, as far as `allow_strategy_switch`, this feels unnecessary. Logging a user out without telling them seems like a bad idea, but even if this was opt-in, what is the actual value? Logged in users should be able to see everything a free user sees, so switching back to unauthenticated shouldn't fix the 4xx error, right?
> 
> I don't have a premium account currently so I can't test the outputs outside of free/unauthenticated, do you have some sample outputs for Bronze/Silver/Gold that you could post here?
> 
> Also - just a suggestion, but since the API is much faster than the previous scraping implementation, consider calling `check_logon` directly in `set_login_cookies`.


<p>Thanks for the detailed review, I've reworked the PR. Point by point:</p>
<p>1. Agreed, switched to the integer.</p>

<p>2. <code>allow_strategy_switch</code> — agreed, dropped it. I reverted <code>_make_request</code> to the plain <code>if response.status_code >= 400:</code>.

<p>3. Sample outputs — below. Real captures for Gold/Free/Anonymous (+ an expired-cookie case). I don't have a Bronze account, so I can't 100% comfirm it's tier 3</p>

<p>4. <code>check_login()</code> in <code>set_login_cookies()</code>, agreed and adopted.

;Sample outputs
<pre><code class="language-jsonc">
// GOLD  ->  subscription_tier() == 'gold'   (HTTP 200)
{
  "result": {
    "guid": "<redacted>",
    "subscriptionView": [
      { "action": "ACTIVE", "tier": 6, "baseTier": 5, "premiumTier": 2,
        "inTrial": true, "paymentCycle": "monthly", "platform": "WEB",
        "sku": "<redacted>", "premiumSku": "<redacted>", "baseSku": "<redacted>",
        "productId": "<redacted>", "oathGuid": "<redacted>", "requestId": "<redacted>",
        "videoSubscriptionCrumb": "<redacted>", "trialEndDate": "<redacted>" }
    ],
    "premiumTierFeatures": { all 23 = true },
    "tierRanking": [3, 4, 5, 6]
  }
}

// FREE  ->  'free'   (HTTP 200)
{
  "result": {
    "guid": "<redacted>",
    "subscriptionView": [],
    "premiumTierFeatures": {
      "technicalEventsTeaser": true,
      "workspace": false, "communityInsights": false, "stockPicks": false,
      "researchReportsList": false, "companyOutlook": false, "researchReports": false,
      "technicalEvents": false, "premiumDashboard": false, "significantDevelopments": false,
      "coinbaseOne": false, "historicalTranscripts": false, "corporateEvents": false,
      "premiumNews": false, "historicalFinancials": false, "screenersDownload": false,
      "unlimitedPriceAlerts": false, "portfolioAnalytics": false, "adLite": false,
      "fairValue": false, "premiumScreeners": false, "investmentIdeas": false,
      "historicalStatistics": false
    },
    "tierRanking": [3, 4, 5, 6]
  }
}

// NOT LOGGED IN (no cookies)  ->  None   (HTTP 401, body: "failed", text/plain)
// EXPIRED / INVALID cookies   ->  None   (HTTP 403, body: "failed", text/plain)
</code></pre>
<p>I'll push these as updates to this PR.</p>

etbala:
There should also be some documentation changes here to reflect the updated `set_logon_cookies` and other functions you added.

Otherwise, LGTM.

avesed:
Doc: Just to be sure I edit the right file, <code> doc/source/reference/yfinance.auth.rst</code>? 

There’s also a potential issue, while <code>allow_strategy_switch</code>doesn't make to much sense here. Every other time a response >= 400 from any endpoint will trigger cookie strategy to switch, which silently clear user set cookie. I do think there still need a toggle (or some other way to preserve setted cookie from being wiped). I didn’t do anything here since it’s deeper into the cookie handling and out of scope for this pr.

etbala:
Yes, `yfinance.auth.rst` is the correct place. Also look at the `auth.py` example.

I'm good with leaving the cookie clearing issue for now, since it was a pre-existing issue, but ultimately up to @ValueRaider. If you don't end up fixing it please create an issue so someone else can fix it.

avesed:
Updated docs

PR #2847, opened 2026-06-09
Add regression test for empty ticker download index

### LINKED ISSUE #1382: Dataframe with incorrect index type returned when combining two tickers (one without data)

gyx09212214-prog:
## Summary
- Add a regression test for mixed successful and empty ticker downloads.
- Cover the case from #1382 where an empty ticker DataFrame must not coerce the combined download result index away from `DatetimeIndex`.

## Testing
- `.\.venv\Scripts\python -m pytest tests\test_multi.py -q`
- `.\.venv\Scripts\python -m pytest tests\test_multi.py tests\test_download_concurrency.py -q`

Closes #1382

### DIFF
--- a/tests/test_multi.py
+++ b/tests/test_multi.py
@@ -10,6 +10,33 @@ from tests.context import yfinance as yf
 
 class TestDownloadThreadSafety(unittest.TestCase):
 
+    def test_download_preserves_datetime_index_with_empty_ticker(self):
+        idx = pd.DatetimeIndex(['2024-01-02 09:30', '2024-01-02 09:31'], tz='America/New_York')
+        good_df = pd.DataFrame(
+            {'Open': [185.0, 186.0], 'Close': [185.5, 186.5]},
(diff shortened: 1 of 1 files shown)


ValueRaider:
rebase to dev https://ranaroussi.github.io/yfinance/development/code.html

PR #2851, opened 2026-06-09
Replace valuation-measures HTML scrape with timeseries API

avesed:
`Ticker.valuation` (valuation measures) scraped the `finance.yahoo.com/quote/{symbol}/key-statistics` web page. Yahoo rate-limits scraping of its consumer web pages far more aggressively than the query1/query2 JSON APIs, so this is a common source of HTTP 429 / throttling, and it silently returned an empty DataFrame whenever the page layout changed.

This fetches the values from the fundamentals-timeseries API instead (the same source as the income / balance-sheet / cash-flow statements), via `cache_get`. Output shape/format is unchanged (`Current` + M/D/YYYY columns, display strings); all 9 measures are always emitted (missing cells render as `--`, matching the page). Removes the only BeautifulSoup use in `quote.py`.

Verified byte-identical to the previous scrape for BRK-B, and live for AAPL/MSFT/BRK-B. Anonymous/free endpoint — no login required; independent of the Auth PRs (#2761 / #2845).

Tests in `tests/test_ticker.py`.

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -17,6 +17,7 @@ from tests.context import session_gbl
 from yfinance.exceptions import YFPricesMissingError, YFInvalidPeriodError, YFNotImplementedError, YFTickerMissingError, YFTzMissingError, YFDataException
 from yfinance.config import YfConfig
 
+import json
 import unittest
 # import requests_cache
 from unittest.mock import patch, MagicMock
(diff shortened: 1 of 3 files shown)


avesed:
Also adds a `freq` option to `get_valuation_measures` — "quarterly"(5 quarters, default, unchanged) / "monthly"(16 months) / "yearly"(4 years) / "trailing". "Current" always comes from the trailing series.("yearly" maps to the API’s annual type prefix.)

avesed:
After more tests, two follow-up changes:

1. **Unify value formatting** — abbreviate any measure ≥ 1000 with a k/M/B/T
   suffix + thousands grouping, matching the key-statistics page (which renders
   e.g. a P/E of 2303.92 as `2.30k`); NaN/inf/None → `--`. Replaces the
   market-cap-only abbreviation.

2. **Add a `periods` flag (default 5)** — the timeseries API occasionally
   returns more periods than the key-statistics page showed, so this flag caps
   the period columns to keep the output matching the original by default.
   `periods=None` returns everything.

ValueRaider:
I appreciate being byte-identical, but fix the types to be proper numeric. I know it is technically breaking, but `valuation` is relatively new so adoption should be limited.

avesed:
@ValueRaider Done. Now it returns raw floats straight from the API(full precision, e.g. `Market Cap` is `4275929952280.0`), with `NaN` for missing cells. The display formatter is reverted, I rebased to drop the now-obsolete formatting commit, so the branch is just the numeric values plus the `periods` cap.

ValueRaider:
Looks good. Git squash https://ranaroussi.github.io/yfinance/development/code.html

avesed:
> Looks good. Git squash https://ranaroussi.github.io/yfinance/development/code.html

Squashed.

PR #2849, opened 2026-06-09
Migrate to pyproject.toml

etbala:
Replaced `requirements.txt`, `setup.py`, etc. with the modern replacement `pyproject.toml`.

I didn't want to force adopting a specific tool, so for now you can just install dependencies using `pip install -e ".[dev]"`, or `pip install -e .` if you don't care about the development dependencies. However, I did change the workflow files to take advantage of uv instead of raw pip + python to speed them up.

I tested all the workflows in my fork, and everything worked as expected. Evidence:
https://github.com/etbala/yfinance/actions/runs/27188465084
https://github.com/etbala/yfinance/actions/runs/27188528139
https://github.com/etbala/yfinance/actions/runs/27187492438 # Failed because tests do not pass
https://github.com/etbala/yfinance/actions/runs/27188362416 # Failed because twine secrets not configured in my fork

PR #2861, opened 2026-06-14
ci: re-enable pytest, splitting offline tests from live ones

### LINKED ISSUE #2261: Unit test action can fail to fetch
The new unit test action is encountering Yahoo fetch failures e.g.
> FAILED tests/test_ticker.py::TestTickerEarnings::test_earnings_dates_with_limit - requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

This is despite it having a rate limiter. And test passes locally.

Almost every unit test involves fetching data, so only solution I see is removing the action. @dhruvan2006

dokson:
The pytest workflow was disabled because almost every test fetches from Yahoo and live fetches fail intermittently on CI IPs (`JSONDecodeError` on empty/throttled bodies). But 69 of 255 tests are offline unit tests that never touch the network: `tests/conftest.py` now marks the live ones `network`, so PRs run `pytest -m "not network"` (fast, deterministic) while the live suite moves to a scheduled job (`pytest-network.yml`, plus `workflow_dispatch`). `data.py` additionally retries empty 200 bodies (the throttling failure mode) under the existing `retries` config, which `tests/context.py` enables for the test run.

Rebased onto current `dev`: the live job keeps the uv setup and the version matrix from the disabled workflow, and the cache and price-repair tests each get their own step so their global state and their flakiness stay isolated. The `network` marker is registered in `pyproject.toml` next to `testpaths`.

Worth flagging: the inherited matrix includes 3.6-3.9, where `uv python install` has no builds for 3.6/3.7 and `pytest>=9.0.3` from the `dev` extra needs 3.10+. Those legs will fail regardless of this PR; `fail-fast: false` keeps the others running.

Closes #2261

### DIFF
--- a/.github/workflows/pytest-network.yml
+++ b/.github/workflows/pytest-network.yml
@@ -1,22 +1,24 @@
-name: Pytest
+name: Pytest (live)
 
 on:
-  pull_request:
-    branches:
-      - master
-      - main
-      - dev
+  schedule:
+    - cron: '0 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
 
 jobs:
-  test:
+  network:
+    # Live tests against Yahoo. Flaky by nature, so scheduled-only (off PRs).
     runs-on: ubuntu-latest
 
     steps:
     - name: Checkout code
-      uses: actions/checkout@v3
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
     - name: Set up Python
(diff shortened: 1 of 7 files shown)


ValueRaider:
Solved by #2849?

etbala:
#2849 didn't solve it, just modified the disabled file to use uv and test multiple python versions. This PR will need to rebase and integrate into that workflow.

dokson:
Rebased @etbala @ValueRaider

PR #2867, opened 2026-06-20
Fix missing comma splitting two equity screener EPS fields

he-yufeng:
### Bug

`EQUITY_SCREENER_FIELDS["income_statement"]` in `yfinance/const.py` is missing a comma between two adjacent string literals:

```python
"netepsbasic.lasttwelvemonths"
"netepsdiluted.lasttwelvemonths"},
```

Python concatenates adjacent string literals, so instead of two fields the set gets one invalid entry, `"netepsbasic.lasttwelvemonthsnetepsdiluted.lasttwelvemonths"`, and loses both `netepsbasic.lasttwelvemonths` and `netepsdiluted.lasttwelvemonths`.

Those two are validated against `EQUITY_SCREENER_FIELDS` in `EquityQuery._validate_eq_operand` / `_validate_btwn_operand`, so screening by either currently raises `ValueError: Invalid field`:

```python
EquityQuery('eq', ['netepsbasic.lasttwelvemonths', 0])  # raises on dev
```

They also drop out of the generated valid-fields docs table.

### Fix

Add the missing comma so both fields are present.

### Test

Added a regression test in `tests/test_screener.py` asserting both fields are present in `income_statement` (and the concatenated key is not), and that `EquityQuery` accepts them. It fails on `dev` and passes with this change; `python -m pytest tests/test_screener.py` is green.

### DIFF
--- a/tests/test_screener.py
+++ b/tests/test_screener.py
@@ -1,5 +1,6 @@
 import unittest
 from unittest.mock import patch, MagicMock
+from yfinance.const import EQUITY_SCREENER_FIELDS
 from yfinance.screener.screener import screen
 from yfinance.screener.query import EquityQuery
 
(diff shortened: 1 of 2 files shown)


ValueRaider:
The unit test is dumb. Your AI is dumb, you're supposed to provide the critical thought.

he-yufeng:
Fair point — that test just re-asserted the two field names, which is circular and wouldn't catch the next missing comma somewhere else.

Replaced it with a general invariant instead: every screener field is `metric` or `metric.period`, so a field carrying more than one `.` is two fields concatenated. That's exactly the failure mode here (the merged literal ended up with two `.lasttwelvemonths` segments), and it now guards every category rather than the one pair I happened to hit.

PR #2868, opened 2026-06-20
Fix capital gains parsing when events include a currency field

he-yufeng:
### Problem

`Ticker.history()` raises `ValueError: Length mismatch: Expected axis has 2 elements, new values have 1 elements` for funds/ETFs whose chart data carries capital-gains events with a `currency` field.

Yahoo returns capital-gains events with the same shape as dividends, i.e. `{date, amount, currency}`. In `parse_actions()` the dividends branch already handles that extra `currency` column, but the capital-gains branch assigns column names positionally:

```python
capital_gains.columns = ["Capital Gains"]
```

When `currency` is present the frame has two columns (`amount`, `currency`), so a single-element assignment blows up. The crash happens inside `parse_actions()` (called from `PriceHistory._get_price_history()`) before the `expect_capital_gains` check, so it takes down the whole `history()` call.

### Fix

Mirror what the dividends branch right above already does: drop the `currency` column when it is empty, then rename `amount` by name instead of assigning positionally.

```python
if 'currency' in capital_gains.columns and (capital_gains['currency'] == '').all():
    capital_gains = capital_gains.drop('currency', axis=1)
capital_gains = capital_gains.rename(columns={'amount': 'Capital Gains'})
```

Downstream reads `capital_gains['Capital Gains']` by name, so the result is unchanged for the normal single-column case and no longer crashes when `currency` shows up.

### Test

Added `TestParseActions.test_capital_gains_with_currency_field` in `tests/test_utils.py`, which feeds a `capitalGains` event dict containing `currency` and asserts the parsed column. It fails on `dev` with the `Length mismatch` ValueError and passes with this change. Full `tests/test_utils.py` is green (21 passed, 1 xfailed), no network needed.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -15,7 +15,7 @@ import pandas as pd
 
 import unittest
 
-from yfinance.utils import is_valid_period_format, _dts_in_same_interval, _parse_user_dt
+from yfinance.utils import is_valid_period_format, _dts_in_same_interval, _parse_user_dt, parse_actions
 
 
 class TestPandas(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


ValueRaider:
Example ticker symbol?

he-yufeng:
Honestly, I couldn't pin down a ticker that reproduces it against live data right now. I checked ~20 capital-gains-distributing funds (VWELX, FBALX, PRWCX, DODGX, AGTHX, …) plus a range of US and non-US dividend payers, via both the raw `v8/finance/chart` endpoint and `Ticker.get_capital_gains()`, and currently Yahoo returns dividend/capital-gain events with only `amount` and `date` keys — no `currency` — and capital-gains events are sparse to absent for those funds.

So I should be upfront: this isn't a currently-reproducing crash. It's a defensive-parity fix. The dividends branch right above already tolerates a `currency` column (`if 'currency' in dividends.columns and (dividends['currency'] == '').all(): drop`), which is presumably there because Yahoo did/does emit it in some context. The capital-gains branch instead assigns columns positionally (`capital_gains.columns = ["Capital Gains"]`), so if a `currency` column ever shows up there the same way it can for dividends, it raises `Length mismatch` and takes down the whole `history()` call rather than the capital-gains part. The change just brings capital gains to the same defensive level as dividends.

Given it's latent rather than actively reproducing, I'm happy to leave the call to you — close it if you'd rather not carry the extra guard, or keep it as cheap hardening. Either is fine by me.

ValueRaider:
AI slop

PR #2870, opened 2026-06-23
Document `back_adjust` and `progress` parameters in `download()`

furqan-khan07:
## Summary

The `download()` docstring documents every keyword argument except two that are part of its public signature:

- `back_adjust` (defaults to `False`)
- `progress` (defaults to `True`)

Since the docs site is generated from these docstrings, both parameters are currently missing from the published API reference for `yf.download()`, even though users can and do pass them.

This PR adds entries for both, following the existing format and placement (`back_adjust` next to its sibling `auto_adjust`, `progress` alongside the other download-behaviour flags).

## Changes

- `yfinance/multi.py`: document `back_adjust` and `progress` in `download()`.

Docs-only change — no functional impact.

### DIFF
--- a/yfinance/multi.py
+++ b/yfinance/multi.py
@@ -84,6 +84,8 @@ def download(tickers, start=None, end=None, actions=False, threads=True,
             Default is False
         auto_adjust: bool
             Adjust all OHLC automatically? Default is True
+        back_adjust: bool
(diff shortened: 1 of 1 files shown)


ValueRaider:
back_adjust should be deprecated for future removal. It's just duplicating auto_adjust

PR #2701, opened 2026-02-21
Fix "Nonetype is not iterable" crash in _fetch_info()

GimleLarpes:
Currently, if both `result` and `additional_info` are `None`, the line `if quote in result and len(result[quote]["result"]) > 0:` crashes.

This PR adds a simple check whether `result` is `None`.

Note this error is quite rare and unpredictable. It does not consistently occur for the same tickers.

I'll attach my log of when I encountered the crash:
```
Company - A1ZXY7.SG
HTTP Error 404: {"quoteSummary":{"result":null,"error":{"code":"Not Found","description":"Quote not found for symbol: A1ZXY7.SG"}}}
Failed to perform, curl: (16) . See https://curl.se/libcurl/c/libcurl-errors.html first for more details.
ERROR: Company - Builder: Error while parsing StaticCompanyData t=A1ZXY7.SG: argument of type 'NoneType' is not iterable
ERROR: CompanyList - Builder: Error while building CompanyList: argument of type 'NoneType' is not iterable
FATAL ERROR - The program will now exit
Traceback (most recent call last):
  File "/home/user/VSProjects/StockPicker9000/./stockpicker9000.py", line 166, in <module>
    companies = sd.CompanyList(ticker_dataframe)
  File "/home/user/VSProjects/StockPicker9000/stockdata.py", line 49, in __init__
    Log.e(self._TAG, f"Builder: Error while building CompanyList", exception=e, fatal=True)
    ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/VSProjects/StockPicker9000/utils.py", line 210, in e
    raise exception
  File "/home/user/VSProjects/StockPicker9000/stockdata.py", line 43, in __init__
    self._company_list_init(i)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/home/user/VSProjects/StockPicker9000/stockdata.py", line 58, in _company_list_init
    self.company_list[ticker] = Company(ticker)
                                ~~~~~~~^^^^^^^^
  File "/home/user/VSProjects/StockPicker9000/stockdata.py", line 390, in __init__
    self.set_static_company_data()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/user/VSProjects/StockPicker9000/stockdata.py", line 437, in set_static_company_data
[...]

PR #2871, opened 2026-06-24
Fix TypeError in Quote._fetch_info when Yahoo returns no data (#2865)

### LINKED ISSUE #2865: TypeError: argument of type 'NoneType' is not a container or iterable in _fetch_info
### Describe bug

`_fetch_info` crashes with a `TypeError` when both `_fetch()` and `_fetch_additional_info()` return `None`. The assignment on line 625 sets `result = additional_info` (i.e. `None`), and then line 629 performs a membership test on it:

```python
# quote.py lines 622–629
if additional_info is not None and result is not None:
    result.update(additional_info)
else:
    result = additional_info  # result is now None if both fetches returned None

...
if quote in result and ...  # TypeError: argument of type 'NoneType' is not iterable
```

Every other `_fetch` call in the file guards with `if result is None: return` before proceeding, so this looks like an oversight specific to `_fetch_info`.

The issue reproduces when requests are routed through a custom session (e.g. a proxy) that receives a null/blocked response from Yahoo. Direct connections return valid data for the same tickers, so the `None` path is only reachable when Yahoo returns an empty payload.

```python
import yfinance as yf
from curl_cffi import requests as curlRequests

# Proxy or any session that causes Yahoo to return a null payload
session = curlRequests.Session(impersonate="chrome", proxies={"http": "<proxy>", "https": "<proxy>"})

ticker = yf.Ticker("NPCE", session=session)
info = ticker.info  # raises TypeError
```

Direct call (no session) succeeds for the same tickers — the `None` path is only hit when the session causes Yahoo to return a null `result` field.

```python
yf.config.debug_logging = True
```

```
File ".../yfinance/scrapers/quote.py", line 629, in _fetch_info
    if quote in result and len(result[quote]["result"]) > 0:
       ^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not a container or iterable
```

### Bad data proof

N/A — this is a crash, not incorrect data.

### `yfinance` version

1.4.1

### Python version

3.14

### Operating system

Windows 11

SriniV-1:
### Summary

`Quote._fetch_info()` raises `TypeError: argument of type 'NoneType' is not iterable` when **both** `_fetch()` and `_fetch_additional_info()` return `None` — e.g. when Yahoo returns an empty/blocked payload through a proxy session (reported in #2865). The membership test `if quote in result` then runs on `None`.

While fixing this I noticed the same merge block also **silently discarded a valid primary result** whenever the additional-info fetch returned `None`: the old `else` branch set `result = additional_info` (`None`), throwing away good data.

### Changes

- Restructure the merge so a valid `result` is preserved when there's no additional info, and `additional_info` is used as a fallback when `result` is missing.
- Return early (leaving `_info` unset, which the `info` property and `_fetch_complementary` already handle) when **both** fetches yield no data, instead of crashing on the membership test below. This matches the `if result is None: return` guard used elsewhere in the file.
- Add offline regression tests (`tests/test_quote.py`) covering all three branches (both-None, valid-result-only, additional-only).

### Testing

```
pytest tests/test_quote.py   # 3 passed (network mocked)
```

Fixes #2865.

_Note: #2701 attempted the crash fix but has been conflicting/stale for ~2 months and has no test. This PR is rebased on current `main`, additionally fixes the data-loss path, and adds regression coverage. Happy to coordinate with the original author if preferred._

### DIFF
--- a/tests/test_quote.py
+++ b/tests/test_quote.py
@@ -1,0 +1,50 @@
+"""
+Unit tests for yfinance.scrapers.quote.Quote.
+
+These are offline tests (the network fetches are mocked), covering the
+no-data / partial-data paths of ``_fetch_info`` — see issue #2865.
+"""
+
+import unittest
+from unittest.mock import MagicMock, patch
+
+from yfinance.scrapers.quote import Quote
+
+
+class TestQuoteFetchInfo(unittest.TestCase):
+    def _make_quote(self):
+        # _data is only used by the (mocked) fetch helpers, so a stub is fine.
+        return Quote(MagicMock(), "TEST")
+
+    def test_no_crash_when_both_fetches_return_none(self):
+        """Both Yahoo endpoints empty must not raise TypeError (#2865)."""
+        q = self._make_quote()
+        with patch.object(q, "_fetch", return_value=None), \
+             patch.object(q, "_fetch_additional_info", return_value=None):
+            q._fetch_info()  # previously: "argument of type 'NoneType' is not ..."
+        self.assertIsNone(q._info)
+
+    def test_valid_result_kept_when_additional_info_is_none(self):
+        """A valid primary result must not be discarded when there's no
+        additional info (the old ``else`` branch nulled it)."""
+        q = self._make_quote()
+        valid = {"quoteSummary": {"result": [{"symbol": "TEST", "shortName": "Test Co"}]}}
+        with patch.object(q, "_fetch", return_value=valid), \
+             patch.object(q, "_fetch_additional_info", return_value=None):
+            q._fetch_info()
+        self.assertIsNotNone(q._info)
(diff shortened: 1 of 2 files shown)


ValueRaider:
Does this fix? https://github.com/ranaroussi/yfinance/commit/eea80d7ed3a21f31405b3a841b3731ebef7848b0

SriniV-1:
Yes — `eea80d7` (#2869) covers it: setting `result = {}` when both fetches return `None` prevents the `NoneType` membership crash, and only updating from `additional_info` when it's present avoids discarding a valid `result`. That's the same outcome this PR aimed for, so it's redundant now. Closing in favor of #2869 — thanks for checking, and apologies for the duplicate.

PR #2873, opened 2026-06-25
fix(funds): log raw response instead of unbound `data` in FundsData error handler

he-yufeng:
### Bug

`FundsData._fetch_and_parse()` binds `data` inside the `try` block and then logs it from the broad `except Exception` handler:

```python
result = self._fetch()
try:
    data = result["quoteSummary"]["result"][0]
    ...
except Exception as e:
    ...
    logger.debug(f" {data}")   # data may be unbound here
```

When the failure happens on that first statement, `data` is never assigned, so the handler raises `UnboundLocalError` and hides the real error.

This is reachable in normal use, not a malformed-input edge case: for a symbol that has no fund data, Yahoo's quoteSummary endpoint returns `{"quoteSummary": {"result": None, "error": {...}}}`. Then `result["quoteSummary"]["result"]` is `None`, `None[0]` raises `TypeError` (not `KeyError`, so it skips the `except KeyError` branch), and with the default `hide_exceptions=True` the code reaches the `data` log line and blows up with:

```
UnboundLocalError: cannot access local variable 'data' where it is not associated with a value
```

masking the underlying `'NoneType' object is not subscriptable`.

### Fix

Log `result` instead of `data`. `result` is assigned before the `try` block, so it's always bound, and it's the raw response the debug line was meant to show anyway.

### Testing

Added `tests/test_funds.py` reproducing the case (`_fetch` returns a no-data response). It fails on `dev` with `UnboundLocalError` and passes with the fix. Ran locally:

- `pytest tests/test_funds.py` — passes with the fix, fails before it
- `pyright . --level error` — clean
- `ruff check` — clean

(The pytest workflow is currently disabled in CI, so the new test won't run there, but it documents the regression and is ready if pytest is re-enabled.)

### DIFF
--- a/tests/test_funds.py
+++ b/tests/test_funds.py
@@ -1,0 +1,37 @@
+"""Tests for yfinance.scrapers.funds error handling."""
+import unittest
+from unittest.mock import MagicMock, patch
+
+from yfinance.config import YfConfig
+from yfinance.scrapers.funds import FundsData
+
+
+class TestFundsDataErrorHandling(unittest.TestCase):
+    def setUp(self):
+        self._orig_hide = YfConfig.debug.hide_exceptions
+        YfConfig.debug.hide_exceptions = True
+
+    def tearDown(self):
+        YfConfig.debug.hide_exceptions = self._orig_hide
+
+    def test_no_fund_data_does_not_raise_unbound_local(self):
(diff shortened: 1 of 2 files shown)


ValueRaider:
AI slop

PR #2881, opened 2026-06-30
Preserve cached history metadata when the tradingPeriods enrichment fetch fails (#2745)

### LINKED ISSUE #2745: get_history_metadata() raises YFPricesMissingError for valid tickers when intraday data is unavailable

8910work-cell:
Fixes #2745.

### Problem
`get_history_metadata()` makes a secondary intraday request (`period="5d", interval="1h"`) to populate `tradingPeriods`. For some valid, non-delisted tickers/exchanges (e.g. `ROG.SW`) that request returns no prices and raises `YFPricesMissingError` — and before raising, `_get_history_cache` overwrites `self._history_metadata` with `{}`.

As a result, even though metadata (`currency`, `exchangeName`, `timezone`, …) was already populated by a prior `history()` call, the method both **raises** and **destroys** that cached metadata.

### Fix
In `get_history_metadata()`, snapshot the existing metadata before the optional enrichment fetch. If that fetch raises `YFPricesMissingError`, restore the snapshot and return it instead of propagating — `tradingPeriods` is optional enrichment. When there was no prior metadata at all, the error is still raised (unchanged behaviour). The shared `_get_history_cache` is left untouched to keep the change minimal.

### Tests
Adds `tests/test_history_metadata.py` (network-free, mocks the secondary fetch):
- metadata already populated → preserved & returned when the enrichment fetch fails
- no prior metadata → `YFPricesMissingError` still raised

Verified red→green: the first test fails on `dev` without the fix and passes with it.

### DIFF
--- a/tests/test_history_metadata.py
+++ b/tests/test_history_metadata.py
@@ -1,0 +1,52 @@
+"""
+Regression tests for PriceHistory.get_history_metadata().
+
+These do not hit the network: the secondary intraday request that
+get_history_metadata() makes to populate 'tradingPeriods' is mocked.
+"""
+import unittest
+from unittest.mock import patch
+
+from yfinance.scrapers.history import PriceHistory
+from yfinance.exceptions import YFPricesMissingError
+
+
+def _make_price_history():
+    # PriceHistory only needs (data, ticker, tz) for these code paths;
+    # _data is never touched because _get_history_cache is mocked.
+    return PriceHistory(data=None, ticker="TEST", tz="UTC")
+
+
+class TestGetHistoryMetadata(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


ValueRaider:
I'm thinking something different - convert `history_metadata` to a lazy-loading dict like `FastData`, so `tradingPeriods` only gets fetched on access.

ValueRaider:
#2922

PR #2892, opened 2026-07-08
Fix pandas Timedelta DeprecationWarning for interval strings (#2882)

### LINKED ISSUE #2882: DeprecationWarning: The 'generic' unit for NumPy timedelta is deprecated` in `utils.py` with recent NumPy/Pandas
### Describe bug

When fetching data using `yfinance` with recent versions of `numpy` (>= 2.5.0) and `pandas` (>= 2.2.x), the console is flooded with `DeprecationWarning` messages originating directly from `yfinance/utils.py`. 

The warning indicates that `pd.Timedelta()` is being called in a way that relies on the deprecated "generic" unit for NumPy timedeltas (e.g., implicit conversion of bare integers or missing explicit units).

While this does not break the execution flow yet, it clutters the console output significantly, especially when batch-downloading multiple symbols and intervals. More importantly, as the warning states, this implicit conversion will raise a hard `Exception` in future major releases of Pandas/NumPy, which will break `yfinance` data fetching entirely.

### Minimal Reproducible Example
```python
import yfinance as yf

# Fetching any symbol with multiple intervals triggers the warning internally
df = yf.download("^GVZ", period="max", interval="1h")
````

### Environment

- **OS**: Windows 10/11 (reproducible across platforms)
- **Python version**: 3.12.13
- **yfinance version**: 1.5.1
- **pandas version**: >= 2.2.0
- **numpy version**: >= 2.5.0 (Released June 2026)

### Traceback / Console Output

```text
C:\...\yfinance\utils.py:484: DeprecationWarning: The 'generic' unit for NumPy timedelta is deprecated, and will raise an error in the future. This includes implicit conversion of bare integers (e.g. `+ 1`).Please use a specific unit instead.
  return _pd.Timedelta(interval)

C:\...\yfinance\utils.py:667: DeprecationWarning: The 'generic' unit for NumPy timedelta is deprecated, and will raise an error in the future. This includes implicit conversion of bare integers (e.g. `+ 1`).Please use a specific unit instead.
  last_rows_same_interval = (dt2 - dt1) < _pd.Timedelta(interval)
```

### Additional Context & Root Cause Analysis

The root cause seems to be located in `yfinance/utils.py` around lines **484** and **667**, where `_pd.Timedelta(interval)` is called.

Depending on the internal type of `interval` at runtime, Pandas/NumPy now strictly requires explicit time units (e.g., passing `unit='s'` or ensuring the string is parsed correctly like `"1h"`) rather than relying on generic implicit integer conversions.

### Temporary Workaround for Users

Until an official patch is released, users can suppress this specific warning at the entry point of their scripts to keep their logs clean:

```python
import warnings

# TODO: Remove this filter once yfinance updates utils.py for Pandas 2.2+ / NumPy 2.5+
warnings.filterwarnings(
    "ignore", 
    message=".*The 'generic' unit for NumPy timedelta is deprecated.*"
)
```

Thank you for maintaining this awesome library!

### Bad data proof

_No response_

### `yfinance` version

1.5.1

### Python version

Python 3.12.13

### Operating system

Windows

gottostartsomewhere:
## Summary

Fixes #2882.

Interval strings such as `"5d"` were passed straight to `pd.Timedelta(interval)`, which relies on pandas' single-letter unit aliases (e.g. `'d'`). On **pandas >= 3.0 / numpy >= 2.5** these emit a `DeprecationWarning` (`Pandas4Warning: 'd' is deprecated ... use 'D'`) that is slated to become a hard error in a future release — which would break history fetching.

`"5d"` is a valid interval and reaches the generic `else` branch of `_dts_in_same_interval`, so it warns in normal use.

## Change

- Add `_interval_to_pd_timedelta()`, which builds the `Timedelta` from an interval string using **explicit keyword units** (`Timedelta(days=5)` etc.), avoiding the deprecated aliases entirely.
- Route the two affected call sites (`_interval_to_timedelta` and `_dts_in_same_interval`) through it. Values are unchanged.

## Tests

- `test_multi_day_interval` — functional check for the `5d` path.
- `test_no_deprecation_warning_for_valid_intervals` — asserts every valid interval (`1m`…`3mo`) is warning-free (with `DeprecationWarning`/`FutureWarning` promoted to errors). This fails on the current code and passes with the fix.
- `test_interval_to_pd_timedelta_values` — value checks for the helper.

Full `tests/test_utils.py` passes under `python -W error::DeprecationWarning` (24 tests).

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -10,12 +10,19 @@ Specific test class:
 """
 from datetime import datetime
 from unittest import TestSuite
+import warnings
 
 import pandas as pd
 
 import unittest
 
-from yfinance.utils import is_valid_period_format, _dts_in_same_interval, _parse_user_dt
+from yfinance.utils import (
+    is_valid_period_format,
+    _dts_in_same_interval,
+    _parse_user_dt,
+    _interval_to_timedelta,
+    _interval_to_pd_timedelta,
+)
 
 
 class TestPandas(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


ValueRaider:
Beat by #2891

PR #2896, opened 2026-07-11
Fix AttributeError in Lookup error handling (self.ticker -> self.query)

nyxst4ck:
### Problem

`Lookup` stores the search term as `self.query`, but two error-handling paths in `Lookup._fetch_lookup` reference `self.ticker`, which the class never defines:

```python
# yfinance/lookup.py
self._logger.error(f"{self.ticker}: 'lookup' fetch received faulty data")   # faulty-JSON path
...
raise YFDataException(f"{self.ticker}: 'lookup' fetch returned error: {error}")  # Yahoo error path
```

When Yahoo returns a `finance.error` payload (e.g. bad request / rate limit) or non-JSON body, the intended `YFDataException` / error log is never produced. Instead the interpreter raises a confusing `AttributeError: 'Lookup' object has no attribute 'ticker'`, masking the real cause.

### Root cause

Copy/paste of the `self.ticker` pattern used elsewhere in the library (e.g. `TickerBase`) into `Lookup`, which exposes the term as `self.query` (see the debug log on the line just above, which already correctly uses `self.query`).

### Fix

Replace both `self.ticker` references with `self.query`. Two-character change per line, no behavior change beyond producing the intended error.

### Test

Added an offline unit test (`tests/test_lookup.py::TestLookupErrorHandling`) that mocks the HTTP layer to return a `finance.error` payload and asserts `YFDataException` is raised with the query in the message.

- On current `dev` (before the source fix): the test fails with `AttributeError: 'Lookup' object has no attribute 'ticker'`.
- With the fix: `YFDataException` is raised as intended and the test passes.

Verified `ruff check` and `pyright --level error` pass on the changed files.

### DIFF
--- a/tests/test_lookup.py
+++ b/tests/test_lookup.py
@@ -3,6 +3,7 @@ import unittest
 import pandas as pd
 
 from tests.context import yfinance as yf, session_gbl
+from yfinance.exceptions import YFDataException
 
 
 class TestLookup(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


ValueRaider:
Looks fine to me. Waiting for you to mark ready for review.

nyxst4ck:
Marked ready for review. I do not see any checks reported on the branch yet.

PR #2898, opened 2026-07-13
Guard get_shares_full against empty/missing share-count timeseries

joaopedroassad:
## Summary

- `TickerBase.get_shares_full` reads `json_data["timeseries"]["result"]` and then indexes `shares_data[0]` with no guard. When Yahoo returns a 200 whose share-count timeseries envelope has an empty `result` (`[]`), a `null` `result`, or omits `timeseries` entirely (the `finance` error envelope), this raises a raw `IndexError` / `TypeError` / `KeyError` that leaks straight to the caller and ignores `hide_exceptions`.
- Empty `result` is the common case: symbols with no share-count history (a lot of non-US tickers, ETFs, indices) come back as `{"timeseries": {"error": null, "result": []}}`, so `shares_data[0]` hits `IndexError: list index out of range`.
- The break isn't limited to the public `get_shares_full()` call. `Quote.shares` calls `get_shares_full` internally to populate the `shares` complementary field, so the same exception can take down `.info` for the whole ticker.
- Fix is minimal and matches how the rest of the function already degrades: pull `result` defensively and return `None` when it's empty, exactly like the existing `if "shares_out" not in shares_data[0]: return None` exit right below it. Every other failure path in this function already returns `None`.

```python
shares_data = (json_data.get("timeseries") or {}).get("result") or []
if not shares_data or "shares_out" not in shares_data[0]:
    return None
```

This is the same class of empty/malformed-timeseries guard as #2877 (Complementary info: guard against empty or missing timeseries result), which fixed the identical `json_result["result"][0]` pattern in `quote.py`. This is the next-door site in `base.py` that reads the same share-count timeseries shape and was still unguarded.

## Testing

Ran in a fresh venv with an editable install:

```
python -m venv .venv && . .venv/bin/activate
pip install -e . pytest ruff
```

Red -> green on the new test:

```
# stock dev (fix reverted): FAILS with IndexError: list index out of range at base.py:541
python -m pytest tests/test_ticker.py::TestTickerInfo::test_shares_full_empty_result_timeseries -v
# with the fix: PASSES
python -m pytest tests/test_ticker.py::TestTickerInfo::test_shares_full_empty_result_timeseries -v
```

Adjacent complementary-info guard tests still pass:

```
python -m pytest tests/test_ticker.py::TestTickerInfo::test_complementary_info_sparse_timeseries tests/test_ticker.py::TestTickerInfo::test_complementary_info_empty_result_timeseries -v
```

Lint / compile clean on changed files:

```
ruff check yfinance/base.py tests/test_ticker.py    # All checks passed!
python -m py_compile yfinance/base.py tests/test_ticker.py
git diff --check
```

The new test is fully mock-based (patches `_data.cache_get`, sets `_tz` to skip the network tz lookup), so it needs no network. It covers the four failing shapes (empty list, missing `result`, null `result`, missing `timeseries`) plus a happy path that still returns a populated Series.

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -1288,6 +1288,47 @@ class TestTickerInfo(unittest.TestCase):
             q._fetch_complementary()
         self.assertEqual(q._info["trailingPegRatio"], 1.23)
 
+    def test_shares_full_empty_result_timeseries(self):
+        # get_shares_full reads json_data["timeseries"]["result"] and then
+        # shares_data[0] without guarding. Yahoo can return a 200 whose
+        # share-count timeseries envelope has an empty/missing/null 'result'
+        # (common for symbols with no share-count history, and for the
+        # 'finance' error envelope that omits 'timeseries'). The unguarded
+        # access raised IndexError/KeyError/TypeError, which also broke the
+        # 'shares' complementary field feeding .info. It should return None
+        # instead, matching the function's other graceful-degradation exits.
+        ticker = yf.Ticker("TEST", session=self.session)
+        ticker._tz = "America/New_York"  # avoid network tz lookup
+
+        return_none = [
+            # 'result' is an empty list -> would raise IndexError
+            {"timeseries": {"error": None, "result": []}},
+            # 'result' key missing -> would raise KeyError
+            {"timeseries": {"error": None}},
+            # 'result' is null -> would raise TypeError
+            {"timeseries": {"error": None, "result": None}},
+            # 'timeseries' absent (finance error envelope) -> would raise KeyError
+            {"finance": {"result": None, "error": {"code": "Not Found"}}},
+        ]
+        for payload in return_none:
+            mock_resp = MagicMock()
+            mock_resp.json.return_value = payload
+            with patch.object(ticker._data, "cache_get", return_value=mock_resp):
+                self.assertIsNone(ticker.get_shares_full(),
+                                  f"Expected None for payload: {payload}")
+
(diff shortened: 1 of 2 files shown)


ValueRaider:
> When Yahoo returns a 200 whose share-count timeseries envelope has an empty result

Example?

joaopedroassad:
I checked and can't get an example on this call, I was mistaken.

For symbols with no share count (indices, ETFs, FX, crypto, even invalid or delisted tickers), Yahoo returns 200 with a one-item, meta-only result:

    {"timeseries": {"error": null,
                    "result": [{"meta": {"symbol": ["SPY"],
                                         "type": ["shares_out"]}}]}}

So it never reaches the IndexError. `result[0]` exists, and the existing `"shares_out" not in shares_data[0]` check already returns None. No reproducible crash on get_shares_full as it stands.

The only way I could get an empty or null envelope was to ask for a `type` that Yahoo doesn't recognize, which comes back as a 404 with `"result": null`. But `get_shares_full` doesn't send a `type`, it just relies on the server default, so that path isn't reachable here.

So this is defensive hardening rather than a fix for a real crash. It keeps `get_shares_full` returning `None` on an empty, null, or missing timeseries envelope, which matches how the rest of the function already degrades (including the existing `finance` "Bad Request" branch), and the test just documents that. If you'd rather hold off until there's a case that actually reproduces, I'm fine with closing the PR.

ValueRaider:
Come back when you have an example. It matters because I want to see if structure contains different data.

Issue #2052, opened 2024-09-10
💥 Is yahoo finance broken? 


paintdog:
### Describe bug

Since yesterday, Yahoo Finance has not been providing data for several stocks, making it impossible for me to download them using your software. For example, for [Honeywell](https://finance.yahoo.com/quote/HON/history/?frequency=1d&period1=779068800&period2=1725915845) – a Nasdaq 100 stock (!!!) – Yahoo Finance states: "There are no [data] in the selected time period."

I understand that this is likely a Yahoo Finance issue, but your software becomes unusable if the platform behind it stops providing data. As far as I can tell, this issue has affected a wide range of stocks (some US, mostly European) since yesterday.

Are others experiencing this problem as well? Does this happen from time to time, or is it a sign of a bigger issue? Any suggestions? Will Yahoo fix this problem, or are they already aware of it? Is Yahoo discontinuing data support?

### Simple code that reproduces your problem

No simple code, as I use your software within my own modules.

### Debug log

see above

### Bad data proof

see above

### `yfinance` version

0.24.3

### Python version

Python 3.11

### Operating system

Linux Mint 21.3

divyankm:
Seems some Symbols are working, while some symbols are not working for yahoofinance since yesterday.
yfinance version: 0.2.43
python version: 3.10.12

Working Symbol: INDIGO.NS
```
import yfinance as yf
symbol = 'INDIGO.NS'  
df = yf.download(symbol, start='2022-11-05', interval='1h')
df.head()
>> 
Open | High | Low | Close | Adj Close
-- | -- | -- | -- | --

1774.849976 | 1817.699951 | 1774.849976 | 1801.349976 | 1801.349976
1800.599976 | 1802.000000 | 1769.050049 | 1777.800049 | 1777.800049
1778.050049 | 1788.000000 | 1774.150024 | 1778.050049 | 1778.050049
1784.849976 | 1784.849976 | 1676.300049 | 1683.650024 | 1683.650024
1683.400024 | 1721.000000 | 1680.150024 | 1720.000000 | 1720.000000
```

Non Working Symbol since Yesterday:
```
import yfinance as yf
symbol = 'TCS.NS'  
df = yf.download(symbol, start='2022-11-05', interval='1h')
df.head()
>> ERROR:yfinance:Could not get exchangeTimezoneName for ticker 'TCS.NS' reason: 'chart'
[*********************100%***********************]  1 of 1 completed
ERROR:yfinance:
[...]

Issue #2044, opened 2024-09-03
Unable to get history of ASIANPAINT.NS on May 4, 2022


tapadeep:
### Describe bug

My previously working script has started to throw the following error:
> ERROR:yfinance:Could not get exchangeTimezoneName for ticker 'ASIANPAINT.NS' reason: 'chart'
ERROR:yfinance:$ASIANPAINT.NS: possibly delisted; no timezone found

### Simple code that reproduces your problem

```
import yfinance as yf
import datetime as datetime0
from datetime import datetime
yf.enable_debug_mode()
yf.__version__
stock=yf.Ticker('ASIANPAINT.NS')
date=datetime.strptime('4-May-22','%d-%b-%y').date()
details=stock.history(start=date,end=date+datetime0.timedelta(days=1))

### Debug log

> DEBUG    Entering history()
DEBUG:yfinance:Entering history()
DEBUG     Entering _fetch_ticker_tz()
DEBUG:yfinance: Entering _fetch_ticker_tz()
ERROR      Could not get exchangeTimezoneName for ticker 'ASIANPAINT.NS' reason: 'chart'
ERROR:yfinance:  Could not get exchangeTimezoneName for ticker 'ASIANPAINT.NS' reason: 'chart'
DEBUG      Got response: 
DEBUG:yfinance:  Got response: 
DEBUG      -------------
DEBUG:yfinance:  -------------
DEBUG       {'finance': {'result': None, 'error': {'code': 'Internal Server Error', 'description': 'Duplicate key 1497411900'}}}
DEBUG:yfinance:   {'finance': {'result': None, 'error': {'code': 'Internal Server Error', 'description': 'Duplicate key 1497411900'}}}
DEBUG      -------------
DEBUG:yfinance:  -------------
DEBUG     Exiting _fetch_ticker_tz()
DEBUG:yfinance: Exiting _fetch_ticker_tz()
DEBUG     Entering history()
DEBUG:yfinance: Entering history()
ERROR      $ASIANPAINT.NS: possibly delisted; no timezone found
ERROR:yfinance:  $ASIANPAINT.NS: possibly delisted; no timezone found
DEBUG     Exiting history()
DEBUG:yfinance: Exiting history()
DEBUG    Exiting history()
DEBUG:yfinance:Exiting history()

### Bad data proof

_No response_

### `yfinance` version

0.2.43

### Python version

3.10.12

### Operating system

Ubuntu v22.04

ValueRaider:
[...]

Issue #1797, opened 2023-12-22
Error downloading data for "AAPL": No price data found (1d 2019-01-01 -> 2021-06-12)


YeakubSadlil:
### Describe bug

I'm encountering an error when attempting to download AAPL data and stock ticker using yfinance. The specific error message is:
```
[*********************100%%**********************]  1 of 1 completed

1 Failed download:
['AAPL']: Exception('%ticker%: No price data found, symbol may be delisted (1d 2019-01-01 -> 2021-01-01)')
```

I tried running the code snippet on other laptops and on Kaggle, where it worked well. However, it fails to download data only on my laptop.

_Troubleshooting Steps Attempted:_

- Reinstalling yfinance
- Checking several date range
- Checking several times on different days
- Checking with different stock tickers

### Simple code that reproduces your problem

```
import pandas as pd
import yfinance as yf

aapl_df = yf.download('AAPL', 
                    start='2019-01-01', 
                    end='2021-06-12',
)
```

### Debug log

```
DEBUG    Entering download()
DEBUG:yfinance:Entering download()
DEBUG     Disabling multithreading because DEBUG logging enabled
DEBUG:yfinance: Disabling multithreading because DEBUG logging enabled
DEBUG     Entering history()
DEBUG:yfinance: Entering history()
DEBUG      AAPL: Yahoo GET parameters: {'period1': '2019-01-01 00:00:00-05:00', 'period2': '2021-01-01 00:00:00-05:00', 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG:yfinance:  AAPL: Yahoo GET parameters: {'period1': '2019-01-01 00:00:00-05:00', 'period2': '2021-01-01 00:00:00-05:00', 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG      Entering get()
DEBUG:yfinance:  Entering get()
DEBUG       url=https://query2.finance.yahoo.com/v8/finance/chart/AAPL
DEBUG:yfinance:   url=https://query2.finance.yahoo.com/v8/finance/chart/AAPL
DEBUG       params=frozendict.frozendict({'period1': 1546318800, 'period2': 1609477200, 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'})
[...]

Issue #1713, opened 2023-10-02
yFinance crash every midnight: No Price data found, Symbol may be delisted. 


LauravanVught:
### Describe bug

I'm checking yFinance historical data for my tradingbot. However, every night (around midnight) the bot crashed because yahoo returns the exception: 

`2023-10-02 **02:00:19,167** - ERROR: LINK-USD: No price data found, symbol may be delisted (period=1d)`

[...]

PR #2903, opened 2026-07-14
Don't claim 'possibly delisted' when Yahoo explains the missing data

### LINKED ISSUE #2902: Error messages claim "possibly delisted" even when Yahoo explains the real cause
### Describe bug

Every `YFPricesMissingError` message starts with `$TICKER: possibly delisted; ...`, even when Yahoo's response contains an explicit reason that has nothing to do with delisting. Example, a 30m request older than Yahoo's 60 day intraday window:

```
$SPY: possibly delisted; no price data found  (30m 2026-03-16 -> 2026-03-23) (Yahoo error = "15m data not available for startTime=1773633600 and endTime=1774238400. The requested range must be within the last 60 days.") (30m resampled from 15m)
```

SPY is not delisted. The request was outside the allowed range, and Yahoo said so in the quoted error. The speculation sends users down the wrong path (see #1713, #1797, #2044, #2052 for examples of the confusion). When a ticker really is dead, Yahoo's own description already says "No data found, symbol may be delisted", so the prefix adds nothing there either.

### Suggested fix

When history() has an explicit error description from Yahoo, suppress the "possibly delisted" prefix and let Yahoo's stated reason stand. Keep the prefix where Yahoo returns nothing and delisting is a plausible explanation. Happy to submit a PR, as a follow-up to #2900 which fixed the interval part of this same message.

```python
import yfinance as yf
import datetime as dt
start = dt.date.today() - dt.timedelta(days=120)
yf.Ticker("SPY").history(start=start, end=start + dt.timedelta(days=7), interval="30m", raise_errors=True)
```

```
  File "<string>", line 9, in <module>
  File "yfinance\utils.py", line 124, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "yfinance\base.py", line 129, in history
    return self._lazy_load_price_history().history(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "yfinance\utils.py", line 124, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "yfinance\scrapers\history.py", line 299, in history
    raise _exception
yfinance.exceptions.YFPricesMissingError: $SPY: possibly delisted; no price data found  (30m 2026-03-16 -> 2026-03-23) (Yahoo error = "15m data not available for startTime=1773633600 and endTime=1774238400. The requested range must be within the last 60 days.") (30m resampled from 15m)
```

### Bad data proof

_No response_

### `yfinance` version

dev (a9de4ba)

### Python version

_No response_

### Operating system

_No response_

vignesh-nagarajan-vn:
Every YFPricesMissingError message speculates 'possibly delisted', even when Yahoo's response carries an explicit reason that has nothing to do with delisting, e.g. an intraday range outside the last 60 days. That speculation has misled users for years (see #1713, #1797, #2044, #2052). And when a ticker really is dead, Yahoo's quoted description already says 'No data found, symbol may be delisted', making the prefix redundant there too.

Add possibly_delisted=True parameter to YFTickerMissingError and YFPricesMissingError (default keeps every other callsite unchanged) and pass False in the one history() branch where Yahoo supplied an error description, letting Yahoo's stated reason stand.

Adds an offline test of both message formats and a live test that a range error no longer claims delisting.

Closes #2902

### DIFF
--- a/tests/test_prices.py
+++ b/tests/test_prices.py
@@ -496,6 +496,31 @@ class TestPriceHistory(unittest.TestCase):
         self.assertIn("(30m ", msg)
         self.assertNotIn("(15m ", msg)
 
+    def test_prices_missing_error_delisted_prefix(self):
+        # Offline check of YFPricesMissingError message formats: the
+        # 'possibly delisted' speculation must be suppressible when the
+        # real cause of missing data is known.
+        from yfinance.exceptions import YFPricesMissingError
+        e = YFPricesMissingError("AAA", "(1d 2020-01-01 -> 2020-01-02)")
+        self.assertEqual(str(e), "$AAA: possibly delisted; no price data found (1d 2020-01-01 -> 2020-01-02)")
+        e = YFPricesMissingError("AAA", "(1d 2020-01-01 -> 2020-01-02)", possibly_delisted=False)
+        self.assertEqual(str(e), "$AAA: no price data found (1d 2020-01-01 -> 2020-01-02)")
+
+    def test_range_error_does_not_claim_delisting(self):
+        # A request Yahoo rejects with an explicit range error is not
+        # delisting evidence, so the message must not speculate 'possibly
+        # delisted'. When a ticker really is dead, Yahoo's quoted error
+        # text still communicates it.
+        from yfinance.exceptions import YFPricesMissingError
+        dat = yf.Ticker("SPY", session=self.session)
+        start_d = _dt.date.today() - _dt.timedelta(days=120)
+        end_d = start_d + _dt.timedelta(days=7)
+        with self.assertRaises(YFPricesMissingError) as cm:
+            dat.history(start=start_d, end=end_d, interval="30m", raise_errors=True)
+        msg = str(cm.exception)
+        self.assertNotIn("possibly delisted", msg)
+        self.assertIn("no price data found", msg)
+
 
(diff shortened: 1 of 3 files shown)


ValueRaider:
The remaining prefix still seems repetitive

```
$SPY: no price data found  (period=1y) (Yahoo error = "1m data not available for startTime=1752606940 and endTime=1784142940. Only 8 days worth of 1m granularity data are allowed to be fetched per request.")
$SPSPS: no price data found  (period=1y) (Yahoo error = "No data found, symbol may be delisted")
```

If we're gonna improve messaging, let's go all the way:
```
$SPY: 1m data not available for startTime=1752606940 and endTime=1784142940. Only 8 days worth of 1m granularity data are allowed to be fetched per request.
$SPSPS: No data found, symbol may be delisted
```

vignesh-nagarajan-vn:
Gotcha, simplified the messaging as you suggested, but one thing to note for future changes is that the resample note from [#2900](https://github.com/ranaroussi/yfinance/pull/2900) still appends when the fetched interval differs, so a 30m failure doesn't regress to claiming '15m'

PR #2910, opened 2026-07-21
Fix: don't treat curl_cffi>=0.16 sessions as caching (presence of .cache != active cache)

dokson:
## Summary

`_set_session` flagged a session as "caching" (and refused it) based on the mere *presence* of a `.cache` attribute. curl_cffi >= 0.16 always exposes `.cache` (`None` when disabled), so yfinance's own `new_session()` tripped the check and raised `YFDataException` on **every** call, even with no session passed. curl_cffi 0.15.x lacks the attribute, so the break only shows on 0.16+.

0.16 is currently in beta (`0.16.0b1`), but yfinance pins `curl_cffi>=0.15` with no upper bound: the moment 0.16 final ships, every fresh install breaks. Verified against 0.16.0b1: on `dev`, plain `YfData()` raises; with this fix it works, and a `requests_cache.CachedSession` is still rejected (its `.cache` is a non-None backend).

## Fix

Test for an *active* cache instead: `getattr(session, "cache", None) is not None`. Still rejects `requests_cache` and curl_cffi with caching enabled; accepts a normal session whose `.cache` is `None`.

Also removed the dead `_session_is_caching` machinery: since `_set_session` raises on any active cache, the flag could never be `True`, so the four `expire_after` branches were unreachable (and referenced `self._expire_after`, which is never defined). Worse, setting the flag before the raise left a live singleton in a broken state if a caching session was passed after init.

## Tests

`tests/test_data.py::TestSetSessionCacheDetection` (offline, version-independent): `.cache = None` accepted, active cache rejected. Red on the base branch, green here. Also validated end-to-end in a downstream project pinned to `curl_cffi==0.16.0b1` running this branch.

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -2,7 +2,9 @@
 import unittest
 from functools import lru_cache
 
+from yfinance._http import new_session
 from yfinance.data import SingletonMeta, YfData, _normalize_proxy, lru_cache_freezeargs
+from yfinance.exceptions import YFDataException
 from yfinance.utils import frozendict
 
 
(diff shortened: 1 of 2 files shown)


ValueRaider:
This is an urgent fix, move to main branch.

dokson:
Done: rebased onto main and base switched.

ValueRaider:
You clearly did not review this code.

dokson:
Right: `_session_is_caching` could never be `True` past the raise, and `_expire_after` does not exist. Removed the dead code in 39d55f8. FWIW I am already running this branch from my fork against curl_cffi 0.16.0b1 in another project of mine (dokson/hedge-fund-tracker@859ff4a).

ValueRaider:
Better. Git squash

dokson:
ok boss

Issue #2914, opened 2026-07-22
The 'generic' unit for NumPy timedelta is deprecated.


andriy-git:
Getting depreciation messages even in the latest version 1.5.1:
```
/.venv/lib/python3.14/site-packages/yfinance/scrapers/quote.py:254: DeprecationWarning: The 'generic' unit for NumPy timedelta is deprecated, and will raise an error in the future. This includes implicit conversion of bare integers (e.g. `+1`).Please use a specific unit instead. shares = self._tkr.get_shares_full(start=pd.Timestamp.now('UTC').date()-pd.Timedelta(days=548))

/.venv/lib/python3.14/site-packages/yfinance/utils.py:667: DeprecationWarning: The 'generic' unit for NumPy timedelta is deprecated, and will raise an error in the future. This includes implicit conversion of bare integers (e.g. `+ 1`).Please use a specific unit instead. last_rows_same_interval = (dt2 - dt1) < _pd.Timedelta(interval)
```

DMZ22:
Looked into this. The utils.py site was already reworked on dev in d8e2cfe (sub-day intervals now map to explicit units), and #2915 fixes the remaining quote.py site plus a TypeError regression d8e2cfe introduced for the "5d" interval.

One thing I could not do is reproduce the generic-unit warning itself — tried pandas 2.3.3/3.0.3 x numpy 2.4.6/2.5.1 on Python 3.11/3.12/3.14, all clean. Could you share your numpy version (`python -c "import numpy; print(numpy.__version__)"`)? Wondering if it needs a numpy dev build to trigger.

andriy-git:
I know I didn't install any dev packages on this venv. Here are the versions I am using:
NumPy is version 2.5.1
Pandas is version 2.3.3

After bumping Pandas to version 3.0.3, I stopped getting the deprecation warning. So there is some incompatibility between Pandas and NumPy.

DMZ22:
Thanks for the versions — that was the missing piece, and it let me reproduce it. My earlier "can't reproduce" was wrong: the trigger is **numpy ≥ 2.5.0**, and I had been testing against 2.4.6.

Built your exact stack (Python 3.14, numpy 2.5.1, pandas 2.3.3) and pinned it down. The warning has nothing to do with yfinance — it comes from inside pandas, on *any* `Timedelta` construction:

```python
>>> import pandas as pd
>>> pd.Timedelta('1d')
[...]

PR #2915, opened 2026-07-22
Fix multi-day interval TypeError and actually silence the numpy>=2.5 Timedelta warning

DMZ22:
Two related fixes in the interval/date-arithmetic code. This supersedes the earlier version of this PR, which reordered the `Quote.shares` arithmetic on a rationale that turns out to be wrong — I verified the reordering silenced nothing.

### 1. The numpy>=2.5 "generic unit" warning is not actually being silenced

Issue #2914 reports the `'generic' unit for NumPy timedelta is deprecated` `DeprecationWarning`. #2882 added a workaround in `_interval_to_timedelta` that passes `minutes=` / `hours=` keywords "so numpy>=2.5 does not emit the warning" — but on pandas 2.x + numpy>=2.5 **the keyword form still warns**. Measured on numpy 2.5.1 / pandas 2.3.3:

| construction | warns? |
|---|---|
| `pd.Timedelta("5d")` (bare string) | yes |
| `pd.Timedelta(days=548)` (keyword) | **yes** |
| `pd.Timedelta(minutes=90)` (keyword) | **yes** |
| `pd.Timedelta(90, unit="m")` | no |
| `pd.Timedelta(548, unit="D")` | no |

Only the `(value, unit=...)` form is silent. So this switches `_interval_to_timedelta`'s sub-day branches and `Quote.shares` to that form.

Concretely: **`tests/test_utils.py::TestDateIntervalCheck::test_no_generic_timedelta_deprecation_warning` — the regression test #2882 added — currently fails on numpy 2.5.1 / pandas 2.3.3**, because the keyword workaround doesn't hold. This change makes it pass. (It presumably passes in CI on a numpy<2.5 or pandas-3.0 runner, where the warning doesn't fire at all.)

### 2. TypeError on multi-day intervals

`_interval_to_timedelta` returns a `relativedelta` for day intervals, and `relativedelta` can't be compared with the `Timedelta` `(dt2 - dt1)`, so `_dts_in_same_interval` raises for a multi-day interval:

```python
>>> _dts_in_same_interval(pd.Timestamp("2024-01-01"), pd.Timestamp("2024-01-08"), "5d")
TypeError: '<' not supported between instances of 'Timedelta' and 'relativedelta'
```

Still reproduces on current `dev`. The new `d` branch builds a `Timedelta` directly (with `unit="D"`, so it doesn't reintroduce the warning).

### Verification

Built the exact stack from #2914 (Python 3.14, numpy 2.5.1, pandas 2.3.3):

- every `_interval_to_timedelta` / `_dts_in_same_interval` construction is now clean, with values unchanged (`90m` → 90 min, `1h` → 1 hour, `5d` compares correctly);
- `Timedelta(n, unit=...)` is also clean and correct on numpy 2.4.6 and on pandas 3.0.3, so this doesn't trade one deprecation for another;
- added `TestMultiDayInterval` covering the TypeError case and the multi-day warning. Reverting only the source while keeping the tests fails those two **and** the existing `test_no_generic_timedelta_deprecation_warning`;
- `ruff check` clean on the changed files.

I did **not** touch the pandas-internal side: a user constructing their own `Timedelta(days=...)` will still see the warning until they move to pandas 3.0 (which fixes it at source). This only stops yfinance's own internal constructions from emitting it — which is what the two linked issues are about.

Rebased onto `dev` (`f20d53e`, before the ruff-CI pin) so the branch carries no workflow-file change.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -221,6 +221,32 @@ class TestDateIntervalCheck(unittest.TestCase):
         with self.assertRaises(ValueError):
             self.assertEqual(_parse_user_dt(float(epoch), exchange_tz), expected)
 
+
+class TestMultiDayInterval(unittest.TestCase):
+    def test_multiday_interval_does_not_raise(self):
+        # _interval_to_timedelta returns a relativedelta for day intervals,
+        # which cannot be compared against the Timedelta (dt2 - dt1), so a
+        # multi-day interval used to raise TypeError here.
+        dt1 = pd.Timestamp("2024-01-01")
+        dt2 = pd.Timestamp("2024-01-08")
+        self.assertTrue(_dts_in_same_interval(dt1, dt2, "10d"))
+        self.assertFalse(_dts_in_same_interval(dt1, dt2, "5d"))
+
+    def test_multiday_interval_does_not_warn(self):
+        # Building the comparison Timedelta must not trigger numpy's
+        # "generic unit" DeprecationWarning on pandas 2.x + numpy>=2.5.
+        dt1 = pd.Timestamp("2024-01-01")
+        dt2 = pd.Timestamp("2024-01-04")
+        for interval in ("3d", "5d", "10d"):
+            with warnings.catch_warnings(record=True) as caught:
+                warnings.simplefilter("always")
+                _dts_in_same_interval(dt1, dt2, interval)
+            generic = [w for w in caught
+                       if issubclass(w.category, DeprecationWarning)
+                       and "generic" in str(w.message)]
+            self.assertEqual(generic, [], f"generic-unit warning for {interval!r}")
+
+
 if __name__ == "__main__":
     unittest.main()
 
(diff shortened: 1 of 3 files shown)


ValueRaider:
Is `Timedelta(days=N)` OK?

DMZ22:
No — just re-measured on numpy 2.5.1 / pandas 2.3.3: `Timedelta(days=548)` and `Timedelta(minutes=90)` raise the same generic-unit DeprecationWarning as the bare-string forms. The warning fires in `Timedelta.__new__` whenever pandas has to infer the unit, and the keyword path goes through it too. `Timedelta(548, unit="D")` / `Timedelta(90, unit="m")` are the only silent constructions, and they're also clean on numpy 2.4.6 and pandas 3.0.3 — that's why the PR switches to the `unit=` form. It's the same reason dev's own `test_no_generic_timedelta_deprecation_warning` currently fails under numpy 2.5.1: d8e2cfe switched `_interval_to_timedelta` to `minutes=`/`hours=` kwargs, which still warn.

PR #2918, opened 2026-07-24
Remove dead CachedLimiterSession example from tests/context.py

### LINKED ISSUE #2913: nospam extra installs packages unusable with curl_cffi
The `nospam` extra still installs `requests_cache` and `requests_ratelimiter` (setup.py, also listed in requirements.txt), but both wrap requests-based sessions: caching sessions are rejected by `YfData` and rate-limiter sessions do not work with curl_cffi. So `pip install yfinance[nospam]` pulls in packages that cannot be used. Suggest dropping the extra, or repurposing it once a curl_cffi-compatible limiter exists.

dokson:
The nospam extra and setup.py/requirements.txt were already dropped by the pyproject.toml migration. This PR now only removes the dead CachedLimiterSession example and commented import in tests/context.py, unusable with curl_cffi.

Closes #2913

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,8 +7,6 @@ pytz>=2022.5
 beautifulsoup4>=4.11.1
 lxml>=4.9.0
 peewee>=3.16.2
-requests_cache>=1.0
-requests_ratelimiter>=0.3.1
 scipy>=1.6.3
 # curl_cffi < 0.15 has a CVE
 curl_cffi>=0.15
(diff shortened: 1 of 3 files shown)


dokson:
Heads-up: this overlaps with #2920 (pyproject migration), which already drops the `nospam` extra. Whichever lands first, the other needs a small rebase.

ValueRaider:
merge conflicts

dokson:
already fixed...comment cleaning only now

PR #2920, opened 2026-07-24
Migrate packaging to pyproject.toml

dokson:
Migrates packaging from setup.py + setup.cfg to `pyproject.toml`, with `[project.dependencies]` as the single source of truth, and updates the publish workflow to `python -m build`.

- `setup.py` and `setup.cfg` removed; metadata moved to `pyproject.toml` (dynamic version from `yfinance/version.py`).
- Declare `lxml` as a dependency (pandas.read_html needs it; it was only in `requirements.txt`).
- Add Python 3.11-3.13 classifiers.
- Wheel is now tagged `py3-none-any` instead of the misleading `py2.py3-none-any`.
- Drop the nospam extra (requests_cache/requests_ratelimiter are unusable with curl_cffi).
- Add `[project.urls]` (Documentation, Repository, Issues, Changelog) and set the license to `Apache-2.0`.
- Remove the broken `sample` console script (it pointed at a non-existent module) and a dead `recursive-include` in `MANIFEST.in`.
- Sort dependencies alphabetically.

`requirements.txt` is kept as-is (still used by the no-curl_cffi install path in the docs).

Verified: `python -m build` produces sdist + wheel, `twine check` passes on both, and the installed metadata matches the previous `setup.py` build except for the intended changes above.


ValueRaider:
How did you test the publishing? That's my only worry changing this stuff.

dokson:
Locally with `python -m build`: same sdist + wheel as before, and `twine check` passes on both. I didn't exercise the actual `twine upload` (no PyPI creds).

ValueRaider:
setuptools 61.0.0 requires Python >=3.7

Yeah I know it's EOL but not our problem.

dokson:
You're right. setuptools 59.6.0 is the last version that runs on 3.6, but it is too old for the `[project]` table (that needs >=61, which is py3.7+). So no single version does both. The `py3-none-any` wheel still installs fine on 3.6; only building from source needs 3.7+. I left the classifier as is.

PR #2921, opened 2026-07-25
Fix generic timedelta warning in FastInfo

parthozyy:
## Summary

Replace pandas `Timedelta` arithmetic on a Python `date` with the standard-library `datetime.timedelta`.

This avoids the generic NumPy timedelta deprecation warning in `FastInfo.shares` while preserving the existing 548-day lookback period.

The equivalent deprecated date arithmetic in `tests/test_ticker.py` is also updated.

## Changes

* Use `datetime.timedelta(days=548)` in `FastInfo.shares`.
* Replace `pd.Timedelta(days=1)` with the standard-library `timedelta` in `test_download`.
* Add a deterministic regression test ensuring `FastInfo.shares` does not use pandas `Timedelta` with a Python `date`.

## Testing

```text
python -m unittest \
  tests.test_quote.TestFastInfo.test_shares_does_not_use_pandas_timedelta \
  tests.test_ticker.TestTickerHistory.test_download \
  -v

python -m compileall -q yfinance
git diff --check
```

Result:

```text
Ran 2 tests in 14.484s

OK
```

Addresses the remaining `FastInfo.shares` warning reported in #2914.

### DIFF
--- a/tests/test_quote.py
+++ b/tests/test_quote.py
@@ -1,0 +1,31 @@
+import datetime
+import unittest
+from unittest.mock import MagicMock, patch
+
+import pandas as pd
+
+from yfinance.scrapers.quote import FastInfo
+
+
+class TestFastInfo(unittest.TestCase):
+    def test_shares_does_not_use_pandas_timedelta(self):
+        ticker = MagicMock()
+        ticker.get_shares_full.return_value = pd.Series([100])
+
+        with patch(
+            "yfinance.scrapers.quote.pd.Timedelta",
+            side_effect=AssertionError(
+                "FastInfo.shares must not use pandas Timedelta with a Python date"
+            ),
+        ):
+            shares = FastInfo(ticker).shares
+
+        self.assertEqual(shares, 100)
+        ticker.get_shares_full.assert_called_once()
+
+        start = ticker.get_shares_full.call_args.kwargs["start"]
+        self.assertIsInstance(start, datetime.date)
+
+
+if __name__ == "__main__":
+    unittest.main()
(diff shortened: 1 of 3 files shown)


ValueRaider:
Why did `pd.Timedelta(days=1)` trigger warning?

parthozyy:
> Why did `pd.Timedelta(days=1)` trigger warning?

`pd.Timedelta(days=1)` itself is not generic. The warning is triggered because it is added to a plain Python `datetime.date` returned by `.date()`. That mixed pandas/Python date arithmetic causes conversion through NumPy’s datetime/timedelta machinery, where use of a generic `timedelta64` unit is deprecated in NumPy 2.5. Using `datetime.timedelta(days=1)` keeps the operation entirely within Python’s standard datetime types and avoids that conversion.

ValueRaider:
What is Pandas doing about it?

parthozyy:
@ValueRaider 

Pandas has already fixed this at the source in pandas 3.0, where `pd.Timedelta(days=...)` no longer triggers NumPy’s generic-unit warning.

The warning still occurs with pandas 2.3.x and NumPy 2.5, which is the dependency combination reported in #2914. Since yfinance still supports users on pandas 2.x, it should avoid triggering the deprecated construction path itself.

For yfinance, that means:

- using `datetime.timedelta(...)` when performing arithmetic with a Python `datetime.date`;
- using `pd.Timedelta(value, unit=...)` where a pandas `Timedelta` is actually required.

So upgrading to pandas 3.x also resolves it, but the yfinance-side change prevents the warning for users who remain on supported pandas 2.x versions.

ValueRaider:
You are supposed to be reviewing AI output. Remove the test, restore the line formatting.

parthozyy:
Sorry for the sloppy PR. I should stop and close this one, since the same issue is already being addressed in #2915 with a more complete reproduction and test coverage.

I also understand the review feedback: I should have verified the generated test and explanation more carefully before submitting them.

PR #1726, opened 2023-10-17
Feature proposal for ticker.calendar

rickturner2001:
This is a clearner implementation regarding #1712

PR #2947, opened 2026-08-18
Fix _fetch_calendar using local timezone instead of UTC for event dates

JaneSmith540:
## Summary

`_fetch_calendar` converts Yahoo's calendar-event epochs (always **UTC** timestamps: `dividendDate`, `exDividendDate`, `earningsDate`) with `datetime.datetime.fromtimestamp(ts)` **without a timezone**. On any machine not running in UTC the resulting `.date()` was off by one day — e.g. a `2024-01-01 12:00 UTC` event rendered as `2024-01-02` in UTC+13.

This is related to the open issue #1726 (which refactors `calendar` but does not fix the timezone handling of these three date lines).

### Changes
- **yfinance/scrapers/quote.py** — the 3 `fromtimestamp(...)` calls in `_fetch_calendar` now pass `tz=datetime.timezone.utc`, consistent with the existing `fromtimestamp(..., tz=timezone.utc)` usage in `data.py`.
- **tests/test_ticker.py** — adds a deterministic regression test `test_calendar_dates_use_utc_not_local`: it swaps `datetime.datetime` for a UTC+13 wrapper (only the no-`tz` path is shifted) and asserts the returned Dividend / Ex-Dividend / Earnings dates equal the UTC date. The test **fails on the buggy code** and **passes after the fix** (verified locally).

### Validation
- Environment: Python 3.13 + numpy 2.5.1 + pandas 2.3.3.
- New test `tests.test_ticker.TestTickerMiscFinancials.test_calendar_dates_use_utc_not_local` passes; a negative control (reverting the 3 lines to the buggy form) makes it fail as expected.

### Checklist
- [x] Targets the `dev` branch (not `main`)
- [x] Self-contained, minimal change
- [x] Includes a regression test

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -1045,6 +1045,48 @@ class TestTickerMiscFinancials(unittest.TestCase):
         data_cached = self.ticker.calendar
         self.assertIs(data, data_cached, "data not cached")
 
+    def test_calendar_dates_use_utc_not_local(self):
+        # Regression: _fetch_calendar must interpret Yahoo's calendar-event
+        # epoch (always UTC) as a UTC calendar date, not a local-timezone date.
+        # On a non-UTC machine the buggy code returned a date off by one day.
+        # We simulate a local timezone of UTC+13 (by swapping the datetime
+        # class for a wrapper) so the test fails even on UTC CI runners if the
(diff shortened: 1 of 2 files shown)


ValueRaider:
AI slop

PR #2948, opened 2026-08-18
Fix custom-period window start timezone (use exchange tz, bound events)

JaneSmith540:
## Summary
For custom (non-standard) periods, `PriceHistory.history()` computed the window start with `datetime.date.fromtimestamp(end)`, which is **local-timezone dependent**, and then only used the value as a non-`None` sentinel — the actual trimming of dividends/splits/capital-gains used the first price row instead.

This PR makes the start computation correct **and** actually uses it:
- Compute the window start in the **exchange timezone** via the UTC epoch, independent of the machine's local timezone (consistent with the normal-period branch, e.g. `pd.Timestamp.now('UTC').tz_convert(tz).date()`).
- Use that start date to trim the returned events to the requested window, instead of only gating on `start` being set.

The start computation was extracted into `_custom_period_start()` for clarity.

## Why this matters
On a non-UTC machine (e.g. UTC+8) the old `fromtimestamp(end)` could shift the window start by a day. More importantly, the old code never bounded events by the custom period at all, so dividend/split/capital-gains rows that fell before the first price row but outside the requested window were incorrectly kept.

## Tests
- `test_custom_period_start_uses_exchange_tz` — asserts the helper returns the exchange-timezone date (fails if it ever reverts to local/UTC time).
- `test_custom_period_binds_events_to_window` — end-to-end (mocked fetch): a dividend between the first price row and the true window start is dropped by the fix, but was kept before.

## Scope / risk
Minimal: only the custom-period branch is touched. The `isinstance(start, date)` guard ensures the normal / `max` / user-start paths (where `start` is an int epoch) are unchanged.

### DIFF
--- a/tests/test_ticker.py
+++ b/tests/test_ticker.py
@@ -1045,6 +1045,89 @@ class TestTickerMiscFinancials(unittest.TestCase):
         data_cached = self.ticker.calendar
         self.assertIs(data, data_cached, "data not cached")
 
+    def test_custom_period_start_uses_exchange_tz(self):
+        # Regression: _custom_period_start() must compute the window start as a
+        # date in the *exchange* timezone, independent of the machine's local
+        # timezone. The old code used datetime.date.fromtimestamp(ts) (local tz),
+        # which could shift the start by a day on non-UTC machines.
+        import datetime as _dt
+        import yfinance.scrapers.history as hist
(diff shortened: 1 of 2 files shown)


ValueRaider:
AI slop

PR #2949, opened 2026-08-19
fix(calendars): degrade to an empty frame on a malformed response

mateosandoval10:
## Summary

`Calendars._get_data` deliberately falls back to an empty payload when the response body cannot be decoded, then indexes straight into it, so the fallback raises `KeyError` immediately after logging that the fetch failed.

```python
try:
    json_data = response.json()
except json.JSONDecodeError:
    self._logger.error(f"{calendar_type}: Failed to retrieve calendar.")
    json_data = {}

# Error returned
if json_data.get("finance", {}).get("error", {}):
    raise YFException(...)

self.calendars[calendar_type] = self._create_df(json_data)   # <-- KeyError: 'finance'
```

`_create_df` reads `json_data["finance"]["result"][0]["documents"][0]`, which has no tolerance for the `{}` the branch above just assigned. The graceful path the `try`/`except` was written for never runs — a caller hitting a non-JSON response (an HTML error page, a rate-limit interstitial) gets a `KeyError` rather than an empty calendar, and the logged "Failed to retrieve calendar" is followed by an exception that does not mention the calendar at all.

The same indexing raises `IndexError` on a well-formed envelope with no documents, which is the shape a query matching nothing would take.

```python
cal = yf.Calendars()
cal._create_df({})                                          # KeyError: 'finance'
cal._create_df({"finance": {"result": []}})                 # IndexError: list index out of range
cal._create_df({"finance": {"result": [{"documents": []}]}})  # IndexError: list index out of range
```

## Change

`_create_df` resolves the document once and returns an empty `DataFrame` if it is not there, and reads `columns` / `rows` with defaults:

```python
try:
    document = json_data["finance"]["result"][0]["documents"][0]
except (KeyError, IndexError, TypeError):
    self._logger.debug("Calendar response contained no result document.")
    return pd.DataFrame()
```

`TypeError` covers a null envelope (`{"finance": None}`). Nothing downstream needed changing — `_cleanup_df` already returns early on an empty frame, so `_get_data` now returns an empty calendar in exactly the case its error handling was aiming for.

The duplicate-column rename (the second `Event Start Date` becoming `Timing`) is untouched, and is covered by a new test so it stays that way.

## Testing

Four new tests in `tests/test_calendars.py`, in a separate `TestCalendarsResponseHandling` case. They need no network, which the existing calendar tests all do:

- `test_create_df_undecodable_response` — the `{}` payload `_get_data` falls back to.
- `test_create_df_no_results` — three shapes of empty envelope, as subtests.
- `test_create_df_parses_document` — a normal document still parses, including the `Timing` rename.

```
$ python -m pytest tests/test_calendars.py -k ResponseHandling -q
3 passed, 3 subtests passed
```

On `main` the same tests fail with the `KeyError` / `IndexError` above (4 failed, 2 passed).

`ruff check` is clean on both files.

### DIFF
--- a/tests/test_calendars.py
+++ b/tests/test_calendars.py
@@ -51,5 +51,49 @@ class TestCalendars(unittest.TestCase):
         self.assertEqual(len(result), 5)
 
 
+class TestCalendarsResponseHandling(unittest.TestCase):
+    """Offline tests for parsing of malformed or empty calendar responses."""
+
+    def setUp(self):
+        self.calendars = yf.Calendars(session=session_gbl)
+
+    def test_create_df_undecodable_response(self):
+        # _get_data falls back to {} when response.json() raises, so parsing
+        # that payload has to degrade to an empty frame rather than KeyError.
+        df = self.calendars._create_df({})
+
+        self.assertIsInstance(df, pd.DataFrame)
+        self.assertTrue(df.empty)
+
+    def test_create_df_no_results(self):
+        for payload in (
+            {"finance": {"result": []}},
+            {"finance": {"result": [{"documents": []}]}},
+            {"finance": None},
+        ):
+            with self.subTest(payload=payload):
+                df = self.calendars._create_df(payload)
+
+                self.assertIsInstance(df, pd.DataFrame)
+                self.assertTrue(df.empty)
+
+    def test_create_df_parses_document(self):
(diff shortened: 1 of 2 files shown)


ValueRaider:
AI slop

Issue #2925, opened 2026-07-27
Last days data missing


plajos1:
### Describe bug

Downloadig with

yf.download(ticker, start = start, end = end)

everything is fine except the last day: only the volume appears in the last line.

### Simple code that reproduces your problem

datar = yf.download(ticker, start = start, end = end)

### Debug log with debug logging enabled

[*********************100%***********************]  1 of 1 completed
DEBUG     Disabling multithreading because DEBUG logging enabled
DEBUG     Entering history()
DEBUG      Entering history()
DEBUG       MU: Yahoo GET parameters: {'period1': '2021-07-28 00:00:00-04:00', 'period2': '2026-07-27 00:00:00-04:00', 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG       MU: yfinance received OHLC data: 2021-07-28 13:30:00 -> 2026-07-24 13:30:00
DEBUG       MU: OHLC after cleaning: 2021-07-28 09:30:00-04:00 -> 2026-07-24 09:30:00-04:00
DEBUG       MU: OHLC after combining events: 2021-07-28 00:00:00-04:00 -> 2026-07-24 00:00:00-04:00
DEBUG       MU: yfinance returning OHLC: 2021-07-28 00:00:00-04:00 -> 2026-07-24 00:00:00-04:00
DEBUG      Exiting history()
DEBUG     Exiting history()
DEBUG    Exiting download()
DEBUG    Entering download()
DEBUG     Disabling multithreading because DEBUG logging enabled
DEBUG     Entering history()
DEBUG      Entering history()
DEBUG       MU: Yahoo GET parameters: {'period1': '2021-07-28 00:00:00-04:00', 'period2': '2026-07-27 00:00:00-04:00', 'interval': '1d', 'includePrePost': False, 'events': 'div,splits,capitalGains'}
DEBUG       MU: yfinance received OHLC data: 2021-07-28 13:30:00 -> 2026-07-24 13:30:00
DEBUG       MU: OHLC after cleaning: 2021-07-28 09:30:00-04:00 -> 2026-07-24 09:30:00-04:00
DEBUG       MU: OHLC after combining events: 2021-07-28 00:00:00-04:00 -> 2026-07-24 00:00:00-04:00
DEBUG       MU: yfinance returning OHLC: 2021-07-28 00:00:00-04:00 -> 2026-07-24 00:00:00-04:00
DEBUG      Exiting history()
[...]

PR #2952, opened 2026-08-21
Drop volume-only rows with missing prices

mameikagou:
## Summary

Yahoo chart responses can contain rows with volume but no OHLC or adjusted-close values. Treat these as missing when `keepna=False`, while preserving event-only rows and the existing `keepna=True` behavior.

Addresses #2925.

## Tests

- Added an offline chart-response regression test covering malformed volume-only rows, event-only rows, and `keepna=True`
- `tests/test_prices.py` focused regression test
- `tests/test_utils.py` excluding the environment-dependent timezone case
- Ruff and `git diff --check`

### DIFF
--- a/tests/test_prices.py
+++ b/tests/test_prices.py
@@ -1,5 +1,6 @@
 from tests.context import yfinance as yf
 from tests.context import session_gbl
+from yfinance.scrapers.history import PriceHistory
 
 import unittest
 import socket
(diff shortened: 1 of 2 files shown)


ValueRaider:
You got a real example? And what happens when you enable `repair=True`?

mameikagou:
Yes. #2925 includes a concrete MU example on 2026-07-24: all OHLC fields are `NaN` while volume is `40,626,435`; comments report the same shape for AAPL and several indices.

I replayed that payload against this PR. With `repair=False`, the volume-only row is dropped when `keepna=False`. With `repair=True`, the 1-hour repair path runs first; if it recovers prices, the repaired row remains, and if it still has no OHLC data, the final cleanup drops it. `keepna=True` continues to preserve the original row. The separate event-only row remains preserved in both modes, and the focused regression test passes.

ValueRaider:
The volume is real data, it should not be dropped.

The real fix is applying price-repair to that row - which should be happening already via `_fix_zeroes`.

mameikagou:
Thanks, agreed. I replayed the MU 2026-07-24 shape against current `dev`. With `repair=False`, the non-zero volume row is preserved. With `repair=True`, `_fix_zeroes` reconstructs the missing OHLC from the nested 1h response and preserves the volume. My fixture returned the same malformed daily payload for the nested 1h request, so it could not exercise repair and then incorrectly dropped the row. The row-dropping change should be discarded.

PR #2953, opened 2026-08-24
Improve robustness of `YfData` when behind a SOCKS5 proxy

amosli:
## Summary

Fixes four related robustness problems in `YfData` that break `history()` behind SOCKS5 proxies, even when the chart API itself is reachable:

1. **Proxy wipe (the critical bug).** `_make_request()` unconditionally ran
   `self._session.proxies = _normalize_proxy(YfConfig.network.proxy)`. With no global proxy configured (the default), this reset `proxies=None`, silently disabling any SOCKS5 proxy the user set directly on their `curl_cffi` session → all data requests timed out (`curl: (28)`), while direct use of the same session worked.
   Now proxies are only overwritten when `YfConfig.network.proxy is not None`.

2. **`fc.yahoo.com` timeouts were fatal.** `_get_cookie_basic()` caught only `DNSError`; a curl error 28 (common for `fc.yahoo.com` through proxies) aborted the entire data request. Transient errors are now handled like DNS errors: log a warning and continue without the cookie (the chart API doesn't need it).

3. **Crumb fetch 429/timeouts were fatal.** `_get_crumb_basic()` raised `YFRateLimitError` on 429 and transient network errors propagated out of `_get_cookie_and_crumb()`, aborting requests to endpoints (e.g. chart API) that work without a crumb. `_make_request()` now degrades to a crumb-less request. Endpoints that do require a crumb still surface their own failure and trigger the existing cookie-strategy-toggle retry; a genuine 429 from the target endpoint still raises `YFRateLimitError`.

4. **`'crumb': None` written into params.** In the >=400 retry branch, a failed secondary crumb fetch put `crumb=None` into the request params. Now a failed refetch simply omits the parameter.

No public API changes. Cookie/crumb reuse and both cookie strategies are untouched.

## Testing

New test module `tests/test_cookie_crumb_degradation.py`:

- `test_fc_yahoo_timeout_is_not_fatal` — timeout degrades instead of raising
- `test_getcrumb_429_degrades_to_no_crumb_request` — 429 on getcrumb, chart still 200
- `test_transient_crumb_failure_degrades` — getcrumb timeout, chart still 200
- `test_target_429_still_raises_rate_limit` — real rate limiting still surfaces
- `test_custom_proxies_not_overwritten` / `test_make_request_does_not_wipe_session_proxies` — regression tests for the proxy wipe

All pass locally alongside existing `tests/test_data.py` (21 passed); `ruff check` clean.

Real-world verification with `socks5h://` + `curl_cffi` (firefox impersonation, HTTP/2):

```
✅ proxy check: HTTP 200, HTTP/3
✅ crumb check: HTTP 200, HTTP/3
Fetching 005930.KS history...
✅ got 14 rows
```

### DIFF
--- a/tests/test_cookie_crumb_degradation.py
+++ b/tests/test_cookie_crumb_degradation.py
@@ -1,0 +1,122 @@
+import unittest
+from unittest import mock
+
+from yfinance._http import requests
+from yfinance.data import SingletonMeta, YfData
+
+
+def _mock_response(status_code=200, text="", url="https://query1.finance.yahoo.com/v8/finance/chart/AAPL"):
+    resp = mock.MagicMock()
+    resp.status_code = status_code
+    resp.text = text
+    resp.content = text.encode() or b"<html></html>"
+    resp.url = url
+    resp.json.return_value = {}
+    return resp
+
+
+class _DegradeTestBase(unittest.TestCase):
+    """YfData is a singleton: use throwaway instances so mocked state
+    cannot leak into other tests."""
+
+    def setUp(self):
+        SingletonMeta._instances.pop(YfData, None)
+        self.data = YfData()
+        self.data._cookie = None
+        self.data._crumb = None
+        self._no_persistent_cookie = mock.patch.object(
+            YfData, '_load_cookie_curlCffi', return_value=False)
+        self._no_persistent_cookie.start()
+
+    def tearDown(self):
+        self._no_persistent_cookie.stop()
+        SingletonMeta._instances.pop(YfData, None)
+
+
+class TestCookieFetchDegradation(_DegradeTestBase):
+    def test_fc_yahoo_timeout_is_not_fatal(self):
+        # fc.yahoo.com timing out behind a SOCKS5 proxy must degrade,
+        # not abort the eventual data request.
+        self.data._session.get = mock.MagicMock(
+            side_effect=requests.exceptions.Timeout("curl: (28) Connection timed out"))
+        self.assertFalse(self.data._get_cookie_basic())
+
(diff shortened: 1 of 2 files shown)


ValueRaider:
Rebase to `dev` branch.

1. How is proxy being set if not via `YfConfig.network.proxy`?

amosli:
Rebased onto `dev`, thanks for the pointer.

Regarding how a proxy gets set without `YfConfig.network.proxy`: the common pattern is passing your own session with the proxy configured directly on it — this is the standard setup when using `curl_cffi` for browser impersonation (the proxy has to be applied inside curl, so it lives on the session):

```python
from curl_cffi import requests

session = requests.Session(
    impersonate="firefox",
    http_version=CurlHttpVersion.V2TLS,
    proxies={"http": "socks5h://127.0.0.1:1080", "https": "socks5h://127.0.0.1:1080"},
)
tk = yf.Ticker("AAPL", session=session)
```

`_set_session()` handles this correctly — it only overrides `session.proxies` when `YfConfig.network.proxy is not None`. But `_make_request()` unconditionally ran:

```python
self._session.proxies = _normalize_proxy(YfConfig.network.proxy)
```

With no global proxy configured (the default), that assigns `None` on **every** data request, silently disabling the user's session-level proxy. All requests then bypass the proxy; behind a network where Yahoo is only reachable through the proxy, every request times out (`curl: (28)`), even though the same session works fine when used directly.

The fix simply makes `_make_request()` consistent with `_set_session()`: only overwrite `session.proxies` when the global config actually defines a proxy. Explicitly setting `yf.config.network.proxy` still takes precedence as before.

Covered by regression test `test_make_request_does_not_wipe_session_proxies`.

PR #2961, opened 2026-08-28
fix: set business-day freq on daily price history

### LINKED ISSUE #1083: Lack of Freq String for Daily Prices
From https://github.com/sktime/sktime/issues/3543#issuecomment-1272305001
It seems that when acquiring the daily data, the `freq` string is not set to `'B'` or "business days", so some of the code may not work normally.

LeonxLJX:
Set index.freq to BusinessDay on daily (1d) price history so that
downstream libraries (e.g. sktime) can infer the frequency instead of
receiving None.

- Only applies to daily intervals; weekly/monthly/intraday untouched.
- Wrapped in try/except so indices with gaps (trading halts, delisting)
  degrade gracefully and keep returning data as before.
- Adds a regression test (`test_daily_freq`).

Fixes #1083

### DIFF
--- a/tests/test_prices.py
+++ b/tests/test_prices.py
@@ -32,6 +32,20 @@ class TestPriceHistory(unittest.TestCase):
                 f = df.index.time == _dt.time(0)
                 self.assertTrue(f.all())
 
+    def test_daily_freq(self):
+        # Daily data should expose a business-day frequency so that
+        # downstream libraries (e.g. sktime) can rely on it (issue #1083).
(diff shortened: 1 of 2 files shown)


ValueRaider:
Can this cover more intervals, as many as Pandas supports? Also rebase to dev branch https://ranaroussi.github.io/yfinance/development/code.html

LeonxLJX:
Superseded by #2962, which addresses the review feedback: covers more intervals (1d/1wk/1mo/3mo) and is rebased onto the dev branch as requested.

Issue #565, opened 2021-01-24
Hourly Prices not adjusting for dividends


erlc05:
stockdb = yf.download(tickers=stock_list2,period='730d', interval='60m',auto_adjust = True, prepost = False)

for the ticker AAPL

2020-10-29   09:30:00-04:00 | 112.37 | 114.365 | 112.2 | 114.36 | 27783125

These are not adjusted prices.  Any helps or tips would be greatly appreciated.

Thanks!!
-Eric

erlc05:
How can I help get this corrected? @ranaroussi

PR #2967, opened 2026-09-07
State whether price adjustment was applied, on every return path

arian-gogani:
## The problem

`download()` and `history()` default to `auto_adjust=True`. When adjustment succeeds, `Close` holds the adjusted value and `Adj Close` is dropped. When adjustment is *requested but raises*, the exception is swallowed on shipped defaults and `Close` holds the **raw** value — under the same column name.

The swallow is by design and I am not proposing to change it:

```python
except Exception as e:
    if raise_errors or (not YfConfig.debug.hide_exceptions):
        raise
    ...
    logger.error('%s: %s' % (self.ticker, err_msg))
```

But on the shipped configuration both signals are off:

```
yfinance/config.py:34   d.hide_exceptions = True     # exception swallowed
yfinance/config.py:35   d.logging = False            # logger.error not surfaced
```

So a caller that asked for adjusted prices can receive unadjusted ones with no exception, no warning, and no log line. Across a 4-for-1 split that is a 4x difference in `Close`.

The only tell is `Adj Close` still being present after a run that requested adjustment, which is a column that should have vanished — not something callers check for.

## Reproduction

Replicating this block's control flow verbatim under the shipped defaults, no network:

```
caller asks download(auto_adjust=True), shipped default config

  adjustment succeeds            exception: no    df['Close'] = 100.00   Adj Close: consumed by adjust
  raises: Adj Close absent       exception: no    df['Close'] = 400.00   Adj Close: was never there
  raises: Adj Close non-numeric  exception: no    df['Close'] = 400.00   Adj Close: present
```

All three return a DataFrame and none raises.

## The change

Record on the frame whether adjustment was actually applied, using `DataFrame.attrs`, via a local helper applied on **every** return path:

```python
def _stamp(frame, applied=False):
    frame.attrs['adjustment_requested'] = bool(auto_adjust or back_adjust)
    frame.attrs['auto_adjust'] = bool(auto_adjust) and applied
    frame.attrs['back_adjust'] = bool(back_adjust) and applied
    return frame
```

`history()` has four returns — three early `utils.empty_df()` paths and the success path. All four are stamped, so `df.attrs['auto_adjust']` is always present and a caller never has to guard with `.get()`. Stamping only the success path would have made an empty result raise `KeyError` on the same check, which is a worse contract than none.

- **Non-breaking.** No signature, column, dtype or index change. `attrs` is currently unused anywhere in yfinance, so nothing can collide.
- **Stamped last** on the success path, after every reshaping step, so nothing downstream can drop it.
- **Costs nothing** when unused.

It also answers a second question callers currently cannot answer: given a DataFrame, is `Close` adjusted? Since 0.2.51 the `download()` default has been `auto_adjust=True`, so frames from different call sites carry different meanings under the same column name.

## Tests

10 cases, both directions:

```
adjustment raised (Adj Close absent)      -> attrs says NOT adjusted
adjustment raised (Adj Close non-numeric) -> attrs says NOT adjusted
adjustment succeeded                      -> attrs says adjusted
auto_adjust=False                         -> not adjusted, not requested
```

plus the three early-return paths, and that the key set is identical and non-empty across them.

Verified falsifiable rather than assumed: with the patch 10 pass; with the change reverted the same suite reports **8 failures**. The two that still pass are value assertions that correctly do not depend on this change.

One assertion was rewritten during that check. "Same key set as the success path" passed without the patch too, because both sides were empty dicts — a check that cannot distinguish the feature being absent is not a check. It now also requires the key set to be non-empty.

`attrs` survival was checked through every operation that follows in `history()` — `np.round`, column assignment, index rename, `copy`, column selection, `concat`, `sort_index` — all retain it on pandas 3.0.2. Setting it last means `_resample` and the `drop` paths cannot affect it either.

Also confirmed this adds no lint regressions: `ruff check --select E,W,F --ignore E501` reports the same count on this branch as on `dev`.

## What I could not verify

I have no network access to Yahoo in the environment I built this in, so I exercised the adjustment block's control flow directly rather than through a live `history()` call. A test wired into your suite with your fixtures would be better and I am happy to write it if you want this.

I have also not measured how often adjustment actually raises in practice. The argument here is about what the caller can determine, not about frequency.

## Disclosure

I used Claude (Anthropic) to investigate this and draft the change, at my direction. Every command and figure quoted above was executed and every output is real; the reproduction and the test are included precisely so none of it has to be taken on trust. Happy to adjust anything here on review.


ValueRaider:
Provider a ticker

arian-gogani:
PEP.

```python
import yfinance as yf
t = yf.Ticker('PEP')
a = t.history(period='1mo', interval='1h', auto_adjust=True)
b = t.history(period='1mo', interval='1h', auto_adjust=False)
d = t.history(period='1mo', interval='1d', auto_adjust=True)
e = t.history(period='1mo', interval='1d', auto_adjust=False)
print('1h  max|adjusted - unadjusted| =', round(float((a['Close']-b['Close']).abs().max()), 6))
print('1d  max|adjusted - unadjusted| =', round(float((d['Close']-e['Close']).abs().max()), 6))
print('dividend paid 2026-09-04      =', float(t.dividends.iloc[-1]))
```

On 2026-09-11 that prints:

```
1h  max|adjusted - unadjusted| = 0.0
1d  max|adjusted - unadjusted| = 1.529144
dividend paid 2026-09-04      = 1.48
```

At 1h, `auto_adjust=True` and `auto_adjust=False` return the same numbers to the last decimal place. At 1d the same pair of calls differs. The 1d figure is a little larger than the dividend itself because the adjustment is a ratio applied across the window rather than a flat subtraction, which is expected.

A control, so this is not intraday noise. On PEP sessions with no dividend between them, the last 1h bar and the daily close differ by about a cent: 2026-09-09 by +0.01, 2026-09-10 by -0.01, 2026-09-11 by 0.00. The gap in the comparison above is 1.48.

Four more, same day, each showing the last session before that ticker's dividend as daily adjusted against daily unadjusted, then hourly which is the same value either way:

- HD, 2.33 on 2026-09-03: daily 316.18 against 318.51, hourly 318.54 both ways
- MCD, 1.86 on 2026-09-01: daily 261.68 against 263.54, hourly 263.53 both ways
- JNJ, 1.34 on 2026-08-25: daily 271.70 against 273.04, hourly 273.08 both ways
- MSFT, 0.91 on 2026-08-20: daily 483.40 against 484.31, hourly 484.48 both ways

This is not new and I am not presenting it as a discovery. #565 reported the same symptom in January 2021 and was closed in March 2024 without a fix.

The ask in this issue is narrower than fixing the adjustment. A caller who passes `auto_adjust=True` at an intraday interval receives unadjusted prices and nothing in the return path says so. If the returned object stated which adjustment was actually applied, a caller could tell the difference without having to already suspect the problem and test for it. Whatever is decided about intraday adjustment itself, the silent part is separable and is the part that turns into wrong numbers downstream.

Measured with yfinance 1.7.0 on 2026-09-11.

ValueRaider:
AI slop

